PR #17, opened 2021-11-01
add dtx support

davidliu:
Wasn't able to find a way to monitor traffic differences from ios simulator, may need a real device for that (Instruments doesn't support network monitoring for simulators).

### DIFF
--- a/Sources/LiveKit/core/Engine.swift
+++ b/Sources/LiveKit/core/Engine.swift
@@ -193,11 +193,11 @@ class Engine: MulticastDelegate<EngineDelegate> {
     func addTrack(cid: String,
                   name: String,
                   kind: Livekit_TrackType,
-                  dimensions: Dimensions? = nil) -> Promise<Livekit_TrackInfo> {
+                  _ populator: (inout Livekit_AddTrackRequest) -> ()) -> Promise<Livekit_TrackInfo> {
 
         // TODO: Check if cid already published
 
-        signalClient.sendAddTrack(cid: cid, name: name, type: kind, dimensions: dimensions)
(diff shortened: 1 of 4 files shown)


Sources/LiveKit/core/Engine.swift
```diff
@@ -193,11 +193,11 @@ class Engine: MulticastDelegate<EngineDelegate> {
     func addTrack(cid: String,
                   name: String,
                   kind: Livekit_TrackType,
-                  dimensions: Dimensions? = nil) -> Promise<Livekit_TrackInfo> {
+                  _ populator: (inout Livekit_AddTrackRequest) -> ()) -> Promise<Livekit_TrackInfo> {
```
hiroshihorie, line 196: I think we can just pass a simple `dtx: Bool = true` to keep things simple ? 
like https://github.com/livekit/client-sdk-flutter/blob/721b04753a9e83e7815f694542b65db1f0eaea4d/lib/src/rtc_engine.dart#L149

hiroshihorie: But interesting design is this to support many parameters in the future ?

hiroshihorie (approved):

Sources/LiveKit/core/Engine.swift
davidliu: Hmm, yeah, was doing this to keep things clean so that we don't have to deal with random args polluting the method declaration (since audio args don't care about video args, and vice versa).

hiroshihorie: Oh thats actually a good idea 👌

PR #18, opened 2021-11-01
`VideoCaptureInterceptor` for frame processing

hiroshihorie:
This is the example usage https://github.com/livekit/multiplatform-swiftui-example/commit/c003dc2621a4789172de734f7791379bd36aabd4

**Consideration**
- Current design assumes the frame processing will be quite fast, I haven't tested what will happen if processing is slow (developer obviously needs to be careful).

### DIFF
--- a/Sources/LiveKit/track/LocalVideoTrack.swift
+++ b/Sources/LiveKit/track/LocalVideoTrack.swift
@@ -28,13 +28,24 @@ public class LocalVideoTrack: VideoTrack {
         super.init(rtcTrack: rtcTrack, name: name)
     }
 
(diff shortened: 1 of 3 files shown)


davidzhao:
is it possible to chain them? i.e. in the future, would be flexible enough to do <interceptor 1> -> <interceptor 2>, etc.

Even for things such as end to end encryption, could be done via these kind of interceptors.

boks1971 (approved):
This is great. Like @davidzhao mentioned would be good to think about chaining at some point.

Generally thinking about media as a pipe and processors along the path would be great. Like the gstreamer design.

hiroshihorie:
Ok boss(es) !!! Will check out gstreamer design

boks1971:
hahaha I am just a pawn in the chess of life :-)

PR #20, opened 2021-11-04
Connection quality

hiroshihorie:


Sources/LiveKit/participant/RemoteParticipant.swift
```diff
@@ -4,8 +4,9 @@ import Promises
 
 public class RemoteParticipant: Participant {
 
-    init(sid: String, info: Livekit_ParticipantInfo?) {
+    init(sid: String, info: Livekit_ParticipantInfo?, room: Room) {
         super.init(sid: sid)
+        self.room = room
```
boks1971, line 9: Is this just for consistency in `init` signature with `LocalParticipant` as this new field does not seem to be used anywhere OR is it some interface/implementation making it necessary to change this signature?

boks1971 (approved):
lgtm! One small question.

Sources/LiveKit/participant/RemoteParticipant.swift
hiroshihorie: It's actually used in the parent Participant class, which I forgot to do before! So I fixed it now 😅

hiroshihorie:
Thanks Raja oniichaannn !! 🙏

Sources/LiveKit/participant/Participant.swift
```diff
@@ -8,18 +8,25 @@ public class Participant: MulticastDelegate<ParticipantDelegate> {
...
+            notify { $0.participant(self, didUpdate: self.metadata) }
+            room?.notify { $0.room(self.room!, participant: self, didUpdate: self.metadata) }
+        }
+    }
+
+    public internal(set) var connectionQuality: ConnectionQuality = .unknown {
+        didSet {
+            guard oldValue != connectionQuality else { return }
+            notify { $0.participant(self, didUpdate: self.connectionQuality) }
+            room?.notify { $0.room(self.room!, participant: self, didUpdate: self.connectionQuality) }
```
hiroshihorie, line 29: `room` is used here 😄

Sources/LiveKit/participant/RemoteParticipant.swift
boks1971: Got it. Thank you!

PR #23, opened 2021-11-11
fix compilation error for the iOS target

jeksys:


hiroshihorie:
Oh this is already fixed in the master, thanks!

PR #25, opened 2021-11-21
Dynamically update sender parameters

hiroshihorie:
The intension for this change is to fix that:
Previously, if there are no dimensions info, simulcast would never be enabled.

With this change simulcast will always be enabled (if its true in the options),
and the sender parameters will **lazily** be updated when dimensions info 
is provided by the capturer. (even when dimensions change)

Many of the capturers cannot report "real dimensions" until they produce a frame,
it's only possible to estimate (which is in accurate)

### Thoughts
What should be the desired behavior when the user provides VideoEncoding(max frameRate, maxBitrate) in the publish options, and Simulcast is enabled.

### DIFF
--- a/Sources/LiveKit/types/VideoParameters.swift
+++ b/Sources/LiveKit/types/VideoParameters.swift
@@ -2,7 +2,7 @@ import Foundation
 import WebRTC
 
 extension Collection where Element == VideoParameters {
-    
+
     func suggestedPresetIndex(dimensions: Dimensions? = nil,
                               videoEncoding: VideoEncoding? = nil) -> Int {
         // self must at lease have 1 element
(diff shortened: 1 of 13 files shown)


Sources/LiveKit/types/VideoParameters.swift
```diff
@@ -1,6 +1,32 @@
 import Foundation
 import WebRTC
 
+extension Collection where Element == VideoParameters {
+
+    func suggestedPresetIndex(dimensions: Dimensions? = nil,
+                              videoEncoding: VideoEncoding? = nil) -> Int {
+        // self must at lease have 1 element
+        assert(!isEmpty)
+        // dimensions or videoEndocing is required
```
boks1971, line 10: :nit: `VideoEncoding` spelling

Sources/LiveKit/types/VideoParameters.swift
```diff
@@ -1,6 +1,32 @@
...
+        assert(dimensions != nil || videoEncoding != nil)
+
+        var result = 0
+        for preset in self {
+            if let dimensions = dimensions,
```
boks1971, line 15: There was a similar utility routine ☝️ to search presets. Maybe use that to avoid repetition and keep logic in one place?

boks1971 (commented):
@hiroshihorie Left some comments about ordering of tracks.

Given that dimensions are lazily set, I think what you have is good, i. e. find the nearest preset from the exact dimensions and then adjust accordingly. Do you see anything bad with this approach?

hiroshihorie:
Thanks oniichan! I will need to re-read your Slack comments & JS PR !! 😁🙏

PR #27, opened 2021-11-26
Fix Xcode 12 compile issues

hiroshihorie:
Fixes issue https://github.com/livekit/client-sdk-ios/issues/6#issuecomment-979694895

I have no idea why in some cases Xcode 12 throws compile errors when **unwrapping variable with the same name**

According to SO, this should be valid even for `guard` since Xcode 7.1.1 • Swift 2.1 :
https://stackoverflow.com/a/28709006/1993405

I'm not sure this might be a Swift bug https://bugs.swift.org/browse/SR-214

For now, I will avoid variable shadowing...


davidzhao (approved):
good find. This might be due to swift version differences? Xcode 12 likely doesn't support the swift version where this became possible.

hiroshihorie:
both Xcode 12, 13 are using Swift 5, but maybe there was a change within the Swift 5 version 🤔

PR #28, opened 2021-11-29
Pass capture options for `setCameraEnabled` and `setMicrophone`

hiroshihorie:
I can add the default options to `ConnectOptions` but that will be only used when calling
`setCameraEnabled` `setMicrophoneEnabled` etc since creating track directly by `LocalAudioTrack.createTrack` has no access to the `Room.connectOptions`

https://linear.app/livekit/issue/LK-318/swift-support-default-capture-settings

### DIFF
--- a/Sources/LiveKit/Participant/LocalParticipant.swift
+++ b/Sources/LiveKit/Participant/LocalParticipant.swift
@@ -244,26 +244,17 @@ extension LocalParticipant {
         return set(source: .microphone, enabled: enabled)
     }
 
-    public func setScreenShare(enabled: Bool, source: ScreenShareSource? = nil) -> Promise<LocalTrackPublication?> {
-        return set(source: .screenShareVideo, enabled: enabled, screenShareSource: source)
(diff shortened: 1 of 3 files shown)


davidzhao:
> since creating track directly by `LocalAudioTrack.createTrack` has no access to the `Room.connectOptions`

I think the idea is that we could set defaults for the room that get honored. Would it be possible to access those options somehow?

hiroshihorie:
@davidzhao I could change the timing of the actual `Engine.factory.audioTrack` being called until `LocalAudioTrack.start` is called.

hiroshihorie:
I think `setCamera(enabled:)` can still have `options` even if there will be default options in `ConnectOptions`.
just like publish options if empty, will fall back to options in `ConnectOptions`.

hiroshihorie:
options have been removed

hiroshihorie:
Ok setXXXX API's params have been simplified, If user wants to pass param they should use `publishTrack`

davidzhao (approved):

PR #29, opened 2021-11-29
Track State Updates

hiroshihorie:
@davidzhao @boks1971 @davidliu 
What is the desired behavior for this signal ?

### DIFF
--- a/Sources/LiveKit/Core/Room.swift
+++ b/Sources/LiveKit/Core/Room.swift
@@ -350,15 +350,7 @@ extension Room: EngineDelegate {
 
         for trackState in trackStates {
             // Try to find participant
-            let participant: Participant?
-
-            if trackState.participantSid == localParticipant?.sid {
-                participant = localParticipant
-            } else {
-                participant = remoteParticipants[trackState.participantSid]
-            }
-
-            if let participant = participant {
+            if let participant = remoteParticipants[trackState.participantSid] {
                 participant.update(state: trackState.state, forTrack: trackState.trackSid)
             }
         }
--- a/Sources/LiveKit/Core/Room.swift
+++ b/Sources/LiveKit/Core/Room.swift
@@ -19,12 +19,21 @@ public class Room: MulticastDelegate<RoomDelegate> {
     //    private let monitorQueue: DispatchQueue
     private var prevPath: NWPath?
     private var lastPathUpdate: TimeInterval = 0
-    internal lazy var engine = Engine(delegate: self)
+    internal lazy var engine = Engine(room: self)
+
+    internal private(set) var connectOptions: ConnectOptions?
+    internal private(set) var roomOptions: RoomOptions?
+
     public var state: ConnectionState {
         engine.connectionState
     }
 
-    init(delegate: RoomDelegate? = nil) {
+    init(delegate: RoomDelegate? = nil,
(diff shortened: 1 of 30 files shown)


boks1971:
@hiroshihorie I think we need to emit (would be good at both room level and participant level) signals about video tracks of a particular remote participant pausing and resuming (this is pause/resume due to bandwidth constraints in the downstream channel).

Interested in thoughts from the Davids :-)

hiroshihorie:
Is this feature turn on by default by the server ?

boks1971:
> Is this feature turn on by default by the server ?

Yes @hiroshihorie .

davidliu:
@boks1971 So to be clear, the end goal of this protocol message simply just be to expose something for devs to know that the stream is currently paused, or is there anything else we would want to do with it? 

This sounds like it would be a separate flag from the `muted` flag that we already have on TrackPublications, am I correct?

boks1971:
> @boks1971 So to be clear, the end goal of this protocol message simply just be to expose something for devs to know that the stream is currently paused, or is there anything else we would want to do with it?

Nope, the goal is to notify/emit events for devs to know so that they can give some treatment to the video element (maybe change frozen video to grayscale. - we used to do this in Slack calls a long time back).

In the future, there may be more in the SDK, but nothing planned for now. So, notification to devs is the only thing to do.

> 
> This sounds like it would be a separate flag from the `muted` flag that we already have on TrackPublications, am I correct?

Yes it is. `muted` is a client initiated thing. It could be
- Publisher muting which essentially becomes `mute` for all subscribers as there is no media track from source
- Subscriber muting which is client initiated.

The `paused` state in this update is purely server side decision. Currently, the only reason this is paused is due to bandwidth constraint. In the future, they may be other business rules or admin controls which may pause without a client initiating a mute.

So, in general, it is good to think of it as
- Client initiated mute/pause
- Server initiated pause
and for every situation ask the question "is it clear to devs why the stream is paused/stopped? i. e. is it clear that it is client side initiated vs server side initiated"

Please let me know if this is all confusing.

Sources/LiveKit/Core/Room.swift
```diff
@@ -345,4 +345,22 @@ extension Room: EngineDelegate {
...
         notify { $0.room(self, didFailToConnect: error) }
     }
+
+    func engine(_ engine: Engine, didUpdate trackStates: [Livekit_StreamStateInfo]) {
+
+        for trackState in trackStates {
+            // Try to find participant
+            let participant: Participant?
+
+            if trackState.participantSid == localParticipant?.sid {
```
boks1971, line 355: You will not get this on local participant.

boks1971 (approved):
Is there no auto generated code for proto changes in ios?

Sources/LiveKit/Core/Room.swift
hiroshihorie: Oh yes so this is fore Remote only

hiroshihorie:
I'm just emitting event and updating `RemoteParticipant.streamState` property. Is that all for now ? 😁

boks1971:
Yeah that's it. I thought @davidliu was also emitting event at room level. But my memory is vague although I reviewed that only yesterday😊

hiroshihorie:
ok thanks onichan 😁🙏

PR #30, opened 2021-12-04
Internal re-structuring attempt for better type inferencing

hiroshihorie:
This is only an attempt until all things are good.


hiroshihorie:
Since true abstract classes don't exist in Swift the main attempt was to
make `Participant`, `TrackPublication`, `Track` etc a `protocol` but that brings
additional complications.... so i will close this for now.

Issue #33, opened 2021-12-10
LocalVideoTrack's source .unknown instead of .camera


doesnotexist:
For my app on iOS (which never uses screen sharing) the firstCameraPublication is always nil.  It appears that source = .unknown for the published LocalVideoTrack associated with the camera capturer.  This breaks the computed property firstCameraPublication on ObservableParticipant, because the filter for the call to first() never finds a track with source == .camera
https://github.com/livekit/client-sdk-ios/blob/main/Sources/LiveKit/SwiftUISupport/ObservableParticipant.swift#L119

Investigating this, I've found that the correct value for source = .camera is passed here:
https://github.com/livekit/client-sdk-ios/blob/main/Sources/LiveKit/Participant/LocalParticipant.swift#L48

However in the next handler in the promise chain, the deserialized trackInfo from the server response results in source = .unknown
https://github.com/livekit/client-sdk-ios/blob/main/Sources/LiveKit/Participant/LocalParticipant.swift#L64

davidzhao:
@doesnotexist what version of the server are you running? I think it's old enough that publishing with Source is not yet supported.

doesnotexist:
@davidzhao v0.13.1
I can see how the no source coming back in the response could result in the .unknown value.
Let me see if upgrading to a newer version fixes the behavior.  
If thats the case, is that the intended behavior, that a client should depend on the server state's values even for metadata about their own published local tracks?

davidzhao:
Server versions before v0.13.5 does not support the Source attribute.. so the `getTrackPublication(source:)` functions will not work properly for versions before then.

Even if it uses the local information (instead of remote), it won't work properly for all of the remote participants' publications. So you really need the right version of server in order to utilize the *Source APIs.

doesnotexist:
Got it, thanks.  Closing this issue.

PR #34, opened 2021-12-10
Fix LocalVideoTrack's source .camera instead of .unknown

doesnotexist:
WIP.  Current band-aid fix resolves issue #33 for my app


doesnotexist:
This is unnecessary for server versions > v13.5 since the response includes accurate source field within the published TrackInfo.

PR #39, opened 2022-01-10
Ensure & wait publisher data channel is `open`

hiroshihorie:
Additional checks

### DIFF
--- a/Sources/LiveKit/Core/Engine.swift
+++ b/Sources/LiveKit/Core/Engine.swift
@@ -211,6 +211,10 @@ class Engine: MulticastDelegate<EngineDelegate> {
             return
         }
 
+        guard publisher.iceConnectionState != .checking {
+            return
+        }
+        
         hasPublished = true
         publisher.negotiate()
     }
--- a/Sources/LiveKit/Core/Engine.swift
+++ b/Sources/LiveKit/Core/Engine.swift
@@ -17,10 +17,10 @@ class Engine: MulticastDelegate<EngineDelegate> {
         subscriberPrimary ? subscriber : publisher
     }
 
-    private(set) var reliableDC: RTCDataChannel?
-    private(set) var lossyDC: RTCDataChannel?
-    private(set) var reliableDCSub: RTCDataChannel?
-    private(set) var lossyDCSub: RTCDataChannel?
+    private(set) var dcReliablePub: RTCDataChannel?
+    private(set) var dcLossyPub: RTCDataChannel?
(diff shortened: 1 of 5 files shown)


Sources/LiveKit/Core/Engine.swift
```diff
@@ -211,23 +211,52 @@ class Engine: MulticastDelegate<EngineDelegate> {
             return
         }
 
+        guard publisher.iceConnectionState != .checking else {
```
davidzhao, line 214: Is this used only for data channels? If so, we probably shouldn't re-negotiate if it's already connected.

Sources/LiveKit/Core/Engine.swift
```diff
@@ -211,23 +211,52 @@ class Engine: MulticastDelegate<EngineDelegate> {
...
+            guard subscriberPrimary,
+                  !publisher.isIceConnected,
+                  publisherDataChannel(for: reliability)?.readyState != .open else {
+                // aleady connected and ready, no-op
+                return Promise(())
+            }
+
+            publisherShouldNegotiate()
+
+            return waitForIceConnect(transport: publisher).then {
```
davidzhao, line 244: FWIW, you could just do `waitForPublisherDataChannelOpen` directly. When dc.readyState changes, it should be good to go.

Sources/LiveKit/Core/Engine.swift
```diff
@@ -236,27 +265,49 @@ class Engine: MulticastDelegate<EngineDelegate> {
...
+            self.add(delegate: engineDelegate!)
+            // detect signal close while waiting
+            var signalDelegate: SignalClientDelegateClosures?
+            signalDelegate = SignalClientDelegateClosures(
+                didClose: { _, _ in
+                    fail(SignalClientError.close("Socket closed while waiting for ice-connect"))
+                    signalDelegate = nil
+                }
+            )
+            self.signalClient.add(delegate: signalDelegate!)
```
davidzhao, line 306: does this get removed automatically?

davidzhao (approved):
lgtm! just a couple of Qs

Sources/LiveKit/Core/Engine.swift
hiroshihorie: This should get cleaned up automatically since it's only weak reference.

Sources/LiveKit/Core/Engine.swift
hiroshihorie: In the previous lines there is logic to prevent re-negotiation if already connected 🙂

PR #42, opened 2022-01-13
Capture client info

hiroshihorie:
LK-502

**macOS**:
`&os=macOS&os_version=12.1&device_model=MacBookPro18,3`


davidzhao (approved):
Looks good! solid support for device model

PR #45, opened 2022-01-25
Use `ParticipantTracks` for `UpdateSubscription`

hiroshihorie:


davidzhao (approved):
looks good!

Sources/LiveKit/Core/Room.swift
```diff
@@ -224,18 +224,24 @@ extension Room {
...
+            Livekit_ParticipantTracks.with {
+                $0.participantSid = participant.sid
+                $0.trackSids = participant.tracks.values
+                    .filter { $0.subscribed != sendUnSub }
+                    .map { $0.sid }
+            }
+        }
+
+        // Backward compatibility
+        let trackSids = participantTracks.map { $0.trackSids }.flatMap { $0 }
```
davidliu, line 237: Is this to send over trackSids for older server versions?

davidliu (approved):

Sources/LiveKit/Core/Room.swift
hiroshihorie: Yup just in case, not sure if needed though 🙂

davidzhao: yeah we should do this for a period of time. since upgrading the server is more painful, some of our users only upgrade every few months.

boks1971: Maybe we can gate this on protocol 6

davidzhao: agreed. we'd just have to send it for pre-protocol6 server

PR #46, opened 2022-01-27
Fix: Subscriber is nil

hiroshihorie:
Fixes LK-611
When server sends offer too quickly after joinResponse, Transports are not fully configured yet
and causes Subscriber is nil error.

Reason
after `waitReceiveJoinResponse`
`configureTransports` was called asynchronously by the Promise chain.
If the server sends offer too quickly, offer arrives between that flow, while subscriber is still nil.


boks1971 (approved):

davidliu:
Not sure if it's still needed here, but Android buffers any messages that comes in before the join response is fully processed, and then re-emits once the engine/room is ready.

hiroshihorie:
That's very nice, I was wondering if I would need to do that or this quick fix is enough.

PR #47, opened 2022-01-29
Fix encoder & renderer crashes

hiroshihorie:
Fixes: LK-400

## Renderer
- Never attempt render frames that are smaller than 8x8 to prevent Metal renderer from crashing. (Even if server sends the frame)

## Capture (Encode)
- Never attempt to encode frames that are smaller than 16x16 (16 was reported in assert message of crash) to prevent h264 encoder from crashing.
- Always target even number dimensions when capturing odd size dimensions

### DIFF
--- a/Sources/LiveKit/Extensions/RTCVideoCapturerDelegate+Buffer.swift
+++ b/Sources/LiveKit/Extensions/RTCVideoCapturerDelegate+Buffer.swift
@@ -1,6 +1,20 @@
 import WebRTC
 import ReplayKit
 
+extension Dimensions {
+
+    // Ensures width and height are even numbers
+    func toEncodeSafeDimensions() -> Dimensions {
+        Dimensions(width: max(encodeSafeSize, width % 2 == 0 ? width : width + 1),
+                   height: max(encodeSafeSize, height % 2 == 0 ? height : height + 1))
+    }
+
+    static func * (a: Dimensions, b: Double) -> Dimensions {
+        Dimensions(width: Int32((Double(a.width) * b).rounded()),
+                   height: Int32((Double(a.height) * b).rounded()))
+    }
+}
+
 extension CGImagePropertyOrientation {
 
     public func toRTCRotation() -> RTCVideoRotation {
@@ -14,15 +28,20 @@ extension CGImagePropertyOrientation {
 }
 
 internal let supportedPixelFormats = DispatchQueue.webRTC.sync { RTCCVPixelBuffer.supportedPixelFormats() }
+internal let renderSafeSize: Int32 = 8
+internal let encodeSafeSize: Int32 = 16
 
 extension RTCVideoCapturerDelegate {
 
-    /// capture a `CVPixelBuffer`
+    public typealias OnTargetDimensions = (Dimensions) -> Void
+
+    /// capture a `CVPixelBuffer`, all other capture methods call this method internally.
     public func capturer(_ capturer: RTCVideoCapturer,
                          didCapture pixelBuffer: CVPixelBuffer,
                          timeStampNs: UInt64,
                          rotation: RTCVideoRotation = ._0,
-                         scale: Double = 1.0) {
+                         scale: Double = 1.0,
+                         onTargetDimensions: OnTargetDimensions? = nil) {
 
         // check if pixel format is supported by WebRTC
         let pixelFormat = CVPixelBufferGetPixelFormatType(pixelBuffer)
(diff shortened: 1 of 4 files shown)


Sources/LiveKit/Extensions/RTCVideoCapturerDelegate+Buffer.swift
```diff
@@ -1,6 +1,20 @@
 import WebRTC
 import ReplayKit
 
+extension Dimensions {
+
+    // Ensures width and height are even numbers
+    internal func toEncodeSafeDimensions() -> Dimensions {
+        Dimensions(width: max(encodeSafeSize, width % 2 == 0 ? width : width + 1),
+                   height: max(encodeSafeSize, height % 2 == 0 ? height : height + 1))
```
boks1971, line 9: Maybe better to reduce one pixel as one pixel of extra blank could be visible and make it look odd.

boks1971 (approved):

Sources/LiveKit/Extensions/RTCVideoCapturerDelegate+Buffer.swift
boks1971: On second thoughts wondering if this should also round down to multiple of 8 to be safe. Have you tried non-multiple of 8?

davidzhao (approved):
this is great! fantastic this would resolve the issue for older versions of the server

Sources/LiveKit/Extensions/RTCVideoCapturerDelegate+Buffer.swift
hiroshihorie: I think the following code should stretch the frame before passing it to the encoder (if required), so there are no black 1 px lines.

It's definitely worth experimenting with multiples of 8, but as of now I haven't encountered a single crash with multiple of 2. 🙂

```swift
            if sourceDimensions != targetDimensions {
                // no adjustments required
                rtcBuffer = RTCCVPixelBuffer(pixelBuffer: pixelBuffer)
            } else {
                // apply adjustments
                rtcBuffer = RTCCVPixelBuffer(pixelBuffer: pixelBuffer,
                                             adaptedWidth: targetDimensions.width,
                                             adaptedHeight: targetDimensions.height,
                                             cropWidth: sourceDimensions.width,
                                             cropHeight: sourceDimensions.height,
                                             cropX: 0,
                                             cropY: 0)
            }

```

davidliu (approved):

Sources/LiveKit/Extensions/RTCVideoCapturerDelegate+Buffer.swift
```diff
@@ -31,33 +50,55 @@ extension RTCVideoCapturerDelegate {
...
 
-            let rtcBuffer = RTCCVPixelBuffer(pixelBuffer: pixelBuffer,
-                                             adaptedWidth: Int32((Double(width) * scale).rounded()),
-                                             adaptedHeight: Int32((Double(height) * scale).rounded()),
-                                             cropWidth: Int32(width),
-                                             cropHeight: Int32(height),
+            let rtcBuffer: RTCCVPixelBuffer
+
+            if sourceDimensions != targetDimensions {
+                // no adjustments required
```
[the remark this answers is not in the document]
hiroshihorie: Omg you're right haha 
Nice find! 👀

Sources/LiveKit/Extensions/RTCVideoCapturerDelegate+Buffer.swift
hiroshihorie: After testing, it feel like power of 8 or even 4 seems to work better 🙂
Is it ok to roundUp vs roundDown ?

boks1971: If you are scaling, I would suggest keeping the round up/round down to a minimum. If mod 2 works and if you have round up or round down by 1, that is better because scaling will not introduce any artefacts. If you are rounding up/down by 4/8 and scaling, the image may be stretched and may look a bit off compared to the capture as it is stretching/shrinking by more.

boks1971: I was thinking about cropping. If you are cropping, it is better to round down to lose one pixel rather than add a pixel and have that be a black/grey/white line which will stand out.

hiroshihorie: Thanks for the info 😁🙏🏻 i will do more experimenting 🙂

PR #48, opened 2022-02-01
Send track settings when reconnect

hiroshihorie:
Should it be the same time as `sendSyncState` ?


boks1971 (approved):
lgtm @hiroshihorie . The update should actually be sent after migration is complete.

davidzhao (approved):
lg!

PR #49, opened 2022-02-08
Room metadata update

hiroshihorie:
The missing room metadata update event 😅


hiroshihorie:
The roomUpdate signal's `Room` object contains a few other info I wonder if we should expose that 🤔

boks1971 (approved):

PR #50, opened 2022-02-14
Enqueue Signal requests while reconnecting

hiroshihorie:
But should we wait to send the queued requests until engine reconnected (Ice connected) ?
or it's ok to just send when signal is connected. 🤔


Sources/LiveKit/Core/SignalClient.swift
```diff
@@ -78,20 +85,53 @@ internal class SignalClient: MulticastDelegate<SignalClientDelegate> {
...
+        // clear the queue
+        queue.removeAll()
+
+        return promises
+    }
+
+    // send request or enqueue while reconnecting
+    func sendRequest(_ request: Livekit_SignalRequest, enqueueIfReconnecting: Bool = true) -> Promise<Void> {
+
+        guard !(connectionState.isReconnecting && enqueueIfReconnecting) else {
```
davidzhao, line 122: I think this also has to enqueue when the queue is not yet empty. This is to ensure that we don't send requests out of order.

davidzhao:
> But should we wait to send the queued requests until engine reconnected (Ice connected) ? or it's ok to just send when signal is connected. 🤔

I think we need to wait until ICE is re-established. we also need to ensure that SyncState can always be sent and not queued.

hiroshihorie:
> > But should we wait to send the queued requests until engine reconnected (Ice connected) ? or it's ok to just send when signal is connected. 🤔
> 
> I think we need to wait until ICE is re-established. we also need to ensure that SyncState can always be sent and not queued.

Got it ! We might need to change the JS SDK too https://github.com/livekit/client-sdk-js/commit/51086e368f3ccb2225a29df636b39d5a42564c24

davidliu (approved):

davidzhao:
> Got it ! We might need to change the JS SDK too [livekit/client-sdk-js@51086e3](https://github.com/livekit/client-sdk-js/commit/51086e368f3ccb2225a29df636b39d5a42564c24)

yeah you are right. will look into it.

Sources/LiveKit/Core/SignalClient.swift
```diff
@@ -401,3 +438,17 @@ internal extension SignalClient {
         return sendRequest(r)
     }
 }
+
+internal extension Livekit_SignalRequest {
+
+    func canQueue() -> Bool {
```
davidzhao, line 444: this is awesome!

Sources/LiveKit/Core/SignalClient.swift
```diff
@@ -84,20 +88,30 @@ internal class SignalClient: MulticastDelegate<SignalClientDelegate> {
...
 
 // MARK: - Private
 
 private extension SignalClient {
 
-    func sendRequest(_ request: Livekit_SignalRequest) -> Promise<Void> {
+    // send request or enqueue while reconnecting
+    func sendRequest(_ request: Livekit_SignalRequest, enqueueIfReconnecting: Bool = true) -> Promise<Void> {
+
+        guard !(connectionState.isReconnecting && request.canQueue() && enqueueIfReconnecting) else {
```
davidzhao, line 102: when reconnection is finished, I think we'd still have to enqueue requests *as long as* the queue is not empty.

this is to ensure ordering is kept with the signal messages. i.e. replacing connectionState.isReconnecting => connectionState.IsReconnecting || !queue.isNotEmpty

davidzhao (approved):
lg! one comment about sending messages before queue is drained.

Sources/LiveKit/Core/SignalClient.swift
hiroshihorie: After reconnection is finished, and sending out the queued requests, clearing the queue wouldn't work ?

davidzhao: yeah it should.. I'm thinking about this case:

1. one thread is flushing the queue after reconnecting, while that's happening
2. user initiates an action that causes a signal event to be sent

would 2 be queued until after the flush is complete? or would it be sent as it comes in?

hiroshihorie: Good point about mutex, i'll make some changes to have peace of mind.

hiroshihorie: I have fixed this so `requestQueue` cannot be modified (promise will wait) while a thread is flushing the queue.

PR #52, opened 2022-02-19
Video Codec preference

hiroshihorie:


davidzhao:
I have a couple of questions about the approach, as ideally the behavior across clients are consistent.

1. Should this be a global setting, or a track-level setting? WebRTC defines codec preferences to be at the track level via [setCodecPreferences](https://developer.mozilla.org/en-US/docs/Web/API/RTCRtpTransceiver/setCodecPreferences).
2. How does this fit in with the client configuration? Are they achieving the same purpose? (i.e. limiting clients from using certain codecs?)

I think there's probably a few different things at play.

1. Which codecs can the current device support? - WebRTC defines the base encoders/decoders, so it should know some of this. But it's possible that it's getting it wrong (we've seen Safari crash with h.264, even though it should support it)
2. Which codecs/profile levels can the server support? - Server is defining this set, and WebRTC negotiations will determine it
3. Which codec does the user want to use for the session and for a particular track?

hiroshihorie:
Unfortunately, [setCodecPreferences](https://developer.mozilla.org/en-US/docs/Web/API/RTCRtpTransceiver/setCodecPreferences) doesn't seem to be implemented for iOS WebRTC SDK.

davidzhao:
> Unfortunately, [setCodecPreferences](https://developer.mozilla.org/en-US/docs/Web/API/RTCRtpTransceiver/setCodecPreferences) doesn't seem to be implemented for iOS WebRTC SDK.

We can handle this somewhat on the server side, by removing other codecs from the answer. AddTrackRequest could indicate the codec that is desired for that track.

hiroshihorie:
> We can handle this somewhat on the server side, by removing other codecs from the answer. AddTrackRequest could indicate the codec that is desired for that track.

Ah, that would be a much better non-hacky way.

PR #53, opened 2022-02-19
Update computations

hiroshihorie:
from PRs 👉
https://github.com/livekit/client-sdk-js/pull/146
https://github.com/livekit/client-sdk-js/pull/149

But I think dimensions should be compared by area (width * height) instead of the largest side (max)
👉 https://github.com/livekit/client-sdk-js/pull/146/files#diff-4c7fc41e072f8a2f5272ccae3b53d650e87c64d6f4b74541343e28fbb0a7100aR180-R195

for example comparing by max would be
1000 x 300 > 900 x 900
but comparing the area would be
300,000 < 810,000

Just want to check if this is intended behavior 🙂

### DIFF
--- a/Sources/LiveKit/Support/Utils.swift
+++ b/Sources/LiveKit/Support/Utils.swift
@@ -188,28 +188,28 @@ internal class Utils {
     ) -> [RTCRtpEncodingParameters] {
 
         let publishOptions = publishOptions ?? VideoPublishOptions()
-        let useSimulcast: Bool = !isScreenShare && publishOptions.simulcast
-        let encoding: VideoEncoding? = isScreenShare ? publishOptions.screenShareEncoding : publishOptions.encoding
-
-        // get suggested presets for the dimensions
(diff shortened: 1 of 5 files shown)


lukasIO:
@hiroshihorie that's a good point about sorting by area!
The reason I went with comparing by max, was that the `scaleResolutionDownBy` parameter of the RDPEndoncdingParameters is being computed similarily https://github.com/livekit/client-sdk-js/blob/main/src/room/participant/publishUtils.ts#L189 (though this is using `min`, so my `max` is still buggy. 

When implementing it, I also wondered if just sorting my `maxBitrate` would suffice. 
I have the feeling that providing two layer presets where, for the second one, the area is smaller, but the bitrate is higher compared to the first one, should not be possible?

lukasIO (approved):

hiroshihorie:
Yes the comparison priority order is Dimensions, Bitrate, FPS.... 
Not so great when 2 Dimensions are very near but bitrates are very different.... but i guess it'll work for now..

PR #55, opened 2022-02-28
Simulcast screen share by default

hiroshihorie:


Sources/LiveKit/Types/VideoParameters.swift
```diff
@@ -38,6 +38,38 @@ public struct VideoParameters {
...
+internal extension VideoParameters {
+
+    func defaultScreenShareSimulcastLayers() -> [VideoParameters] {
+
+        struct Layer {
+            let scaleDownBy: Double
+            let fps: Int
+        }
+
+        let layers = [Layer(scaleDownBy: 2, fps: 3)]
```
davidzhao, line 52: how does it work with a low fps like 3? is it sufficient?

davidzhao (approved):

lukasIO (approved):

Sources/LiveKit/Types/VideoParameters.swift
hiroshihorie: Pretty low fps but it does work haha, this is same with JS SDK you can try it.

PR #57, opened 2022-03-04
Suspend local video tracks in background

hiroshihorie:
This will mute local video tracks when the app goes background and unmutes when back to the foreground.
Would like to hear opinions 🙂

Without this, when the app goes background the video will appear to be stuck to the other participants.

### DIFF
--- a/Sources/LiveKit/Core/Room.swift
+++ b/Sources/LiveKit/Core/Room.swift
@@ -559,10 +559,7 @@ extension Room: AppStateDelegate {
         guard options.suspendLocalVideoTracksInBackground else { return }
 
         guard let localParticipant = localParticipant else { return }
-        let promises = localParticipant.videoTracks.values
-            .compactMap { $0 as? LocalTrackPublication }
-            .filter { $0.kind == .video }
-            .map { $0.suspend() }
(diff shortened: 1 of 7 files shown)


Sources/LiveKit/Core/Room.swift
```diff
@@ -546,3 +549,40 @@ extension Room: EngineDelegate {
...
+// MARK: - AppStateDelegate
+
+extension Room: AppStateDelegate {
+
+    func appDidEnterBackground() {
+
+        guard options.suspendLocalVideoTracksInBackground else { return }
+
+        guard let localParticipant = localParticipant else { return }
+        let promises = localParticipant.videoTracks.values
```
davidzhao, line 562: if we do these frequently, I wonder if we can make a helper function localParticipant.localVideoTracks that performs these filtering? or if `videoTracks` on localParticipant be changed to return the right types?

davidzhao (approved):
awesome! this makes a ton of sense.

lukasIO (approved):

Sources/LiveKit/Core/Room.swift
hiroshihorie: yup that would make it cleaner 🙂

PR #59, opened 2022-03-06
Quick-reconnect when user switches network

hiroshihorie:
Previously, switching network doesn't initiate a quick-reconnect (or the user needs to wait until PeerConnectionState becomes `failed`)

This change instantly resumes the session by quick-reconnecting to improve user experience.


Sources/LiveKit/Extensions/TimeInterval.swift
```diff
@@ -3,7 +3,8 @@ import Foundation
 /// Default timeout `TimeInterval`s used throughout the SDK.
 internal extension TimeInterval {
     static let captureStart: Self = 5
-    static let defaultConnect: Self = 10
+    static let defaultConnect: Self = 5
```
boks1971, line 6: Is this the connect timeout? We have seen that this is not enough on long delay networks? We are using 15 seconds in JS SDK for connection timeout.

Sources/LiveKit/Support/ConnectivityListener.swift
```diff
@@ -0,0 +1,70 @@
...
+        }
+    }
+
+    public private(set) var path: NWPath {
+        didSet {
+            defer { self.hasConnectivity = path.hasConnectivity() }
+
+            if oldValue.hasConnectivity(), path.hasConnectivity(),
+               // first interface changed
+               (oldValue.availableInterfaces.first != path.availableInterfaces.first)
```
boks1971, line 40: Is `first` the active one always?

Sources/LiveKit/Support/ConnectivityListener.swift
```diff
@@ -0,0 +1,70 @@
...
+                // same interface but gateways changed (detect wifi network switch)
+                || (oldValue.gateways != path.gateways) {
```
boks1971, line 42: Is this a `route` change? When does this happen?

Is it possible to detect IP address change? I guess if it is NATted it is not possible. To keep it simple, I guess letting any IP address change take the ICE failed route is reasonable.

boks1971 (approved):
lgtm! but I barely under `swift` code :-)

davidzhao (approved):
it's great to have this!

Sources/LiveKit/Support/ConnectivityListener.swift
hiroshihorie: https://developer.apple.com/documentation/network/nwpath/2998720-availableinterfaces
"in order of preference" 

yes it should be !

PR #63, opened 2022-03-17
Keep mic on until unpublish

hiroshihorie:


Sources/LiveKit/Track/Local/LocalAudioTrack.swift
```diff
@@ -42,15 +42,15 @@ public class LocalAudioTrack: LocalTrack, AudioTrack {
     }
 
     @discardableResult
-    override func start() -> Promise<Void> {
-        super.start().then(on: .sdk) {
+    override func publish() -> Promise<Void> {
```
davidzhao, line 45: is this a function that's exposed to users?

hiroshihorie: no, this is internal only

davidzhao (approved):
👍

PR #64, opened 2022-03-20
Fix: Always include dimensions or quality in `UpdateTrackSettings` when enabling

hiroshihorie:


davidzhao (commented):
This change itself looks good. But I'm not confident that we've found the root cause. The behavior that we had seen was:

1. publisher muted their video
2. we sent a disabled without dimensions --> this is correct
3. publisher unmuted their video
4. we sent an enabled event without dimensions --> this is wrong

It seems that the only case it could happen is if `videoViewVisibilities.values.largestVideoViewSize` returns 0? Is that what's actually happening here? how could that happen?

hiroshihorie:
Oh, thats weird, I couldn't repro that. 🤔
So it's ok no dimensions/quality for **disable** ?

davidzhao:
> Oh, thats weird, I couldn't repro that. 🤔 So it's ok no dimensions/quality for **disable** ?

That's what the logs indicate... No, dimension/quality gets ignored when `disabled: true`.

hiroshihorie:
I think triggering compute on `track(_ track: VideoTrack, didAttach videoView: VideoView)` was not a good idea, i removed this. dimensions are unknown at this point VideoView initialized with .zero

davidzhao:
> I think triggering compute on `track(_ track: VideoTrack, didAttach videoView: VideoView)` was not a good idea, i removed this. dimensions are unknown at this point VideoView initialized with .zero

that sounds like what's happening given Raja's report!

davidzhao (approved):

PR #66, opened 2022-03-23
Suspend remote video tracks in background

hiroshihorie:
Also suspend all remote video in background.

This change sends `UpdateTrackSettings(disable: true)` to server
or
should it be a simple `mediaStreamTrack.enabled = false` ?
(But server will still send video frames in this case)

### DIFF
--- a/Sources/LiveKit/Publications/RemoteTrackPublication.swift
+++ b/Sources/LiveKit/Publications/RemoteTrackPublication.swift
@@ -25,8 +25,6 @@ public enum SubscriptionState {
 }
 
 public class RemoteTrackPublication: TrackPublication {
-    // user's preference to subscribe or not
-    private var preferSubscribed: Bool?
 
     public private(set) var subscriptionAllowed = true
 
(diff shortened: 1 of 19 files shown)


Sources/LiveKit/Publications/RemoteTrackPublication.swift
```diff
@@ -197,6 +197,28 @@ public class RemoteTrackPublication: TrackPublication {
...
+        return send(trackSettings: newSettings).then {
+            self.suspended = true
+        }
+    }
+
+    @discardableResult
+    internal override func resume() -> Promise<Void> {
+        // do nothing if was not suspended
+        guard suspended else { return Promise(()) }
+        let newSettings = trackSettings.copyWith(enabled: true)
```
davidzhao, line 217: what if it was disabled by the user, would this now be enabling it?

Sources/LiveKit/Publications/RemoteTrackPublication.swift
```diff
@@ -197,6 +197,28 @@ public class RemoteTrackPublication: TrackPublication {
...
+        return send(trackSettings: newSettings).then {
+            self.suspended = false
```
davidzhao, line 219: seems that this should be set regardless of if the request succeeds?

Sources/LiveKit/Publications/RemoteTrackPublication.swift
hiroshihorie: if disabled by user, `suspended` will be false so nothing should happen.

hiroshihorie: going into background `suspend()` gets called and back to foreground `resume()` gets called but
if user does some own logic it might conflict. So maybe this can be an option.

Sources/LiveKit/Publications/RemoteTrackPublication.swift
hiroshihorie: yes that might be better!

davidliu (approved):
LGTM!

hiroshihorie:
A lot of state management has changed since last updated , i think a new PR is better.

PR #68, opened 2022-03-25
VideoView: Stop rendering when hidden

hiroshihorie:
`VideoView.isHidden = true` will stop rendering
`VideoView.isHidden = false` will resume rendering

This will be useful for optimization in CollectionViews etc.


davidzhao (approved):
fantastic work! this should improve adaptive stream compatibility significantly.

PR #69, opened 2022-03-28
AdaptiveStream redesign

hiroshihorie:
Simplify logic and improve reliability

Relying on async delegates may cause state mismatch so instead 
a timer will scan all related VideoViews in this design.

### DIFF
--- a/Sources/LiveKit/Publications/RemoteTrackPublication.swift
+++ b/Sources/LiveKit/Publications/RemoteTrackPublication.swift
@@ -186,10 +186,11 @@ public class RemoteTrackPublication: TrackPublication {
                                didLayout size: CGSize) {
 
         DispatchQueue.main.async { [weak self] in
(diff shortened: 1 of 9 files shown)


Sources/LiveKit/Publications/RemoteTrackPublication.swift
```diff
@@ -63,20 +45,37 @@ public class RemoteTrackPublication: TrackPublication {
...
+
+    private var metadataMuted: Bool = false
+
+    private var trackSettings = TrackSettings() {
+        didSet {
+            guard oldValue != trackSettings else { return }
+            log("did update trackSettings: \(trackSettings)")
+            // TODO: emit event when trackSettings has been updated by adaptiveStream.
+        }
+    }
```
hiroshihorie, line 59: I just moved these code

davidzhao (approved):
this looks good! seems more reliable with a timer.

PR #73, opened 2022-04-05
VideoView optimizations

hiroshihorie:
* Internal renderer hidden until first frame is rendered.
* Quickly render previously cached frame when attached.

### DIFF
--- a/Sources/LiveKit/Views/VideoView.swift
+++ b/Sources/LiveKit/Views/VideoView.swift
@@ -302,7 +302,7 @@ private extension VideoView {
             log("Renderer: attaching...")
 
             if let frame = track.videoFrame {
-                log("rendering cached frame")
+                log("rendering cached frame \(frame.hashValue)")
                 renderFrame(frame)
             }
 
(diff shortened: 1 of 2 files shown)


Sources/LiveKit/Views/VideoView.swift
```diff
@@ -303,10 +300,19 @@ private extension VideoView {
 
         if shouldAttach {
             log("Renderer: attaching...")
+
+            if let frame = track.videoFrame {
+                log("rendering cached frame")
+                renderFrame(frame)
```
boks1971, line 306: Will this cause a double effect kind of thing on re-attach? Maybe, it is blank. So, when scrolled into view, it goes from blank -> cached frame (frozen video of last frame) -> jumps ahead to live video?

boks1971 (approved):

Sources/LiveKit/Views/VideoView.swift
hiroshihorie: logically, that could happen, but in reality it seems to be instantaneously rendered.

https://user-images.githubusercontent.com/548776/161813898-aadd1ef2-700a-4f1e-90d4-3ad452583a3e.mov

PR #75, opened 2022-04-11
Simplify wait logic

hiroshihorie:
Use the good old completer pattern to simplify wait logic (and improve reliability)


boks1971 (approved):
Wow, a lot of deleted code, always a good sign :-)

PR #76, opened 2022-04-13
Thread safe design for internal states

hiroshihorie:
new thread safe design for internal state


Sources/LiveKit/Core/Engine.swift
```diff
@@ -136,8 +140,6 @@ internal class Engine: MulticastDelegate<EngineDelegate> {
             completer.reset()
         }
 
-        connectStopwatch.clear()
```
boks1971, line 139: No need to clear this in the new `state` struct?

boks1971 (approved):

Sources/LiveKit/Core/Engine.swift
hiroshihorie: yes should be ok 🙂

PR #78, opened 2022-04-18
iOS simulator rendering via RTCEAGLVideoView

iAmcR:
@davidzhao As previously discussed. Please see PR to enable simulator video rendering.

### DIFF
--- a/Sources/LiveKit/Views/VideoView.swift
+++ b/Sources/LiveKit/Views/VideoView.swift
@@ -428,19 +428,27 @@ extension VideoView {
 
     internal static func createNativeRendererView() -> NativeRendererView {
         let result: NativeRendererView
-        #if os(iOS)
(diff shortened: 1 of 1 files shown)


hiroshihorie:
This uses OpenGL, which is deprecated and probably will emit warnings.
The root cause of the issue is that Metal rendering is not working despite being supported since Xcode 11.

But I think we can use this workaround until the root cause is fixed.

hiroshihorie (approved):

iAmcR:
> This uses OpenGL, which is deprecated and probably will emit warnings. The root cause of the issue is that Metal rendering is not working despite being supported since Xcode 11.
> 
> But I think we can use this workaround until the root cause is fixed.

Agreed. That's what's weird about it. It seems WebRTC for iOS is at odds in certain places.

PR #90, opened 2022-05-20
VideoView: detect rendering / freeze

hiroshihorie:
Useful for developers to update UI when frames are not arriving in bad network etc.
Currently there's no way to know if frames are arriving / rendering.

I think @figelwump was suggesting this feature.

Now, (no render) freeze detection is hardcoded to 2.0 seconds.

Simulating bad network:
https://user-images.githubusercontent.com/548776/169492532-2eb7397f-e2ca-490f-9458-34a5227e5d5c.mov

### DIFF
--- a/Sources/LiveKit/Support/StateSync.swift
+++ b/Sources/LiveKit/Support/StateSync.swift
@@ -22,8 +22,8 @@ internal typealias OnStateMutate<Value> = (_ state: Value, _ oldState: Value) ->
 internal final class StateSync<Value> {
 
     // use concurrent queue to allow multiple reads and block writes with barrier.
-    private let queue = DispatchQueue(label: "LiveKitSDK.state", qos: .default,
-                                      attributes: [.concurrent])
+    let queue = DispatchQueue(label: "LiveKitSDK.state", qos: .default,
+                              attributes: [.concurrent])
 
     // actual value
     private var _value: Value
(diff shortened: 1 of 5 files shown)


Sources/LiveKit/Views/VideoView.swift
```diff
@@ -203,9 +201,18 @@ public class VideoView: NativeView, Loggable {
...
-                track.notify { $0.track(track, videoView: self, didUpdate: state.renderState) }
+            // isRendering updated
+            if state.isRendering != oldState.isRendering, let track = state.track {
+
+                self.log("isRendering \(oldState.isRendering) -> \(state.isRendering)")
+
+                if state.isRendering {
+                    self._renderTimer.restart()
+                } else {
+                    self._renderTimer.suspend()
```
boks1971, line 212: Is this an irreversible thing? i. e. once `isRendering` goes to `false`, will it not change back and restart the timer?

Sources/LiveKit/Support/StateSync.swift
```diff
@@ -22,8 +22,8 @@ internal typealias OnStateMutate<Value> = (_ state: Value, _ oldState: Value) ->
 internal final class StateSync<Value> {
 
     // use concurrent queue to allow multiple reads and block writes with barrier.
-    private let queue = DispatchQueue(label: "LiveKitSDK.state", qos: .default,
-                                      attributes: [.concurrent])
+    let queue = DispatchQueue(label: "LiveKitSDK.state", qos: .default,
+                              attributes: [.concurrent])
```
boks1971, line 26: This cannot be private any more?

hiroshihorie: Oops that can remain private for now 🙂

Sources/LiveKit/Views/VideoView.swift
hiroshihorie: It's reversible because when a frame arrives state.isRendering will be true.

boks1971 (approved):

hiroshihorie:
I think I need to move a few events from `TrackDelegate` to `VideoViewDelegate`.

PR #102, opened 2022-06-24
Protocol 8

hiroshihorie:
use v7 or lower when using deprecated connect method

### DIFF
--- a/Sources/LiveKit/LiveKit.swift
+++ b/Sources/LiveKit/LiveKit.swift
@@ -35,6 +35,7 @@ public class LiveKit {
 
     public static let version = "1.0.0"
 
+    @available(*, deprecated, message: "Use Room.connect() instead")
     public static func connect(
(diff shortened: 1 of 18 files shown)


Sources/LiveKit/LiveKit.swift
```diff
@@ -35,6 +35,7 @@ public class LiveKit {
 
     public static let version = "1.0.0"
 
+    @available(*, deprecated, message: "Use Room.connect() instead, latest protocol version is not supported with this method.")
```
boks1971, line 38: :nit: `Use Room.connect() instead, protocol v8 and higher do not support this method`?

boks1971 (approved):

Sources/LiveKit/LiveKit.swift
hiroshihorie: got it !

PR #104, opened 2022-06-28
Fix missing TrackPublished events

hiroshihorie:
https://github.com/livekit/client-sdk-js/pull/275

### DIFF
--- a/Sources/LiveKit/Core/Engine.swift
+++ b/Sources/LiveKit/Core/Engine.swift
@@ -96,7 +96,7 @@ internal class Engine: MulticastDelegate<EngineDelegate> {
                     self.executeQueuedBlocks()
                 } else {
                     // make sure pending blocks are cleared
-                    self.clearQueuedBlock()
+                    self.clearQueuedBlocks()
                 }
             }
 
@@ -249,7 +249,7 @@ internal extension Engine {
         }
     }
 
-    private func clearQueuedBlock() {
+    private func clearQueuedBlocks() {
 
         _blockProcessQueue.async { [weak self] in
             guard let self = self else { return }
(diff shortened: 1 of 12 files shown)


davidzhao:
Would this PR cause all of the initial updates *before* a room is connected to fire as changed events? I think that's actually not the desired behavior.

The spec indicates that TrackPublished events should only fire *after* the participant is fully connected to the room. It should not queue up tracks published before the participant is connected. This is also true for the speaker changed/etc events.

The JS PR fixes a different issue, which is regarding when other participants are connected to the room *after* the local participant is already in the room. If we received that first update when other participants already have published tracks, it would not fire TrackPublished.

hiroshihorie:
👌 I misunderstood the JS PR.

Sources/LiveKit/Core/Engine.swift
```diff
@@ -647,7 +719,14 @@ extension Engine: TransportDelegate {
     func transport(_ transport: Transport, didAdd track: RTCMediaStreamTrack, streams: [RTCMediaStream]) {
         log("did add track")
         if transport.target == .subscriber {
-            notify { $0.engine(self, didAdd: track, streams: streams) }
+
+            // execute block when connected
+            execute(when: { state, _ in state.connectionState == .connected },
```
davidzhao, line 724: this is really neat!

Sources/LiveKit/Participant/RemoteParticipant.swift
```diff
@@ -57,13 +57,16 @@ public class RemoteParticipant: Participant {
             validTrackPublications[trackInfo.sid] = publication!
         }
 
-        if hadInfo {
-            // ensure we are updating only tracks published since joining
-            for publication in newTrackPublications.values {
-                notify(label: { "participant.didPublish \(publication)" }) {
+        // always emit events for new publications, Room will not forward them unless it's ready
```
davidzhao, line 60: nit: this comment does not apply here.

davidzhao (approved):

PR #106, opened 2022-07-07
Allow multiple `VideoView`s to be attached on a single `VideoTrack`

hiroshihorie:
Making sure it doesn't break AdaptiveStream.

Can't remember exactly why I disabled this in the first place.
Maybe in the process of debugging Leap's issues 🤔


Sources/LiveKit/Track/Track.swift
```diff
@@ -17,30 +17,6 @@
...
-    mutating func add(weakElement: VideoView) {
-        guard !contains(weakElement: weakElement) else { return }
-        append(WeakContainer(weakObject: weakElement))
-    }
-
-    mutating func remove(weakElement: VideoView) {
-        removeAll { $0.weakObject == weakElement }
-    }
-}
-
```
hiroshihorie, line 43: Not really required anymore, will just use `NSHashTable<T>.weakObjects()`.

PR #110, opened 2022-07-18
Cocoapods

hiroshihorie:


LiveKit.podspec
```diff
@@ -0,0 +1,29 @@
...
+  spec.summary      = 'LiveKit Swift Client SDK. Easily build live audio or video experiences into your mobile app, game or website.'
+  spec.homepage     = 'https://github.com/livekit/client-sdk-swift'
+  spec.license      = { :type => 'Apache 2.0', :file => 'LICENSE' }
+  spec.author       = 'LiveKit'
+
+  spec.ios.deployment_target = '13.0'
+  spec.osx.deployment_target = '10.15'
+
+  spec.swift_versions = ['4.2', '5']
+  spec.source         = { :git => 'https://github.com/livekit/client-sdk-swift.git', :branch => 'main' }
```
cloudwebrtc, line 15: >`:branch => 'main' `

Do we need to use fixed tags when we officially publish to Cocoapods repo?
For example, when we release 1.0.2, we will create 1.0.2 in tags

cloudwebrtc (approved):

LiveKit.podspec
hiroshihorie: yup

PR #118, opened 2022-08-11
Use H264 42e032 codec for macOS

hiroshihorie:


Sources/LiveKit/Core/Engine.swift
```diff
@@ -755,6 +755,50 @@ extension Engine: ConnectivityListenerDelegate {
...
+        }
+
+        // create a new H264 codec with new profileLevelId
+        let newH264 = RTCVideoCodecInfo(name: kRTCH264CodecName,
+                                        parameters: ["profile-level-id": profileLevelId.hexString,
+                                                     "level-asymmetry-allowed": "1",
+                                                     "packetization-mode": "1"])
+
+        // swap the h264 codec
+        let codecs = map { $0.name == kRTCVideoCodecH264Name ? newH264 : $0 }
```
cnderrauber, line 775: Does this line means replace all h264 codec profile to 42e032? is this change will affect iOS devices?

cnderrauber (approved):

Sources/LiveKit/Core/Engine.swift
cnderrauber: if it is, we might need some device compatible test

hiroshihorie: Yes it will replace all for iOS/macOS

cnderrauber (approved):

PR #122, opened 2022-08-24
Ability to override `VideoView`'s rotation (iOS)

hiroshihorie:
Only works for iOS at the moment.

macOS can be supported after https://github.com/webrtc-sdk/webrtc/pull/40 gets merged.


davidzhao (approved):
lgtm! CC @davidliu too as a potential thing for Android

PR #124, opened 2022-08-25
DispatchQueue improvements

hiroshihorie:
Using common static .sdk queue was a bad idea, queue gets too congested when multiple Room/Engine objects.
Concurrent Room connecting works well now.

### DIFF
--- a/Sources/LiveKit/Broadcast/BroadcastScreenCapturer.swift
+++ b/Sources/LiveKit/Broadcast/BroadcastScreenCapturer.swift
@@ -20,7 +20,8 @@ class BroadcastScreenCapturer: BufferCapturer {
     var frameReader: SocketConnectionFrameReader?
 
     override func startCapture() -> Promise<Bool> {
-        super.startCapture().then(on: .sdk) {didStart -> Promise<Bool> in
+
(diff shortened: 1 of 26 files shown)


davidzhao (approved):
lg! should room/engine/signalclient each uses their own dispatch queue? should they share in some cases?

hiroshihorie:
Good point, should be ok because all state is managed through the **state lock** regardless of the thread the promise executes on https://github.com/livekit/client-sdk-swift/blob/main/Sources/LiveKit/Support/StateSync.swift .
But I will double check and be more confident if I write more e2e tests.

hiroshihorie:
Properties **not** protected by the `StateSync` https://github.com/livekit/client-sdk-swift/blob/main/Sources/LiveKit/Support/StateSync.swift should be safe as long as It is private (won't be accessed from another queue/thread).
I think 😅

davidzhao:
> Properties **not** protected by the `StateSync` https://github.com/livekit/client-sdk-swift/blob/main/Sources/LiveKit/Support/StateSync.swift should be safe as long as It is private (won't be accessed from another queue/thread). I think 😅

got it, thanks!

PR #128, opened 2022-09-02
ObjC Support

hiroshihorie:
Following Swift features are not support and needs re-write :
- enums with associated values
- structs
- optional values of primitive types
- Inheriting generic classes

### TODO
- [x] Make sure all ObjC method / protocol signatures looks correct
- [x] ObjC Helper for all Promises
- [x] Override `isEqual`

### DIFF
--- a/Sources/LiveKit/Core/Room.swift
+++ b/Sources/LiveKit/Core/Room.swift
@@ -19,7 +19,8 @@ import Network
 import Promises
 import WebRTC
 
-@objc public class Room: NSObject, MulticastDelegateCapable, Loggable {
+@objc
+public class Room: NSObject, MulticastDelegateCapable, Loggable {
 
     // MARK: - MulticastDelegate
 
@@ -30,19 +31,30 @@ import WebRTC
 
     // MARK: - Public
 
-    @objc public var sid: Sid? { _state.sid }
-    @objc public var name: String? { _state.name }
-    @objc public var metadata: String? { _state.metadata }
-    @objc public var serverVersion: String? { _state.serverVersion }
-    @objc public var serverRegion: String? { _state.serverRegion }
+    @objc
+    public var sid: Sid? { _state.sid }
+
+    @objc
+    public var name: String? { _state.name }
+
+    @objc
+    public var metadata: String? { _state.metadata }
+
+    @objc
+    public var serverVersion: String? { _state.serverVersion }
+
+    @objc
+    public var serverRegion: String? { _state.serverRegion }
 
     public var localParticipant: LocalParticipant? { _state.localParticipant }
     public var remoteParticipants: [Sid: RemoteParticipant] { _state.remoteParticipants }
     public var activeSpeakers: [Participant] { _state.activeSpeakers }
 
     // expose engine's vars
-    @objc public var url: String? { engine._state.url }
-    @objc public var token: String? { engine._state.token }
+    @objc
+    public var url: String? { engine._state.url }
+    @objc
+    public var token: String? { engine._state.token }
 
     public var connectionState: ConnectionState { engine._state.connectionState }
     public var connectStopwatch: Stopwatch { engine._state.connectStopwatch }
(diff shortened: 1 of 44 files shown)


piercifani:
Instead of changing the current implementation, couldn't the ObjC wrapper sit on top of the current Swift code? That way we don't lose all the Swift features we now take for granted and ObjC clients can still integrate with it

hiroshihorie:
Hi @piercifani , I'm being extra careful not to break current API so we can keep all Swift features.

For example for ObjC, there is a new `connectionStateObjC` property but this will be renamed to `connectionState` when compiled for ObjC. This will use the new ObjC compatible `ConnectionStateObjC` enum.
https://github.com/livekit/client-sdk-swift/pull/128/files#diff-78980460aeafb35975431b7cd2aa713d7c9e40b3bb56e58b31b40b48fd1ddef6R66-R67

From Swift, there will be 2 properties visible `connectionState` and `connectionStateObjC` which isn't great but at least this approach doesn't break the current `connectionState` property. (Still looking if I can hide `connectionStateObjC` when compiled for Swift)

So the only downside for this PR would be new `-ObjC` post fixed properties / methods for ObjC compatibility.

Did you find something in particular that breaks ?

Sources/LiveKit/Core/Room.swift
```diff
@@ -19,26 +19,53 @@ import Network
...
+    @objc
     public var url: String? { engine._state.url }
+
+    @objc
     public var token: String? { engine._state.token }
+
     public var connectionState: ConnectionState { engine._state.connectionState }
+
+    @objc(connectionState)
+    public var connectionStateObjC: ConnectionStateObjC { engine._state.connectionState.toObjCType() }
```
hiroshihorie, line 67: Hi @piercifani , I'm being extra careful not to break current API so we can keep all Swift features.

For example for ObjC, there is a new `connectionStateObjC` property but this will be renamed to `connectionState` when compiled for ObjC. This will use the new ObjC compatible `ConnectionStateObjC` enum.
https://github.com/livekit/client-sdk-swift/pull/128/files#diff-78980460aeafb35975431b7cd2aa713d7c9e40b3bb56e58b31b40b48fd1ddef6R66-R67

From Swift, there will be 2 properties visible `connectionState` and `connectionStateObjC` which isn't great but at least this approach doesn't break the current `connectionState` property. (Still looking if I can hide `connectionStateObjC` when compiled for Swift)

So the only downside for this PR would be new `-ObjC` post fixed properties / methods for ObjC compatibility.

Did you find something in particular that breaks ?

Sources/LiveKit/Core/Room.swift
```diff
@@ -19,26 +19,55 @@ import Network
...
+
+    @objc
     public var token: String? { engine._state.token }
+
     public var connectionState: ConnectionState { engine._state.connectionState }
+
+    /// Only for Objective-C.
+    @objc(connectionState)
+    @available(swift, obsoleted: 1.0)
+    public var connectionStateObjC: ConnectionStateObjC { engine._state.connectionState.toObjCType() }
```
hiroshihorie, line 69: Hi @piercifani , I'm being extra careful not to break current API so we can keep all Swift features.

For example for ObjC, there is a new `connectionStateObjC` property but this will be renamed to `connectionState` when compiled for ObjC. This will use the new ObjC compatible `ConnectionStateObjC` enum.

From Swift, there will be 2 properties visible `connectionState` and `connectionStateObjC` which isn't great but at least this approach doesn't break the current `connectionState` property. Still looking if I can hide `connectionStateObjC` completely when compiled for Swift, but for now I can mark it `@available(swift, obsoleted: 1.0)` so Swift users will not accidentally use it (causes compiler error).

So the only downside for this PR would be new `-ObjC` post fixed properties / methods for ObjC compatibility.

Did you find something in particular that breaks ?

cloudwebrtc (approved):
lgtm

PR #133, opened 2022-09-08
Revamp README.md

dsa:
Draft pass at an updated readme, giving a bit more structure and reordering some of the information. There are a few areas that need to be fleshed out with more content, but seems like this may be a good starting foundation we can expand upon further.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,22 +1,24 @@
-# iOS/macOS Swift SDK for LiveKit
+# Swift Multiplatform (iOS and macOS) SDK for LiveKit
 
-Official Client SDK for [LiveKit](https://github.com/livekit/livekit-server).
-Easily add video & audio capabilities to your iOS/macOS apps.
+This is the official LiveKit Swift client SDK. 
 
-## Docs & Example app
+Together with LiveKit [server](https://github.com/livekit/livekit-server), you can easily add real-time video/audio/data capabilities to your iOS or macOS apps.
 
-Docs and guides are at [https://docs.livekit.io](https://docs.livekit.io).
+<insert gif/image here />
 
-There is full source code of a [iOS/macOS Swift UI Example App](https://github.com/livekit/client-example-swift).
-
-For minimal examples view this repo 👉 [Swift SDK Examples](https://github.com/livekit/client-example-collection-swift)
+## Features
 
 ## Installation
 
-LiveKit for Swift is available as a Swift Package.
+### Requirements
 
-### Package.swift
+### Swift Package Manager
 
+#### XCode
+Go to Project Settings -> Swift Packages.
+Add a new package and enter: `https://github.com/livekit/client-sdk-swift`
+
+#### Package.swift
 Add the dependency and also to your target
 
 ```swift title="Package.swift"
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,22 +1,24 @@
...
-
-For minimal examples view this repo 👉 [Swift SDK Examples](https://github.com/livekit/client-example-collection-swift)
+## Features
 
 ## Installation
 
-LiveKit for Swift is available as a Swift Package.
+### Requirements
+
+### Swift Package Manager
```
davidzhao, line 15: we now support Cocoapods too, would be great to document steps here.

README.md
```diff
@@ -107,90 +104,22 @@ extension RoomViewController: RoomDelegate {
...
-Since `VideoView` is a UI component, all operations (read/write properties etc) must be performed from the `main` thread.
+## Documentation
 
-Other core classes can be accessed from any thread.
+LiveKit docs, guides and sample apps are available at [https://docs.livekit.io](https://docs.livekit.io). We also have minimal examples to get you started with Swift, UIKit, or Objective-C here 👉 [Starter Kits](https://github.com/livekit/client-example-collection-swift).
 
-Delegates will be called on the SDK's internal thread.
-Make sure any access to the UI is within the main thread, for example by using `DispatchQueue.main.async`.
-
-# Memory management
```
davidzhao, line 123: I think these are quite useful, and I would place it further down in the details section.

Along with this, it would be good to have a section on how to keep the session running in the background (i.e. project settings, publish audio)

README.md
```diff
@@ -34,16 +36,11 @@ let package = Package(
...
+## Quickstart
 
-Go to Project Settings -> Swift Packages.
-
-Add a new package and enter: `https://github.com/livekit/client-sdk-swift`
-
-## iOS Usage
-
-LiveKit provides an UIKit based `VideoView` class that renders video tracks. Subscribed audio tracks are automatically played.
+### Something Simple with SwiftUI
```
davidzhao, line 41: is this a placeholder title?

davidzhao (commented):
Great start!

README.md
```diff
@@ -1,22 +1,24 @@
...
 
-Official Client SDK for [LiveKit](https://github.com/livekit/livekit-server).
-Easily add video & audio capabilities to your iOS/macOS apps.
+This is the official LiveKit Swift client SDK. 
 
-## Docs & Example app
+Together with LiveKit [server](https://github.com/livekit/livekit-server), you can easily add real-time video/audio/data capabilities to your iOS or macOS apps.
 
-Docs and guides are at [https://docs.livekit.io](https://docs.livekit.io).
+<insert gif/image here />
```
hiroshihorie, line 7: Placeholder for now ?

README.md
dsa: It can be placeholder, or it can be the actual title.

README.md
dsa: Yeah, they're useful, I was thinking they could be on the wiki for the repo and perhaps linked to from the main README

README.md
dsa: Oh, nice, yeah, let's do that.

README.md
dsa: Yeah, happy to use a video that you record =)

README.md
davidzhao: sure, that works. Just don't want to lose this. Memory management can be tricky for folks that are building real applications. If not careful, would lead to crashes/etc.

PR #137, opened 2022-09-13
Ping/Pong heartbeat logic

hiroshihorie:
https://github.com/livekit/client-sdk-js/pull/377

### DIFF
--- a/Sources/LiveKit/Core/SignalClient.swift
+++ b/Sources/LiveKit/Core/SignalClient.swift
@@ -56,7 +56,8 @@ internal class SignalClient: MulticastDelegate<SignalClientDelegate> {
     private var webSocket: WebSocket?
     private var latestJoinResponse: Livekit_JoinResponse?
 
-    private lazy var pingTimer = DispatchQueueTimer(timeInterval: 1, queue: queue)
+    private var pingIntervalTimer: DispatchQueueTimer?
+    private var pingTimeoutTimer: DispatchQueueTimer?
 
     init() {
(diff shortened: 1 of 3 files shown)


Sources/LiveKit/Core/SignalClient.swift
```diff
@@ -608,6 +617,70 @@ internal extension SignalClient {
...
+        pingTimeoutTimer = nil
+        // check received joinResponse already
+        guard let jr = latestJoinResponse,
+              // check server supports ping/pong
+              jr.pingTimeout > 0,
+              jr.pingInterval > 0 else { return }
+
+        log("ping/pong starting with interval: \(jr.pingInterval)")
+
+        pingIntervalTimer = {
```
lukasIO, line 677: I don't think it would be a big problem, but just to make sure I understand correctly, the flow with this implementation is. 
1. Join
2. restartPingTimer 
3. send ping for the first time after pingInterval time elapsed
4. only start pingTimeout if a pong has been received

so there's no actual interval for pings that keeps running, but rather every single ping message relies on a pong message to be received, right?

hiroshihorie: Yes the ping will not be sent out until pong arrives. `pingTimeoutTimer` will close the connection if pong doesn't arrive.

I could change the logic to continuously send out pings.

hiroshihorie:
Logic has been modified to continuously send out pings instead of waiting until pong received.

PR #163, opened 2022-11-23
Set sender maxBitrateBps when publishing audio tracks

hiroshihorie:
From the JS SDK:

- Set sender encoding (`maxBitrateBps`) for audio tracks.
- Include missing audio presets.

### DIFF
--- a/Sources/LiveKit/Types/AudioEncoding.swift
+++ b/Sources/LiveKit/Types/AudioEncoding.swift
@@ -18,44 +18,55 @@ import Foundation
 import WebRTC
 
 @objc
-public class VideoEncoding: NSObject, MediaEncoding {
+public class AudioEncoding: NSObject, MediaEncoding {
 
     @objc
     public var maxBitrate: Int
 
     @objc
-    public var maxFps: Int
-
-    @objc
-    public init(maxBitrate: Int, maxFps: Int) {
+    public init(maxBitrate: Int) {
         self.maxBitrate = maxBitrate
-        self.maxFps = maxFps
     }
 
     // MARK: - Equal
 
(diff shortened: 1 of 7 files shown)


Sources/LiveKit/Types/AudioEncoding.swift
```diff
@@ -0,0 +1,72 @@
...
+        presetMusicHighQuality,
+        presetMusicHighQualityStereo
+    ]
+
+    public static let presetTelephone = AudioEncoding(maxBitrate: 12_000)
+    public static let presetSpeech = AudioEncoding(maxBitrate: 20_000)
+    public static let presetMusic = AudioEncoding(maxBitrate: 32_000)
+    public static let presetMusicStereo = AudioEncoding(maxBitrate: 48_000)
+    public static let presetMusicHighQuality = AudioEncoding(maxBitrate: 64_000)
+    public static let presetMusicHighQualityStereo = AudioEncoding(maxBitrate: 96_000)
```
jbrough, line 71: I'm not sure these presets have any benefit. The WebRTC default is around 32_000 and ok for most things. If high quality is needed then may as well set 510_000 - 96_000 is still only 48kbps per channel which is relatively low quality even with Opus, as maxaveragebitrate applies to the track as a whole and the L/R channels are decoded on the client.

jbrough (approved):
LGTM. I took a look as its similar to my PR previous to this.

I think this still needs server-side changes to work with Opus audio.

Sources/LiveKit/Types/AudioEncoding.swift
hiroshihorie: Thanks for the info I will let my team know.
CC: @davidzhao

PR #166, opened 2022-11-24
Fixed Xcode 13 compilation by helping the compiler

amseddi:
## Issue

Compiling with Xcode 13 fails with the following errors:
```
./client-sdk-swift/Sources/LiveKit/Core/DataChannelPair.swift:111:22: error: cannot infer return type for closure with multiple statements; add explicit type to disambiguate
        let result = { switch reliability {
                     ^
                       () -> <#Result#> in 
```
```
client-sdk-swift/Sources/LiveKit/Core/Engine.swift:260:41: error: unexpected non-void return value in void function
                return self.publisherDC.openCompleter
                                        ^
```

## Solution
Help the compiler by adding explicit closure types.


hiroshihorie (approved):
Nice fix !
I will add Xcode compatibility tests also !

PR #168, opened 2022-12-13
Async/await syntax support

hiroshihorie:
Apple's async / await syntax instead of Promises.


davidzhao:
Any reason these are defined as extensions vs function definitions in the same file? I remember trying to look for some functions earlier.. and it was not intuitive that they were defined in extensions.

hiroshihorie:
Just like the `ObjCSupport.swift` I'm splitting the code into multiple files not to overcrowd the main files.
But I can adjust to your preference , which do you prefer ?

davidzhao (approved):

PR #200, opened 2023-04-11
Allowing subscription settings to update before fully subscribed

hiroshihorie:
https://github.com/livekit/client-sdk-js/pull/471

### DIFF
--- a/Sources/LiveKit/Publications/RemoteTrackPublication.swift
+++ b/Sources/LiveKit/Publications/RemoteTrackPublication.swift
@@ -36,7 +36,7 @@ public class RemoteTrackPublication: TrackPublication {
     // MARK: - Private
 
     // user's preference to subscribe or not
-    private var preferSubscribed: Bool?
+    private var preferSubscribed: Bool
     private var metadataMuted: Bool = false
 
     // adaptiveStream
@@ -46,6 +46,8 @@ public class RemoteTrackPublication: TrackPublication {
     internal override init(info: Livekit_TrackInfo,
                            track: Track? = nil,
                            participant: Participant) {
+
+        preferSubscribed = participant.room.engine._state.connectOptions.autoSubscribe
 
         super.init(info: info,
                    track: track,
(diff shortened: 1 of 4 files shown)


Sources/LiveKit/Publications/RemoteTrackPublication.swift
```diff
@@ -184,7 +204,15 @@ private extension RemoteTrackPublication {
 
     var userCanModifyTrackSettings: Bool {
         // adaptiveStream must be disabled and must be subscribed
-        !isAdaptiveStreamEnabled && subscribed
+        if kind == .video && isAdaptiveStreamEnabled {
+            log("Adaptive stream is enabled, cannot change video track settings", .warning)
+            return false
+        }
+        if !(preferSubscribed || subscribed) {
```
davidzhao, line 211: I believe this is no longer true. they should be able to change settings anytime

Sources/LiveKit/Publications/RemoteTrackPublication.swift
```diff
@@ -115,11 +117,29 @@ public class RemoteTrackPublication: TrackPublication {
...
+        return send(trackSettings: settings)
+    }
+
+    /// For tracks that support simulcasting, adjust subscribed quality.
+    ///
+    /// This indicates the highest quality the client can accept. if network
+    /// bandwidth does not allow, server will automatically reduce quality to
+    /// optimize for uninterrupted video.
+    @discardableResult
+    public func set(preferredVideoQuality newValue: VideoQuality) -> Promise<Void> {
```
davidzhao, line 135: it'd be better to support the dimensions API instead of video quality. we would like to deprecate video quality settings because it's not precise (i.e. if the publisher only publishes two layers, they'd be mapped to LOW/MED). it would not make sense for subscriber to request HIGH

Sources/LiveKit/Publications/RemoteTrackPublication.swift
```diff
@@ -184,7 +199,15 @@ private extension RemoteTrackPublication {
...
+        if kind == .video && isAdaptiveStreamEnabled {
+            log("Adaptive stream is enabled, cannot change video track settings", .warning)
+            return false
+        }
+        if !(preferSubscribed || subscribed) {
+            log("Cannot update track settings when not subscribed", .warning)
+            return false
+        }
+        return true
     }
```
hiroshihorie, line 211: @davidzhao Do you mean this whole check is not required anymore ?

davidzhao: I just mean the `subscribed` check. `isAdaptiveStreamEnabled` is still needed. though re-reading the code, you now have `preferSubscribed` which actually covers it. sorry about the false alarm.

pblazej:
@hiroshihorie is this still relevant?

PR #201, opened 2023-04-11
Support for updating name and metadata of LocalParticipant

hiroshihorie:
https://github.com/livekit/client-sdk-js/pull/599

Should we have a way to fail if `CanUpdateOwnMetadata` permission doesn't exist ?

### DIFF
--- a/Sources/LiveKit/Core/SignalClient.swift
+++ b/Sources/LiveKit/Core/SignalClient.swift
@@ -355,6 +355,10 @@ private extension SignalClient {
             notify { $0.signalClient(self, didUpdate: token) }
         case .pong(let r):
             onReceivedPong(r)
+        case .reconnect:
+            log("received reconnect message")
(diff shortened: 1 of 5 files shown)


lukasIO:
I think that makes sense, @davidliu proposed to wait for a local participant metadata update event to come in before returning. 
using the permissions locally has the disadvantage that the permissions might have just been updated and only the server is aware of them right now. 
using the metadata update as an indicator that it resolved properly has the downside that this update might also be coming from a server sdk and might be unrelated (we could get around that by checking the contents of the update)

hiroshihorie:
That sounds good 👍

hiroshihorie:
@lukasIO Can I merge this ?

lukasIO:
yeah, let's iterate on how to acknowledge it later on. As long as the function is async we can add additional logic for that later!

Issue #219, opened 2023-06-21
use Broadcast Upload Extension, bug screen sharing doesn't work in the background


wzJun1:
I configured screen sharing according to this wiki: https://github.com/livekit/client-sdk-swift/wiki/iOS-Screen-Sharing

But there is a problem, after entering the background, the screen sharing stops.

`background mode` has been set. `Broadcast Upload Extension` has been imported.

but when the App enters the background, it will stop screen sharing, and the system prompts: Your App has stopped screen sharing for some reason. As shown in the picture:

![](https://s2.loli.net/2023/06/21/zIne1X4NBlk69vF.jpg)

I checked the memory usage of this Extension, it is not very high, generally between 8-20M.

By running `Broadcast Upload Extension`, some logs are recorded when the extension stops:

2023-06-21 22:06:22.890540+0800 BroadcastExtension[6966:290838] [Broadcast] server socket closed
2023-06-21 22:06:22.890778+0800 BroadcastExtension[6966:290838] [Broadcast] client connection did close nil
Program ended with exit code: 9

wzJun1:
Screen sharing in my App is normal, but after entering the background for about 1-2 seconds, it will be disconnected and the above pop-up window will be prompted.

Before entering the background, the audio streaming has been going on, and the call is normal.

wzJun1:
I print error log in `setupConnection` method in `LKSampleHandler.swift` file, `print(error)`, it is nil

I don't know what caused it, guess it's like the `Extension` process was killed

wzJun1:
I found a sample App with the same problem:

https://github.com/livekit-examples/swift-example/

elesahich:
👀

wzJun1:
> What version are you using? Same here with 1.0.13. 👀

It should be the latest version, because I am directly clone git

wzJun1:
The problem is that I used the case and the same problem occurred. Case URL:

https://github.com/livekit-examples/swift-example/

elesahich:
In my case, I think this is an iOS specific issue, as Android and Web SDK work fine even if they go into the background or something.

wzJun1:
> In my case, I think this is an iOS specific issue, as Android and Web SDK work fine even if they go into the background or something.

So is this problem insoluble? I want to seek a solution. Because I noticed that other apps do not have this issue, their screen sharing is normal, such as Zoom

elesahich:
[...]

Issue #182, opened 2023-02-24
Screen sharing stop when app in background


vuongxuantuyen:
I take screen sharing with this instruction: https://github.com/livekit/client-sdk-swift/wiki/iOS-Screen-Sharing. Every thing work fine except the screen sharing will stop when enter background. The same issue on the example application too. Do you have any idea?

hiroshihorie:
Can you try publishing an audio track before going into background ?

std-s:
@vuongxuantuyen  hi，any solution？

elesahich:
I've found a few other apps that I can reference, but implementing PiP seems to be the only way to go. Are there any solutions?

elesahich:
For me a day ago : PiP was not an answer :)......

hiroshihorie:
Hello can you try the solution here :
https://github.com/livekit/client-sdk-swift/issues/219
Set RoomOptions.suspendLocalVideoTracksInBackground to false.

vuongxuantuyen:
@hiroshihorie It work, thank you.

PR #221, opened 2023-06-23
Only suspend Camera tracks in background

hiroshihorie:
Related:
https://github.com/livekit/client-sdk-swift/issues/219
https://github.com/livekit/client-sdk-swift/issues/182


hiroshihorie:
I think this is safe to merge I will go ahead and merge this one.

PR #222, opened 2023-06-24
Add `AVSampleBufferDisplayLayer` rendering mode to `VideoView`

hiroshihorie:
This is needed for PiP.


Sources/LiveKit/Views/InternalSampleBufferVideoRenderer.swift
```diff
@@ -0,0 +1,73 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import Foundation
+import WebRTC
+
+internal class InternalSampleBufferVideoRenderer: NativeView {
```
davidzhao, line 20: What does `Internal` reference? or should we call it `AVSampleBufferDisplayLayerRenderer` ?

davidzhao (approved):
lg. just one question re: classname

Sources/LiveKit/Views/InternalSampleBufferVideoRenderer.swift
hiroshihorie: This is not exposed, the `VideoView` uses it internally as a subview.

davidzhao: got it, sounds good!

PR #226, opened 2023-07-05
e2ee

cloudwebrtc:
TODO:
- [x]  E2EE implementation
- [x]  Add to the example app

example app: https://github.com/livekit-examples/swift-example/pull/18

image

image

### DIFF
--- a/Sources/LiveKit/E2EE/E2EEManager.swift
+++ b/Sources/LiveKit/E2EE/E2EEManager.swift
@@ -20,9 +20,8 @@ import WebRTC
 @objc
 public class E2EEManager: NSObject, ObservableObject, Loggable {
     internal var room: Room?
-    internal var enabled: Bool = false
+    internal var enabled: Bool = true
     private let e2eeOptions: E2EEOptions
-    private var keyProvier: BaseKeyProvider?
     var frameCryptors = [String: RTCFrameCryptor]()
     
     public init(e2eeOptions: E2EEOptions) {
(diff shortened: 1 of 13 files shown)


Sources/LiveKit/Protocols/RoomDelegate.swift
```diff
@@ -139,6 +139,10 @@ public protocol RoomDelegateObjC: AnyObject {
     /// Same with ``ParticipantDelegate/participant(_:didUpdate:permission:)``.
     @objc optional
     func room(_ room: Room, participant: RemoteParticipant, didUpdate publication: RemoteTrackPublication, permission allowed: Bool)
+
+    /// ``Room``'e2ee state has been updated.
+    @objc(room:publication:didUpdateE2EEState:) optional
+    func room(_ room: Room, publication: TrackPublication, didUpdate e2eeState: E2EEState)
```
hiroshihorie, line 145: I was thinking of slowly migrating to a more explicit delegate signature,
for example didUpdateXXXXX instead of so many `didUpdate` delegates now.

Sources/LiveKit/E2EE/E2EEManager.swift
```diff
@@ -0,0 +1,139 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import Foundation
+import WebRTC
+
+@objc
+public class E2EEManager: NSObject, ObservableObject, Loggable {
+    internal var room: Room?
```
hiroshihorie, line 22: This might be better a `weak` var.

hiroshihorie (approved):
Will push minor changes, can you run example if everything works as expected ?

Sources/LiveKit/Protocols/RoomDelegate.swift
hiroshihorie: fixed 👇

Sources/LiveKit/E2EE/E2EEManager.swift
hiroshihorie: fixed 👇

cloudwebrtc:
okay, I'll run the example again later to make sure it works correctly

hiroshihorie:
@cloudwebrtc 
I tried the example app looks great !
What do you think about merging this one ?

cloudwebrtc:
yeah, I think the PR can be merged

PR #231, opened 2023-07-19
Convenience method to check cam / mic permissions

hiroshihorie:


Sources/LiveKit/LiveKit+DeviceHelpers.swift
```diff
@@ -0,0 +1,44 @@
...
+
+            let status = AVCaptureDevice.authorizationStatus(for: type)
+            switch status {
+            case .notDetermined:
+                if !(await AVCaptureDevice.requestAccess(for: type)) {
+                    return false
+                }
+            case .restricted, .denied: return false
+            case .authorized: continue // No action needed for authorized status.
+            @unknown default: fatalError("Unknown AVAuthorizationStatus")
```
davidzhao, line 38: would this crash the app? it seems better to just return false and log an error?

davidzhao (approved):

PR #233, opened 2023-07-26
Optimize BufferCapturer for non-screen share use case

hiroshihorie:
H264 encoder uses different optimizations for screen share / non-screen share.


davidzhao (approved):
lg

PR #237, opened 2023-08-14
Expose `RTCAudioProcessingModule` and `AudioTrack.add(audioRenderer:)`

hiroshihorie:


Sources/LiveKit/Track/Local/LocalAudioTrack.swift
```diff
@@ -77,6 +77,16 @@ public class LocalAudioTrack: Track, LocalTrack, AudioTrack {
             return didUnpublish
         }
     }
+
+    public func add(audioRenderer: RTCAudioRenderer) {
```
davidzhao, line 81: I didn't realize this works for local tracks too! that's great

davidzhao (approved):

PR #242, opened 2023-08-30
Maintain specified resolution for simulcast

hiroshihorie:
When simulcast is enabled (or if simulcastLayers is explicitly specified) I think the developer doesn't expect the resolution to be anything other than specified.

If simulcast is not enabled (single layer), I think it's ok to drop resolution when network isn't great.

### DIFF
--- a/Sources/LiveKit/Participant/LocalParticipant.swift
+++ b/Sources/LiveKit/Participant/LocalParticipant.swift
@@ -158,14 +158,17 @@ public class LocalParticipant: Participant {
             track.set(transport: publisher,
                       rtpSender: transceiver.sender)
 
-            // prefer to maintainResolution for screen share
-            if case .screenShareVideo = track.source {
-                self.log("[publish] set degradationPreference to .maintainResolution")
-                let params = transceiver.sender.parameters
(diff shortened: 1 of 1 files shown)


Sources/LiveKit/Participant/LocalParticipant.swift
```diff
@@ -158,14 +158,17 @@ public class LocalParticipant: Participant {
...
-                self.log("[publish] set degradationPreference to .maintainResolution")
-                let params = transceiver.sender.parameters
-                params.degradationPreference = NSNumber(value: RTCDegradationPreference.maintainResolution.rawValue)
-                // changing params directly doesn't work so we need to update params
-                // and set it back to sender.parameters
-                transceiver.sender.parameters = params
+            if track is LocalVideoTrack {
+                let publishOptions = (publishOptions as? VideoPublishOptions) ?? self.room._state.options.defaultVideoPublishOptions
+                // if simulcast is enabled, degrade resolution by using server's layer switching logic instead of WebRTC's logic
+                if publishOptions.simulcast {
```
davidzhao, line 164: should it also set to maintain resolution when it's screenshare?

davidzhao (approved):

PR #245, opened 2023-09-14
WebSocket: Migrate to async/await from Promises lib

hiroshihorie:
First step of migrating internals to use async/await instead of Promises lib.


davidzhao (approved):
lgtm

PR #248, opened 2023-09-19
fix bug for e2ee.

cloudwebrtc:
Remove the storage of rtpSender/rtpReceiver from the participant, which will cause the rtpSender/rtpReceiver of audio and video to be overwritten.


hiroshihorie (approved):
Over use of forced unwrapping gives me anxiety 😅
But other than that it looks good.

cloudwebrtc:
@hiroshihorie oka， added a nil pointer check
https://github.com/livekit/client-sdk-swift/pull/248/commits/691b28df55107163b8e7ecd9774b7f550eebf0dc

hiroshihorie:
lg!

PR #263, opened 2023-10-21
Fix: `CameraCapturer` report actual frame dimensions

hiroshihorie:
Fix: `CameraCapturer` report actual frame dimensions, instead of reporting dimensions in options as initial dimensions.


davidzhao:
nice fix!

PR #272, opened 2023-11-06
Make `LocalParticipant` non-Optional

hiroshihorie:
I found `Room.localParticipant` being optional inconvenient so this PR makes it always available. (for v2)

`sid`, `identity`, `name` will all be an empty string for LocalParticipant until it connects.
I'm still thinking if this should be a `String?` instead.


davidzhao (commented):
nice!

PR #280, opened 2023-12-05
v2 API changes

hiroshihorie:


Sources/LiveKit/Participant/Participant.swift
```diff
@@ -53,7 +53,7 @@ public class Participant: NSObject, ObservableObject, Loggable {
     public var joinedAt: Date? { _state.joinedAt }
 
     @objc
-    public var tracks: [String: TrackPublication] { _state.tracks }
+    public var tracksPublications: [Sid: TrackPublication] { _state.tracks }
```
lukasIO, line 56: nit: in the JS version this is called `trackPublications` with the plural `s` only on publications. I think it would be good to align this across platforms

Sources/LiveKit/Participant/LocalParticipant.swift
```diff
@@ -307,7 +307,7 @@ public class LocalParticipant: Participant {
     @objc
     public func publish(data: Data,
```
lukasIO, line 308: can we rename this to `publishData`, too, so that it's easier to refer to the same API in the docs?

Sources/LiveKit/Participant/Participant.swift
hiroshihorie: Oops, this is a refactoring typo

PR #283, opened 2023-12-18
Explicit delegate signatures

hiroshihorie:
A few delegates had the same signatures (by using overloading) such as:

`room(_ room: Room, didUpdate speakers: [Participant])`
`room(_ room: Room, didUpdate metadata: String?)`
`room(_ room: Room, didUpdate isRecording: Bool)`

This pattern will break if add another method of the same type will be added in the future:
`room(_ room: Room, didUpdate isSomeExampleFlag: Bool)`

Additionally, since labels can be omitted in Swift, the implementation side can be ambiguous:
```Swift
func room(_: Room, didUpdate _: String?) {
  // ...
}

func room(_: Room, didUpdate _: Bool) {
  // ...
}
```

This PR makes delegate signatures explicit and future-proof by including the label in the signature:

`room(_ room: Room, didUpdateSpeakingParticipants participants: [Participant])`
`room(_ room: Room, didUpdateMetadata metadata: String?)`
`room(_ room: Room, didUpdateIsRecording isRecording: Bool)`

I can deprecate the old methods, but since this will be part of v2 release maybe it's not required to deprecate.

### DIFF
--- a/Sources/LiveKit/Core/Room+EngineDelegate.swift
+++ b/Sources/LiveKit/Core/Room+EngineDelegate.swift
@@ -40,7 +40,7 @@ extension Room: EngineDelegate {
             }
 
             delegates.notify(label: { "room.didUpdate connectionState: \(state.connectionState) oldValue: \(oldState.connectionState)" }) {
-                $0.room?(self, didUpdate: state.connectionState, oldValue: oldState.connectionState)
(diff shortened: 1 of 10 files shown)


cloudwebrtc (approved):
lgtm

PR #294, opened 2024-01-08
Organize delegates

hiroshihorie:


Sources/LiveKit/TrackPublications/TrackPublication.swift
```diff
@@ -216,10 +216,10 @@ extension TrackPublication: TrackDelegateInternal {
             }
 
             participant.delegates.notify {
-                $0.participant?(participant, didUpdatePublication: self, isMuted: isMuted)
+                $0.participant?(participant, track: self, didUpdateIsMuted: isMuted)
             }
             participant.room.delegates.notify {
-                $0.room?(participant.room, participant: participant, didUpdatePublication: self, isMuted: self.isMuted)
+                $0.room?(participant.room, participant: participant, track: self, didUpdateIsMuted: self.isMuted)
```
davidzhao, line 222: should we name these `publication` instead of `track`?

PR #296, opened 2024-01-09
Async room id

hiroshihorie:


Sources/LiveKit/Core/Room.swift
```diff
@@ -28,11 +28,19 @@ public class Room: NSObject, ObservableObject, Loggable {
 
     // MARK: - Public
 
-    @objc
+    @objc(sid)
     /// Server assigned id of the Room.
     /// This will be empty until ``RoomDelegate/room(_:didUpdateRoomId:)`` is called.
```
davidzhao, line 33: this comment isn't needed anymore?

Sources/LiveKit/Core/Room.swift
```diff
@@ -28,11 +28,19 @@ public class Room: NSObject, ObservableObject, Loggable {
...
     public var sid: Sid? { _state.sid }
 
+    // Work-around error: Property with 'throws' or 'async' is not representable in Objective-C
+    /// Server assigned id of the Room. async version of ``Room/sidValue``.
+    public var asyncSid: Sid {
```
davidzhao, line 38: how would one use this value? `await room.asyncSid()` ? should we state usage in the comment?

hiroshihorie: Xcode will show it's an await-able property.

hiroshihorie:
After merging, I realized property and method can have same name so I've made it to a method (additionally adds ObjC support) `await room.sid()` 
https://github.com/livekit/client-sdk-swift/commit/7ae1aa5517f269e6743713816cf7d52f902ca041

PR #297, opened 2024-01-09
Simplify `publish(data:options:)` method

hiroshihorie:


Sources/LiveKit/Participant/LocalParticipant.swift
```diff
@@ -312,21 +312,19 @@ public class LocalParticipant: Participant {
     ///   - reliability: Toggle between sending relialble vs lossy delivery.
     ///     For data that you need delivery guarantee (such as chat messages), use Reliable.
     ///     For data that should arrive as quickly as possible, but you are ok with dropped packets, use Lossy.
-    ///   - destinations: SIDs of the participants who will receive the message. If empty, deliver to everyone
+    ///   - options: Provide options with a ``DataPublishOptions`` class.
     @objc
     public func publish(data: Data,
                         reliability: Reliability = .reliable,
```
lukasIO, line 318: I think we wanted to move `reliability` also inside the options object and default to `unreliable` IIRC

lukasIO: yup. see [here](https://github.com/livekit/client-sdk-js/pull/946/files). It should just be a boolean that defaults to false. so no extra type needed. that was an effort to make it simpler

hiroshihorie: Oh, ok we are changing default to lossy ?

lukasIO: yeah, that's what we intend to do!

davidzhao: do we need to change the callback as well in swift?

hiroshihorie: @lukasIO `Reliability` type has been removed and simplified.
@davidzhao What do you mean by callback ?

davidzhao: I meant the delegate:
```
func room(_ room: Room, participant: RemoteParticipant?, didReceiveData data: Data, forTopic topic: String)
```
So we'll note the topic, but it won't indicate if the packet type was reliable or unreliable.. I guess that's fine.

hiroshihorie: Oh ok maybe i should include this in delegate

davidzhao: I think it's ok without actually.

hiroshihorie: if this lg I'll go ahead and merge it for now.

davidzhao (approved):
lgtm

PR #321, opened 2024-01-24
Revert async multicast

hiroshihorie:


hiroshihorie:
This was not the cause of memory leak I was tracking.

PR #335, opened 2024-02-07
Improve reconnect logic

hiroshihorie:


Sources/LiveKit/Core/SignalClient.swift
```diff
@@ -217,6 +210,9 @@ actor SignalClient: Loggable {
         await _addTrackCompleters.reset()
         await _requestQueue.clear()
         await _responseQueue.clear()
+
+        self.disconnectError = LiveKitError.from(error: disconnectError)
+        connectionState = .disconnected
```
davidzhao, line 215: why move this? does it change anything?

Sources/LiveKit/Participant/LocalParticipant.swift
```diff
@@ -421,7 +421,7 @@ extension LocalParticipant {
             }
     }
 
-    func republishTracks() async throws {
+    func rePublishAllTracks() async throws {
```
davidzhao, line 424: nit: republish does not need to be capitalized like rePublish
```suggestion
    func republishAllTracks() async throws {
```

davidzhao (commented):
a few nits.. I'm not 100% about the queuing behavior.. it seems that we may potentially be queuing things when we shouldn't.

with full reconnect cases, nothing should be queued, since it's basically a new connection

Sources/LiveKit/Core/SignalClient.swift
hiroshihorie: I think it changes the timing of `didUpdateConnectionState:` delegate firing.

davidzhao (approved):

PR #338, opened 2024-02-12
Integrate E2EE Manager

hiroshihorie:
* Current E2EE manager holds strong reference to `RemoteParticipant`s, try to solve this since it causes issues.
* Add frame cryptors to simulcast senders also.

New design:
* Internal `E2EEManager` class is removed and `Track` holds `FrameCryptor`s instead.
* RtpSender/Receivers are in pair with `FrameCryptor`s.
* Doesn't break current release.

- [x] isEnabled property
- [x] Delegate


lukasIO:
I like the idea behind this!
How would the new design allow for encryption to be enabled/disabled ?

hiroshihorie:
I haven't ported that from E2EE manager yet, also the delegate.

hiroshihorie:
Build with this changes have been uploaded to tesflight.

lukasIO:
more of a general question, would it make sense to un-set the cryptors also when tracks have ended ? 
Basically just thinking about how this would get cleaned up properly

hiroshihorie:
A good point regarding cleaning up the cryptors and other properties of the track.
Currently sender/receivers and cryptors will be reset if track is re-published.

I'll check how FrameCryptor retains reference to keyprovider etc.
But looks like there are no memory leaks when the Room is closed.

While publishing video / mic:
Screenshot 2024-02-15 17 48 00

After room is closed:
Screenshot 2024-02-15 17 50 23

lukasIO:
Nice, thanks for checking! Is the same true for unpublishing a single track instead of disconnecting the room? Thinking about cases where users publish/unpublish tracks often.

hiroshihorie:
Might need to re-structure/design this.

PR #348, opened 2024-02-26
Synchronous VideoView internal state

hiroshihorie:
Update VideoView's internal state synchronously so properties such as avSampleBufferDisplayLayer can be accessed immediately.


hiroshihorie:
If I enforce all properties of VideoView to be accessed from main thread, I can simplify a lot of internal code.
UI components, inheriting from UIView should always be accessed from main thread, so the change should be ok but I wonder if it affects any current users.

PR #351, opened 2024-03-05
Guarantee serial execution of simplified publish APIs

hiroshihorie:
Prevents races that can occur between internally called `getTrackPublication` and `publish`.
Perhaps this can be improved to allow concurrent execution if `source` types are different.


hiroshihorie:
Using `AsyncChannel` might be more solid:
https://github.com/apple/swift-async-algorithms/blob/main/Sources/AsyncAlgorithms/AsyncAlgorithms.docc/Guides/Channel.md

PR #356, opened 2024-03-11
Remove CHeaders

hiroshihorie:


hiroshihorie:
@davidliu I think this was added with your ios screen share implementation, is it ok to remove ?
Looks like it still works.

davidliu:
Honestly been too long ago I don't remember what these cheaders were for.

PR #363, opened 2024-04-11
Add Privacy manifest file

hiroshihorie:
Apple docs:
[Privacy manifest files](https://developer.apple.com/documentation/bundleresources/privacy_manifest_files)
[Describing use of required reason API](https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_use_of_required_reason_api)


davidliu (commented):
I think cocoapods might need handling to put it in the resource bundle:

https://forums.developer.apple.com/forums/thread/733537

davidzhao (approved):
lg!

PR #373, opened 2024-05-07
Variable sampleRate for `toAVAudioPCMBuffer()`

hiroshihorie:
Dynamically compute sampleRate instead of hardcoding.


hiroshihorie:
A bit worried if it's correct for more than 1 ch (stereo etc.) But once stereo is implemented the whole SDK will need to be throughly checked.

PR #377, opened 2024-05-20
Speed up camera publishing

hiroshihorie:
Maintain a internal list of AVCaptureDevices to speed up publishing.

### DIFF
--- a/Sources/LiveKit/Support/DeviceManager.swift
+++ b/Sources/LiveKit/Support/DeviceManager.swift
@@ -31,6 +31,7 @@ class DeviceManager: Loggable {
 
     struct State {
         var devices: [AVCaptureDevice] = []
+        var didEnumerateDevices = false
     }
 
     private var _state = StateSync(State())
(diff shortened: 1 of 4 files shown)


Sources/LiveKit/Support/DeviceManager.swift
```diff
@@ -0,0 +1,69 @@
...
+    }()
+
+    var _observation: NSKeyValueObservation?
+
+    init() {
+        log()
+
+        DispatchQueue.global(qos: .background).async { [weak self] in
+            guard let self else { return }
+            _observation = session.observe(\.devices, options: [.initial, .new]) { [weak self] _, value in
```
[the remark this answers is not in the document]
hiroshihorie: This is observing the `devices` property so it should be ok. But continuity devices are not included now, maybe I should include it.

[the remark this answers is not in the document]
davidzhao: oh awesome, I missed that. lgtm!

davidzhao (approved):

Sources/LiveKit/Support/DeviceManager.swift
```diff
@@ -0,0 +1,77 @@
...
+            .builtInDualCamera,
+            .builtInDualWideCamera,
+            .builtInTripleCamera,
+            .builtInWideAngleCamera,
+            .builtInTelephotoCamera,
+            .builtInUltraWideCamera,
+        ]
+        #else
+        deviceTypes = [
+            .builtInWideAngleCamera,
```
davidzhao, line 46: what about .external and .continuityCamera?

Sources/LiveKit/Support/DeviceManager.swift
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    public var devices: [AVCaptureDevice] { _state.devices }
+    public let devicesCompleter = AsyncCompleter<[AVCaptureDevice]>(label: "devices", defaultTimeout: 10)
+
+    private lazy var discoverySession: AVCaptureDevice.DiscoverySession = {
+        let deviceTypes: [AVCaptureDevice.DeviceType]
+        #if os(iOS)
+        deviceTypes = [
+            .builtInDualCamera,
```
davidzhao, line 37: would these show up as diff devices? or the same? if they are all different, it would be confusing right?

hiroshihorie: I think the device manager should bring aware of all devices but for the user facing device enumeration API, i can add a filter option (default on)

PR #389, opened 2024-05-31
Option to limit zoom out on VideoView

hiroshihorie:


Sources/LiveKit/Views/VideoView.swift
```diff
@@ -152,6 +152,12 @@ public class VideoView: NativeView, Loggable {
         set { _state.mutate { $0.isAutoZoomResetEnabled = newValue } }
     }
 
+    @objc
+    public var isZoomOutEnabled: Bool {
```
davidzhao, line 156: I think this is not very clear.. should it be `minZoomLevel` instead? i.e. `minZoomLevel = 1` by default.

hiroshihorie: I agree it's confusing, I'll think of another name...

PR #406, opened 2024-06-18
Transcription events CLT-343

hiroshihorie:
- [x] Events for RoomDelegate
- [x] Events for ParticipantDelegate
- [ ] State management


Sources/LiveKit/Core/Room+EngineDelegate.swift
```diff
@@ -194,4 +194,35 @@ extension Room {
             }
         }
     }
+
+    func room(didReceiveTranscriptionPacket packet: Livekit_Transcription) {
+        guard let participant = _state.read({
+            $0.remoteParticipants.values.first { $0.identity == Participant.Identity(from: packet.transcribedParticipantIdentity) }
+        }) else {
+            log("[Transcription] Could not find participant: \(packet.transcribedParticipantIdentity)", .warning)
```
bcherry, line 202: @hiroshihorie you'll need to include support for transcription events coming down on the LocalParticipant as well, to record the results of STT in the agent

bcherry: I did a quick spike and it works to just change the delegate methods to receive generic Participant and TrackPublication instead of the Remote variants, change these lines up here to the following

```
        let participant: Participant?
        
        let identity = Participant.Identity(from: packet.transcribedParticipantIdentity)
        if localParticipant.identity == identity {
            participant = localParticipant
        } else {
            participant = _state.read({ $0.remoteParticipants.values.first { $0.identity == identity } })
        }
            
        guard let participant else {
            log("[Transcription] Could not find participant: \(packet.transcribedParticipantIdentity)", .warning)
            return
        }
 ```

Comment (hiroshihorie) [reply Sources/LiveKit/Core/Room+EngineDelegate.swift:202]: Thank you ! will update.

**Comment by hiroshihorie** [comment]
If looks good will merge first, then think about if any state helpers are necessary.

**Review by bcherry — APPROVED** [review]
looks good - go ahead and file a new issue for any state management follow-ons

File: Sources/LiveKit/Types/TranscriptionSegment.swift [hunk 2]
```diff
@@ -0,0 +1,68 @@
...
+
+    override public func isEqual(_ object: Any?) -> Bool {
+        guard let other = object as? Self else { return false }
+        return id == other.id
+    }
+
+    override public var hash: Int {
+        var hasher = Hasher()
+        hasher.combine(id)
+        return hasher.finalize()
```
bcherry, line 53: @hiroshihorie @lukasIO do you two think it's right to identify and hash these based on ID alone? It doesn't guarantee true equality, because a newer version could be received later that should be used instead.

lukasIO: what's the intended purpose of the hash?

PR #408, opened 2024-06-20
Fix `participant.identity` is nil in `participantDidDisconnect` delegate

hiroshihorie:


davidzhao (approved):
nice to see the tests!

Issue #412, opened 2024-06-26
About iOS Screen Share Bug


liuyang2211:
**Describe the bug**
 iOS Screen Share doesn't work very well

**SDK Version**
2.0.11

**iOS/macOS Version**
17.5.1

**Steps to Reproduce**
1、When I first called [self.room.localParticipant.setScreenShare(enabled: true)]，The screen popped up "BroadcastExtPickerView" . I received a callback -【LiveKit】:LocalParticipant【xxx】didPublishTrack【video,screenShareVideo,Optional("started")】publication:【LocalTrackPublication(sid: TR_VSc3eCye5W3oLK, kind: Kind(rawValue: 1), source: Source(rawValue: 3))】But I didn't click "Start". When I click "Start" I can see screenshare steam.
2、When I turn off screen sharing，and  called [self.room.localParticipant.setScreenShare(enabled: true)] nextTime. Absent popup "BroadcastExtPickerView"，  I received a callback -【LiveKit】: Participant【xxx】的【video,screenShareVideo】didUpdateIsMuted【Open】. But I can't see  screenshare steam.

**Expected behavior**
1.I think the "delegate of didPublishTrack" should be after clicking the “Start”
2.I think every time called [self.room.localParticipant.setScreenShare(enabled: true)]  Should popup "BroadcastExtPickerView" 

**Screenshots**
![2711719382773_ pic](https://github.com/livekit/client-sdk-swift/assets/19897206/56bea081-e170-43f5-a61a-6fc48b4ba300)
![2721719382774_ pic](https://github.com/livekit/client-sdk-swift/assets/19897206/1bc2fc83-6b7d-4f37-b153-a86abd7f5f8b)

**Logs**
-【LiveKit】:LocalParticipant【xxx】didPublishTrack【video,screenShareVideo,Optional("started")】publication:【LocalTrackPublication(sid: TR_VSc3eCye5W3oLK, kind: Kind(rawValue: 1), source: Source(rawValue: 3))】
-【LiveKit】: Participant【xxx】的【video,screenShareVideo】didUpdateIsMuted【isMuted】
-【LiveKit】: Participant【xxx】的【video,screenShareVideo】didUpdateIsMuted【Open】

hiroshihorie:
I think https://github.com/livekit/client-sdk-swift/pull/424 fixes this.

PR #424, opened 2024-07-05
Relaunch screenshare ext. when re-enabling the track

hiroshihorie:
Unless unpublished, the screenshare ext was not re-launching when calling setScreenShare(enabled: true) in iOS causing confusion.

Related issues:
https://github.com/livekit/client-sdk-swift/issues/412
https://github.com/livekit-examples/swift-example/issues/37
https://github.com/livekit-examples/swift-example/issues/39


davidliu:
On the other platforms, the screenshare track is unpublished entirely on disable, maybe we should do that to align the behavior?

hiroshihorie:
Either way would work but I should also align MacOS screen capturing also in that case.

hiroshihorie:
I think i'll just unpublish it 👍

Issue #429, opened 2024-07-12
Unable to publish two LocalVideoTracks


darrenW04:
**Describe the bug**
When publishing two separate LocalVideoTracks into a room, only one track is shown within a room. Using Apple's AVFoundation's AVMultiCamSession, I can locally preview my front and back camera but when I publish both of them, only the most recent track is shown in the room. 
For clarification if I was confusing
- If I push the tracks sequentially, the most recent track will be the only one that isn't frozen 
- Only one track is shown even if I have multiple published
- It disables my local previews (not using live kits track preview, just a preview of the device input locally)
- I made sure my device is capable of hosting the front and back cam simultaneously via the local preview I made.

**SDK Version**
Swift SDK 2.0.11

**iOS/macOS Version**
IOS version 17.0

**Xcode Version**
swift-driver version: 1.90.11.1 Apple Swift version 5.10 (swiftlang-5.10.0.13 clang-1500.3.9.4)
Target: arm64-apple-macosx14.0

**Steps to Reproduce**
In implementing the dual camera feature for both the front and back, I used this repo as an example on how preview multiple cameras simultaneously.
[https://github.com/jwamin/Multicam/blob/master/Trois-cam/CameraView.swift](url)
After you would create a way to connect to the room and have abilities to publish certain tracks. 
Below is how I would publish my front camera track. The device I'm using is from the AVCaptureMultiCamSession as it holds an array of inputs. In my case index 0 the back cam and index 1 the front cam.
`    
func turnFrontCameraOn() {
        Task {
            do {
                let frontCam = captureSession?.inputs[1] as? AVCaptureDeviceInput
                let track = LocalVideoTrack.createCameraTrack(name: "frontCT", options: CameraCaptureOptions(
                    device: frontCam?.device,
                    dimensions: .h720_169,
                    fps: 30
                ))
                try await room?.localParticipant.publish(videoTrack: track, options: VideoPublishOptions(name: "front", encoding: VideoEncoding(maxBitrate: 128, maxFps: 60), streamName: "frontS"))
            } catch {
                print(error)
            }
        }
    }
`
From here on you can connect to a web app such as [https://meet.livekit.io/?tab=custom](url) to view the tracks from another participant.

**Expected behavior**
Wanted Result: 
View two tracks of the front and back camera
My result:
Only the most recent track has a camera shown, I will show an example in the screenshots, with the right image being the most recent **Note when publishing tracks one by one, each track works until another track is published**

**Screenshots**
[...]

PR #432, opened 2024-07-15
Use `AVCaptureMultiCamSession` for iOS

hiroshihorie:
Allows simultaneous capturing of front and back cameras.

Related issue:
https://github.com/livekit/client-sdk-swift/issues/429


cloudwebrtc (approved):
lgtm

PR #438, opened 2024-07-19
Fix rotation in SampleBuffer Renderer

### LINKED ISSUE #437: Camera rotated when using renderMode = .sampleBuffer
PR that fixes this issue #438 

**Describe the bug**
I need to setup `renderMode` on `VideoView` to `.sampleBuffer` (this only works with Camera in PIP mode)

When having `.auto` or `.metal` video is displayed properly.
Simple change to `.sampleBuffer` rorates the video by 90 degrees

**SDK Version**
2.0.12

**iOS/macOS Version**
iOS 18 Public beta

**iPhone version**
iPhone 15 Pro

**Xcode Version**
xcode 16 Beta

**Steps to Reproduce**
Run this code, and switch renderMode from .sampleBuffer to .auto

```swift
import UIKit
import LiveKit
import LiveKitWebRTC

class RoomViewController: UIViewController {
    let myVideoTrack = LocalVideoTrack.createCameraTrack()
    lazy var videoView = {
        let view = VideoView(frame: CGRect(origin: .zero, size: CGSize(width: 300, height: 600)))
        view.track = myVideoTrack
        view.renderMode = .sampleBuffer // ⚡ Changing this parameter rotates the video
        return view
    }()
    
    override func viewDidLoad() {
        view.addSubview(videoView)
        Task {
            try await myVideoTrack.start()
        }
    }
}
```

**Expected behavior**
Orientation should not be reversed

**Screenshots**
.auto:

.sampleBuffer

patryk-sredzinski:
Fixes #437

### DIFF
--- a/Sources/LiveKit/Views/SampleBufferVideoRenderer.swift
+++ b/Sources/LiveKit/Views/SampleBufferVideoRenderer.swift
@@ -22,6 +22,13 @@ internal import LiveKitWebRTC
 
 class SampleBufferVideoRenderer: NativeView, Loggable {
     public let sampleBufferDisplayLayer: AVSampleBufferDisplayLayer
+    
+    private var firstFrameReceived = false
+    private var bufferTransform = CATransform3DIdentity
+    private var mirroredTransform = CATransform3DIdentity
+    private var displayLayerTransform: CATransform3D {
+        return CATransform3DConcat(bufferTransform, mirroredTransform)
+    }
 
     override init(frame: CGRect) {
         sampleBufferDisplayLayer = AVSampleBufferDisplayLayer()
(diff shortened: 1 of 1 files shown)


hiroshihorie (approved):
Thanks !, I was planning to fix this my self.
One concern is that I think we need to check all frames for rotation not only the first one, for example if iPhone rotates the camera the rotation will change mid-stream.

patryk-sredzinski:
@hiroshihorie Unfortunately I can see that occasionaly I still see the camera being flipped. I've got no clue why this sometimes happen.

I would not really like to check it every single frame, I'd better think of finding a solution and cases when this might change, but need to investigate it a bit more.

I'd like to continue my work on this

hiroshihorie:
Thanks, i can try to fix it when j have time.

PR #450, opened 2024-07-30
Prepare connection

hiroshihorie:


bcherry (approved):

Sources/LiveKit/Core/Room.swift
```diff
@@ -315,28 +324,35 @@ public class Room: NSObject, ObservableObject, Loggable {
...
+            $0.connectionState = .connecting
+        }
 
         do {
-            try await fullConnectSequence(url, token)
-
-            // Connect sequence successful
-            log("Connect sequence completed")
+            while true {
+                let region = try await resolveBestRegion()
```
lukasIO, line 335: would we really want to await region selection on initial connection? 
The other SDKs don't block here and just continue immediately in order for region manager not to have any impact on connection speed

Sources/LiveKit/Core/Room+Region.swift
```diff
@@ -0,0 +1,109 @@
...
+
+// MARK: - Room+Region
+
+extension Room {
+    static let defaultCacheInterval: TimeInterval = 3000
+
+    func resolveBestRegion() async throws -> RegionInfo {
+        try await requestRegionSettings()
+
+        let sortedByDistance = _regionState.remaining.sorted { $0.distance < $1.distance }
```
lukasIO, line 27: I think we should trust the server to have sorted the regions already. there might be other reasons for server ordering rather than just distance

lukasIO (approved):
lgtm!

pblazej:
@hiroshihorie I'm digging into the older (stale?) PRs, anything left on this one?

xianshijing-lk:
Hey, just giving this PR a gentle nudge,  Any updates here?

Sources/LiveKit/Core/RegionManager.swift
```diff
@@ -0,0 +1,219 @@
...
+        state = State()
+    }
+
+    func markFailed(region: RegionInfo) {
+        state.remaining.removeAll { $0 == region }
+    }
+
+    func shouldRequestSettings() -> Bool {
+        guard providedUrl.isCloud else { return false }
+        guard let lastRequested = state.lastRequested else { return true }
```
xianshijing-lk, line 63: Just to double check, is it intentional to return true when lastRequested does not exist ?

Sources/LiveKit/Core/RegionManager.swift
```diff
@@ -0,0 +1,219 @@
...
+    private var state = State()
+    private var settingsFetchTask: Task<Void, Error>?
+    private var settingsFetchTaskId: UUID?
+
+    init(providedUrl: URL) {
+        self.providedUrl = providedUrl
+    }
+
+    func cancel() {
+        settingsFetchTask?.cancel()
```
xianshijing-lk, line 40: [question](https://pullpo.io/cc?l=question) 

should you also update the lastRequested if settingsFetchTask is being cancelled ?

Sources/LiveKit/Core/RegionManager.swift
```diff
@@ -0,0 +1,219 @@
...
+
+        let taskId = UUID()
+        settingsFetchTaskId = taskId
+
+        let task = Task { [providedUrl, token, taskId] in
+            do {
+                let data = try await Self.fetchRegionSettings(providedUrl: providedUrl, token: token)
+                let allRegions = try Self.parseRegionSettings(data: data)
+                try Task.checkCancellation()
+                applyFetchedRegions(allRegions)
```
xianshijing-lk, line 123: nit, I think it should be Self.applyFetchedRegions(allRegions)

Sources/LiveKit/Core/RegionManager.swift
```diff
@@ -0,0 +1,219 @@
...
+                clearSettingsFetchTask(if: taskId)
+            } catch {
+                log("[Region] Failed to fetch region settings: \(error)", .error)
+                clearSettingsFetchTask(if: taskId)
```
xianshijing-lk: nit, move the clearSettingsFetchTask(if: taskId) outside the do {} catch {} section ? 

and use 
Self.clearSettingsFetchTask(if: taskId)

Sources/LiveKit/Core/RegionManager.swift
```diff
@@ -0,0 +1,219 @@
...
+                throw error
+            }
+        }
+
+        settingsFetchTask = task
+
+        Task { [weak self] in
+            _ = try? await task.value
+            // If the task failed before it could clear itself.
+            await self?.clearSettingsFetchTask(if: taskId)
```
xianshijing-lk: maybe I miss something, I think we call clearSettingsFetchTask(if: taskId) more than once here

Sources/LiveKit/Core/Room+Region.swift
```diff
@@ -0,0 +1,150 @@
...
+                    nextRegion = nil
+                    log("Connect failed with region: \(region)")
+                    await regionManager.markFailed(region: region)
+                }
+
+                try Task.checkCancellation()
+
+                await prepareAfterFailure()
+
+                let region = try await regionManager.resolveBest(token: token)
```
xianshijing-lk, line 159: [nitpick](https://pullpo.io/cc?l=nitpick) 

I wonder if we should do 
do {
  let region = try await regionManager.resolveBest(token: token)
  nextUrl = region.url
  nextRegion = region
} catch {
  // handle the error in some better way ? 
  throw error
}

xianshijing-lk (commented):
some nits

Sources/LiveKit/Core/RegionManager.swift
hiroshihorie: Yes, intentional. If lastRequested is nil, no settings have been fetched yet, so we should request them. This matches JS/Android behavior.

Sources/LiveKit/Core/RegionManager.swift
hiroshihorie: applyFetchedRegions is an instance method (modifies state), so Self. doesn't apply here, I think that's for static methods only.

Sources/LiveKit/Core/RegionManager.swift
hiroshihorie: Good catch! Fixed, now using defer for cleanup so it's called exactly once.

Sources/LiveKit/Core/RegionManager.swift
hiroshihorie: Intentionally not updating. Cancellation means we didn't get valid data: didn't complete the fetch. So we shouldn't block retry for 30 seconds. The next connect() should be allowed to fetch fresh region settings. 🤔

Sources/LiveKit/Core/Room+Region.swift
hiroshihorie: I think it's probably fine as-is 🤔 Resolving regions shouldn't really fail if there's connectivity, and if it does, we probably shouldn't continue attempting to connect anyway. I believe the JS SDK has similar logic. If you have a specific suggestion, let me know 🙂

PR #459, opened 2024-08-09
Add timestamps to TranscriptionSegment

bcherry:
This adds tracking of transcription receipt times which simplifies the consumption of this feature. I also removed startTime and endTime from the version we expose, since neither is used yet at the protocol level.

The new fields are `firstReceivedTime` and `lastReceivedTime`.

I tried this out in my sample project and it really simplified things. Now I can implement a very basic real-time chat thread without needing multiple data structures or new types. The only thing that still requires extra bookkeeping is tracking participant name, etc, since TranscriptionSegment lives under TrackPublication and Participant, and we don't denormalize the Participant id down the chain.

See the [associated commit in my docs PR](https://github.com/livekit/web/pull/315/commits/0076c753c9bb62b0bbe743f85c6a51f372fcd06b) for a visual of how this simplifies basic uses

### DIFF
--- a/Sources/LiveKit/Core/Room+EngineDelegate.swift
+++ b/Sources/LiveKit/Core/Room+EngineDelegate.swift
@@ -207,6 +207,11 @@ extension Room {
             return
         }
 
+        guard !packet.segments.isEmpty else {
+            log("[Transcription] Received segments are empty", .warning)
+            return
+        }
+
         let segments = packet.segments.map { segment in
             TranscriptionSegment(id: segment.id,
                                  text: segment.text,
@@ -216,14 +221,13 @@ extension Room {
                                  isFinal: segment.final)
         }
 
-        guard !segments.isEmpty else {
-            log("[Transcription] Received segments are empty", .warning)
-            return
-        }
-
         _state.mutate { state in
             for segment in segments {
-                state.transcriptionReceivedTimes[segment.id] = segment.firstReceivedTime
+                if segment.isFinal {
+                    state.transcriptionReceivedTimes.removeValue(forKey: segment.id)
+                } else {
+                    state.transcriptionReceivedTimes[segment.id] = segment.firstReceivedTime
+                }
             }
         }
 
--- a/Sources/LiveKit/Core/Room+EngineDelegate.swift
+++ b/Sources/LiveKit/Core/Room+EngineDelegate.swift
@@ -207,11 +207,24 @@ extension Room {
             return
         }
 
-        let segments = packet.segments.map { $0.toLKType() }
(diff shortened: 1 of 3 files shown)


hiroshihorie (approved):

Sources/LiveKit/Core/Room+EngineDelegate.swift
```diff
@@ -207,13 +207,26 @@ extension Room {
             return
         }
 
-        let segments = packet.segments.map { $0.toLKType() }
+        let segments = packet.segments.map { segment in
+            TranscriptionSegment(id: segment.id,
+                                 text: segment.text,
+                                 language: segment.language,
+                                 firstReceivedTime: _state.transcriptionReceivedTimes[segment.id] ?? Date(),
```
lukasIO, line 214: are you expecting both `startTime` and `firstReceivedTime` to have separate purposes? 
Otherwise we could simply override `startTime`  with `firstReceivedTime`?

Sources/LiveKit/Core/Room+EngineDelegate.swift
```diff
@@ -207,13 +207,26 @@ extension Room {
...
+        }
 
         guard !segments.isEmpty else {
             log("[Transcription] Received segments are empty", .warning)
             return
         }
 
+        _state.mutate { state in
+            for segment in segments {
+                state.transcriptionReceivedTimes[segment.id] = segment.firstReceivedTime
```
lukasIO, line 226: are we sure we want to keep all received transcription segements in memory indefinitely? 
For the components implementation I added an option to configure how many segments should be kept at a maximum before dropping the oldest ones in favor of newly arriving ones.

Sources/LiveKit/Core/Room+EngineDelegate.swift
bcherry: I don't know what startTime and endTime are intended for but I assumed they'd be in a track-local timekeeping system? They don't have a Date - it's UInt64. When we do eventually add support for them from the serve, would that cause a breaking change if we'd overloaded them already in the clients?

Sources/LiveKit/Core/Room+EngineDelegate.swift
bcherry: Good point - we can probably just drop them as soon as `final` comes over and then skip the configurable limits. I'll update it. (FWIW - this implementation is pretty lean as-is because it keeps only the timestamps. storing the segments themselves is left to the client or components framework)

bcherry: i changed it so they're wiped when final

PR #461, opened 2024-08-12
Fix multi-cam support

hiroshihorie:


bcherry (approved):
thanks! this also seems to fix a new crash i get sometimes in main if i start and stop sessions quickly

PR #469, opened 2024-08-28
Add video pausing option on VideoView

VatamanuBogdan:
I added a video pausing option on VideoView to facilitate complex animation synchronization on the client side.

### DIFF
--- a/Sources/LiveKit/Views/VideoView.swift
+++ b/Sources/LiveKit/Views/VideoView.swift
@@ -296,7 +296,7 @@ public class VideoView: NativeView, Loggable {
                         if let track = oldState.track as? VideoTrack {
                             track.remove(videoRenderer: self)
                             
-                            // avoid destroying and re-creating the
+                            // avoid destroying and re-creating the renderers when the video is paused
                             if !newState.isPaused {
                                 if let r = self._primaryRenderer {
                                     r.removeFromSuperview()


hiroshihorie:
Hi, thanks for your contribution. I think I put `isEnabled` for this purpose of pausing rendering. Doesn't it work well for you ?
Screenshot 2024-08-29 14 04 48

Instead of adding more properties, it would be great if you can optimize the logic of existing `isEnabled` property.

VatamanuBogdan:
Hi, thank you for the response. I tried to use `isEnabled` but when it's `false` the `VideoView` becomes transparent instead of pausing on the last frame.
If it's ok for you I can move the pausing logic I wrote to `isEnabled` to stop the `VideoView` on the last frame when it's `false` and to continue the rendering when it's `true`.

adrian-niculescu:
> Hi, thank you for the response. I tried to use `isEnabled` but when it's `false` the `VideoView` becomes transparent instead of pausing on the last frame. If it's ok for you I can move the pausing logic I wrote to `isEnabled` to stop the `VideoView` on the last frame when it's `false` and to continue the rendering when it's `true`.

Hi, @hiroshihorie, what do you think?

pblazej:
@VatamanuBogdan let's try to unblock/update this PR, `isEnabled` + stop sounds fine 👍

PR #474, opened 2024-09-04
Audio visualization helpers

hiroshihorie:


Sources/LiveKit/Support/FFTProcessor.swift
```diff
@@ -0,0 +1,188 @@
...
+
+    @inline(__always) private func _computeAverage(_ array: [Float], _ startIdx: Int, _ stopIdx: Int) -> Float {
+        var mean: Float = 0
+        let count = stopIdx - startIdx
+        array.withUnsafeBufferPointer { bufferPtr in
+            let ptr = bufferPtr.baseAddress! + startIdx
+            vDSP_meanv(ptr, 1, &mean, UInt(count))
+        }
+        return mean
+    }
```
bcherry, line 87: I think you can simplify this function using ArraySlice to give responsibility for safe pointer arithmetic back to swift instead:

```
    @inline(__always) private func _computeAverage(_ array: ArraySlice<Float>) -> Float {
        var mean: Float = 0
        array.withUnsafeBufferPointer { bufferPtr in
            vDSP_meanv(bufferPtr.baseAddress!, 1, &mean, vDSP_Length(array.count))
        }
        return mean
    }
```

then you call it above with `_computeAverage(magnitudes[magsStartIdx..<magsEndIdx])`

Sources/LiveKit/Support/FFTProcessor.swift
```diff
@@ -0,0 +1,188 @@
...
+    @inline(__always) private func _magnitudeIndex(for frequency: Float, sampleRate: Float) -> Int {
+        Int(Float(magnitudes.count) * frequency / sampleRate.nyquistFrequency)
+    }
+
+    @inline(__always) private func _computeAverage(_ array: [Float], _ startIdx: Int, _ stopIdx: Int) -> Float {
+        var mean: Float = 0
+        let count = stopIdx - startIdx
+        array.withUnsafeBufferPointer { bufferPtr in
+            let ptr = bufferPtr.baseAddress! + startIdx
+            vDSP_meanv(ptr, 1, &mean, UInt(count))
```
bcherry, line 84: `vDSP_Length(count)`

Sources/LiveKit/Support/FFTProcessor.swift
```diff
@@ -0,0 +1,188 @@
...
+        case linear
+        case logarithmic
+    }
+
+    public let bufferSize: Int
+    public let windowType: WindowType
+    public let scaleType: ScaleType
+
+    private let bufferHalfSize: Int
+    private let bufferLog2Size: Int
```
bcherry, line 116: you should consider declaring these int types as `vDSP_Length` instead. At the very least, they should be `UInt` here and cast to `vDSP_Length` when passed to accelerate later

Sources/LiveKit/Support/FFTProcessor.swift
```diff
@@ -0,0 +1,188 @@
...
+        imaginaryPointer.initialize(repeating: 0.0, count: bufferHalfSize)
+
+        complexBuffer = DSPSplitComplex(realp: realPointer, imagp: imaginaryPointer)
+        setupWindow()
+    }
+
+    deinit {
+        vDSP_destroy_fftsetup(fftSetup)
+        realPointer.deallocate()
+        imaginaryPointer.deallocate()
```
bcherry, line 147: is there a reason you're managing raw pointers in this class rather than just using arrays and nested withUnsafeMutableBufferPointer calls as in the apple sample code? looks scary to me :shrug: https://developer.apple.com/documentation/accelerate/vdsp/fast_fourier_transforms/finding_the_component_frequencies_in_a_composite_sine_wave

Sources/LiveKit/Support/FloatRingBuffer.swift
```diff
@@ -0,0 +1,51 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import Foundation
+
+// Simple ring-buffer used for internal audio processing. Not thread-safe.
+class FloatRingBuffer {
```
bcherry, line 20: doesn't really matter but you could call this `RingBuffer<T: Numeric>` and replace all of the `Float` with `T` below and it would still work as-is.

hiroshihorie:
@bcherry Optimized the code can you take a look 🙏

bcherry (approved):

PR #477, opened 2024-09-06
Accurate audio track counting

hiroshihorie:
~~Not sure what can cause trackDidStop to be called unbalanced with trackDidStart but better guard this.~~
Accurate counting of audio tracks instead of counting publish / unpublish.

### DIFF
--- a/Sources/LiveKit/Support/SerialRunnerActor.swift
+++ b/Sources/LiveKit/Support/SerialRunnerActor.swift
@@ -21,15 +21,26 @@ actor SerialRunnerActor<Value: Sendable> {
 
     func run(block: @Sendable @escaping () async throws -> Value) async throws -> Value {
         let task = Task { [previousTask] in
-            let _ = try? await previousTask?.value
+            // Wait for the previous task to complete, but cancel it if needed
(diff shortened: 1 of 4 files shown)


bcherry (approved):
is it still possible for a track that's already stopped to call `stopCapture`? in the case everything is already at 0 this won't matter, but if if the values are greater than 0 it could cause miscounting.

a more durable approach may be to store sets of the active track ids, instead of counts, and it will prevent miscounts in the case of extra starts or stops for a single track.

(But this looks like it will fix the issue as reported, where we reached -1 before anything got published, so this should be good to merge)

hiroshihorie:
start/stopCapture is guarded by a trackState lock so it should be ok:
https://github.com/livekit/client-sdk-swift/blob/8f3da0a8297c992450088169345b4fa8fc242de6/Sources/LiveKit/Track/Track.swift#L225-L251

But I agree there with more work I can make it more solid.

bcherry:
@hiroshihorie yeah that's perfect - this should probably work as it is then

hiroshihorie:
Ok when I look again it's still not race proof stop() could get executed concurrently

PR #481, opened 2024-09-10
Incremental migration to strict concurrency checking #1

hiroshihorie:


bcherry (approved):
nice, this will be a good improvement

PR #485, opened 2024-09-17
Fix loudspeaker usage during playback only

bcherry:
It seems that prior to https://github.com/livekit/client-sdk-swift/pull/477 a bug existed where mute/unmute would not necessarily cause audio session reconfiguring, which masked an issue with the session configuration.

To get the system to use the loudspeaker and receiver together, at max available volume, the only reliable method I have found is using the `.playAndRecord` category with the `.videoChat` mode. This was set when the microphone was unmuted, but upon mute it reverted to `.playback` with `.spokenAudio`.  That caused a significant drop in apparent audio as only the receiver would be used.

We already used `.playAndRecord` (with `.voiceChat`) in the case that you had a muted microphone but a preference to _not_ use the speaker, so it was possible to simplify things significantly here and unify the code.

I've tested it a bit but more testing and insight would be appreciated @hiroshihorie!

### DIFF
--- a/Sources/LiveKit/Track/AudioManager.swift
+++ b/Sources/LiveKit/Track/AudioManager.swift
@@ -210,7 +210,7 @@ public class AudioManager: Loggable {
             #if os(iOS)
             let configureFunc = newState.customConfigureFunc ?? self.defaultConfigureAudioSessionFunc
             configureFunc(newState, oldState)
-#endif
+            #endif
         }
     }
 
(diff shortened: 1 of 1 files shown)


hiroshihorie:
For use cases like live streaming watching, using playAndRecord means we can't lower the audio all the way down.

Is it when .play the audio isn't loud ?

bcherry:
closing in favor of https://github.com/livekit/client-sdk-swift/pull/486

PR #486, opened 2024-09-18
AudioManager adjustments

hiroshihorie:


bcherry (approved):

bcherry:
I think @davidzhao 's idea to offer a pinned category is a good alternative. basically instead of `alwaysUsePlayAndRecord` we'd offer `pinnedCategory` and that way you can pin it to anything you want. The default (nil) would be automatic and would maybe set it to soloAmbient -> playback or record -> playAndRecord based on capabilities enabled, and would never go back down the chain as they get unenabled within a session. 

worth considering and maybe more future-proof?

davidzhao:
> I think @davidzhao 's idea to offer a pinned category is a good alternative. basically instead of `alwaysUsePlayAndRecord` we'd offer `pinnedCategory` and that way you can pin it to anything you want. The default (nil) would be automatic and would maybe set it to soloAmbient -> playback or record -> playAndRecord based on capabilities enabled, and would never go back down the chain as they get unenabled within a session.
> 
> worth considering and maybe more future-proof?

maybe we can allow pinning a `LKRTCAudioSessionConfiguration` ? that can include all of the customizations with audio level.. and our job is to sync it with webrtc stack

bcherry:
Yeah that's also a better way to expose a lot of detailed configuration capability, without the customConfigureFunc which is pretty messy to actually use.

davidzhao:
> So a pinned configuration for all states ? or per state ?

IMO just a single state.. users could change it however they want.. but most likely my guess is they would stay in a single configuration.

for those that want dynamic configuration, they could already implement the function override.

Sources/LiveKit/Track/AudioManager.swift
```diff
@@ -121,6 +123,12 @@ public class AudioManager: Loggable {
         set { _state.mutate { $0.isSpeakerOutputPreferred = newValue } }
     }
 
+    /// If this is set, this will be used instead of dynamic configuration.
+    public var sessionConfiguration: AudioSessionConfiguration? {
```
bcherry, line 127: you should add a comment to the customConfigureFunc as well noting that if you set it then the sessionConfiguration and isSpeakerOutputPreferred properties will be ignored

bcherry (approved):
makes sense to me - great change

PR #506, opened 2024-10-25
Fix boundaries setting for "SampleBufferVideoRenderer" re-layouting

VatamanuBogdan:
Hello @hiroshihorie! 
I'm experiencing issues with the camera's front-to-back switch feature, which occasionally causes the video frame to be incorrectly oriented in landscape mode when it should be in portrait mode.

To resolve this issue, I've found that it's essential to update the frame of the **AVSampleBufferDisplayLayer** after applying the transformation matrix. This is because the transformation matrix also rotates the frame, which can result in incorrect orientation if the frame is not updated accordingly.


silviudeac (approved):

adrian-niculescu (approved):

hiroshihorie:
Thank you! Will test your fixes 🙏

hiroshihorie (approved):

PR #507, opened 2024-10-25
Remove leaked renderers appeared on camera switch

VatamanuBogdan:
Hello @hiroshihorie!

I found that inside **VideoView** on the camera switch the previous **primaryRenderer** gets leaked because it's not removed from the view's hierarchy when the transition ends. 

See:
image


hiroshihorie (approved):
Thank you !

PR #512, opened 2024-11-21
Camera frame cropping / scaling

hiroshihorie:
Initially I thought `adaptOutputFormat` would work better but it's not working as expected so I'm processing each frame.

### DIFF
--- a/Sources/LiveKit/Extensions/CustomStringConvertible.swift
+++ b/Sources/LiveKit/Extensions/CustomStringConvertible.swift
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 
+import AVFoundation
 import Foundation
 
 #if swift(>=5.9)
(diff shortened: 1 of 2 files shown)


Sources/LiveKit/Extensions/CustomStringConvertible.swift
```diff
@@ -166,3 +167,14 @@ extension LKRTCRtpEncodingParameters {
             ")"
     }
 }
+
+extension AVCaptureDevice.Format {
+    func toDebugString() -> String {
```
bcherry, line 172: maybe use CustomDebugStringConvertible and debugDescription?

bcherry (approved):
Are there any updates to docs / readme necessary to use this improvement?

davidliu:
what was the issue with `adaptOutputFormat`? Would've figured that would do exactly as needed.

hiroshihorie:
@davidliu Most of the time it doesn't respect the dimensions passed in the parameters (sometimes smaller, sometimes larger). Does it adapt exactly to what dimensions you provide in Android ?

PR #517, opened 2024-12-04
Don't show error dialog when broadcast stops without error (Swift unsafe cast)

bcherry:
When the user calls `setScreenShare(enabled: false)`, we stop the track and close the socket. When the socket is closed without error, we are currently showing a custom error that says "Screen sharing stopped". But this is unnecessary, the user knows they ended screensharing in this scenario (and the red icon is going away). 

[This stackoverflow question](https://stackoverflow.com/questions/63389328/how-to-programatically-stop-a-replaykit-2-system-broadcast-without-rpsystembroad) shows a workaround, which is to call the `finishBroadcastWithError` method with a null error (even though the header suggests error is non-nullable). This works perfectly in testing.

Rather than compiling an objective-c extension as suggested in the SO answer, we can simply cast the method to allow a nullable argument and call that.

I also added a hook to override this behavior in case users prefer to show an error message.

### DIFF
--- a/Sources/LiveKit/Broadcast/Uploader/LKSampleHandler.swift
+++ b/Sources/LiveKit/Broadcast/Uploader/LKSampleHandler.swift
@@ -72,6 +72,35 @@ open class LKSampleHandler: RPBroadcastSampleHandler {
         }
     }
 
+    /// Override point to change the behavior when the socket connection has closed.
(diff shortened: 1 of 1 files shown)


hiroshihorie (approved):
✅ Nice!
Though it’s pretty odd that this workaround is required in the first place. 🤔

PR #519, opened 2024-12-04
Don't show error dialog when broadcast stops without error (Obj-C bridge)

bcherry:
This is an alternative to #517 that uses the same Objective-C bridge as [used in Telegram](https://github.com/TelegramMessenger/Telegram-iOS/blob/9a46522b431782147cc1d95304180d91d398dd82/submodules/BroadcastUploadHelpers/Sources/BroadcastUploadHelpers.m), instead of some selector and IMP casting magic in Swift

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -28,11 +28,16 @@ let package = Package(
     ],
     targets: [
         .target(
+            name: "LKObjCHelpers",
+            publicHeadersPath: "include"
+        ),
+        .target(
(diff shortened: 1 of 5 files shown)


davidzhao (approved):
lg

hiroshihorie:
I had a Mac catalyst fix commit which i forgot to push but no prob. Will make another pr.

Issue #446, opened 2024-07-29
Stop Broadcast Extension by pressing indicator on native time doesn't make isScreenShareEnabled() to return false


mokuzuu:
**Describe the bug**
Even after stopping screen share by pressing indicator on native time doesn't make isScreenShareEnabled() to return false.

**SDK Version**
2.0.12

**iOS/macOS Version**
iOS 17.5.1

**Xcode Version**
15.4
Swift 5

**Steps to Reproduce**
1. Start broadcast extension by calling setScreenShare(enabled: true)
2. Stop screen share by pressing indicator on native time

**Expected behavior**
Expect `room.localParticipant.isScreenShareEnabled()` to return `false` but `true` is returned. 

**Screenshots**

**Logs**
Please provide logs if you can.

mokuzuu:
I think I could workaround this by listening to `iOS_BroadcastStopped` notification from CFNotificationCenter

(DarwinNotificationCenter is extended by following https://gist.github.com/tomlokhorst/7fe49a03b8bac960eeaf2b991faa3680 so that handler can use async)
```
 func notificationCallback() {
        DispatchQueue.global().async {
            Task {
                await performAsyncTask()
            }
        }
    }

DarwinNotificationCenter.shared.addObserver(name: "iOS_BroadcastStopped", callback: notificationCallback)
```

mokuzuu:
Probably I can address fix for #444 with similar approach

ashwin-nath-m:
@mokuzuu did u get this working? can u share your DarwinNotificationCenter class

mokuzuu:
I could workaround like this

```
 @State var darwinNotificationCenterObservation: DarwinNotificationObservation?

 // Define an async function to perform the async task
    func performAsyncTask() async {
        do {
            try await room.localParticipant.setScreenShare(enabled: false)
        } catch {
            print("An error occurred while updating screen share")
        }
    }
    
    // function that matches the C function pointer type
    func notificationCallback() {
        // Dispatch to handle async code
        DispatchQueue.global().async {
            Task {
                await performAsyncTask()
            }
        }
    }
    
    
    func setupNotifications() {
        // We need to store the value in variable/state to keep notification active.
        darwinNotificationCenterObservation = DarwinNotificationCenter.shared.addObserver(name: "iOS_BroadcastStopped", callback: notificationCallback)
    }
```

PR #520, opened 2024-12-04
Actually stop the track when the broadcast extension socket is closed

bcherry:
This closes https://github.com/livekit/client-sdk-swift/issues/446

Two fixes:

~1. It seems that stopping the broadcast via tapping on the system indicator and choosing "stop broadcast" calls the `broadcastPaused` method instead of the `broadcastFinished` method on the SampleHandler. The only way I could trigger `broadcastFinished` in testing was by locking my device. I could not find a way to trigger `broadcastResumed` so I just went ahead and treated "pause" as final, and close the socket (I think maybe this is vestigial from older iterations of ReplayKit? not sure, it's not an area I'm super familiar with)~
2. When the socket closes the frame reader stops but never notified the BroadcastScreenCapturer that owns it, so the track would stay published even as it stopped being updated.  This fix adds a new callback method to pass the closure up the stack and stop the track.

Not sure what happened in original testing but actually broadcastFinished _is_ called as expected, and broadcastPaused is called when you open the "stop capture" dialog, not when you finish it. I also fixed it to properly unpublish the track, not sure why initial testing showed simply ending capture to be enough.

### DIFF
--- a/Sources/LiveKit/Broadcast/BroadcastScreenCapturer.swift
+++ b/Sources/LiveKit/Broadcast/BroadcastScreenCapturer.swift
@@ -63,8 +63,9 @@ class BroadcastScreenCapturer: BufferCapturer {
             self.capture(pixelBuffer, rotation: rotation.toLKType())
         }
         frameReader.didEnd = { [weak self] in
+            guard let self else { return }
             Task {
-                try? await self?.stopCapture()
+                try await self.stopCapture()
             }
         }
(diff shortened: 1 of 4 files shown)


davidzhao (approved):

bcherry:
@davidzhao the original version was incomplete, can you take another look?

davidzhao (approved):
👍

PR #521, opened 2024-12-04
Fix Mac Catalyst errors

hiroshihorie:


bcherry:
This fails on the macos-13 builds because it doesn't know about visionOS. maybe we can leave visionOS off the list?

bcherry (approved):

PR #526, opened 2024-12-10
AVAudioEngine version

hiroshihorie:
Uses lib with : https://github.com/webrtc-sdk/webrtc/pull/158
Needs more testing with devices, please don't ship.

### TODO
- [ ] Allow custom audio session configuration


Sources/LiveKit/Track/AudioManager.swift
```diff
@@ -59,6 +59,29 @@ public class LKAudioBuffer: NSObject {
...
+        }
+
+        func audioSession(_: LKRTCAudioSession, audioEngineWillUpdateStateWithOutputEnabled isOutputEnabled: Bool, inputEnabled isInputEnabled: Bool) {
+            log("isOutputEnabled: \(isOutputEnabled), isInputEnabled: \(isInputEnabled)")
+
+            // Configure audio session
+            let config = LKRTCAudioSessionConfiguration.webRTC()
+            config.category = AVAudioSession.Category.playAndRecord.rawValue
+            config.categoryOptions = [.allowBluetooth, .allowBluetoothA2DP, .allowAirPlay]
+            config.mode = AVAudioSession.Mode.videoChat.rawValue
```
dfed, line 75: my 2c: it might be worth allowing end-users to customize these options. our current setup *requires* `.defaultToSpeaker`. We also did not allow airplay specifically because of issues with noise cancellation.

(THANK YOU for going down the AVAudioEngine rabbit hole. I know it's a huge one)

Sources/LiveKit/Track/AudioManager.swift
```diff
@@ -224,102 +249,38 @@ public class AudioManager: Loggable {
...
+    let _audioSessionDelegateObserver = AudioSessionDelegateObserver()
+    init() {
+        LKRTCAudioSession.sharedInstance().add(_audioSessionDelegateObserver)
+    }
 
-    #if os(iOS) || os(visionOS) || os(tvOS)
-    private func _asyncConfigure(newState: State, oldState: State) async throws {
-        try await _configureRunner.run {
-            self.log("\(oldState) -> \(newState)")
-            let configureFunc = newState.customConfigureFunc ?? self.defaultConfigureAudioSessionFunc
```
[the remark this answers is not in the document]
hiroshihorie: Yes this is not intended to merge "as-is", I'll need to figure out backward compatibility.
The track counting approach was not a good idea. Instead this version gets direct callback from the audio device module, that session configuration is required.

Sources/LiveKit/Track/AudioManager.swift
hiroshihorie: Yes, I will allow devs to customize the session. 🙂

hiroshihorie:
Newer version https://github.com/livekit/client-sdk-swift/pull/536

PR #531, opened 2024-12-12
Warn if you set bypassVoiceProcessing after creation of PeerConnectionFactory

bcherry:
This is an implementation trap as it is, since this property has no effect once the PCF is initialized, and the PCF is also a static property that gets initialized very early in app lifecycle.

In this PR I've just added a warning log, but alternatives worth considering:
- Change the API to be a throwing function and deprecate the original property
- Recreate the PCF if you change this property (implications unknown to me)


davidzhao (approved):
👍

hiroshihorie (approved):
✅

PR #538, opened 2025-01-07
Bug unsupported file format 'public.heic'

creativecbr:
While implementing remote sharing using LiveKit, we encountered an error on the iPad 5th generation running iPadOS 16.7.10 when converting a buffer to a JPEG representation. The suggested solution is to avoid using the native .jpgedRepresentation method, as it appears to contain a bug, and instead handle the conversion manually. After applying this approach, the error seems to be resolved.

The error logged in the Console app was:
```
Error    11:20:25.613787+0100    RemoteSharing    findWriterForTypeAndAlternateType:119: unsupported file format 'public.heic'
```

This issue did not occur on newer devices with higher iOS versions and it worked correctly on newer devices like the iPhone SE (2nd generation) with iOS 15.5. This suggests the problem is specific to platform 16.7.10.

This error caused the screen-sharing session to be interrupted after several occurrences. After changes session was stable.

Please review this change and consider potential improvements.

PS: I have also removed the code responsible for scale factor buffering, as I did not observe any usage of it. If this removal was a mistake, please let me know.

### DIFF
--- a/Sources/LiveKit/Broadcast/Uploader/SampleUploader.swift
+++ b/Sources/LiveKit/Broadcast/Uploader/SampleUploader.swift
@@ -17,7 +17,6 @@
 #if os(iOS)
 
 import Foundation
-import UniformTypeIdentifiers
 
 #if canImport(ReplayKit)
 import ReplayKit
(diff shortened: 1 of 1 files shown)


hiroshihorie (approved):
✅

PR #544, opened 2025-01-14
Implement RPC feature

bcherry:
Pretty much identical to https://github.com/livekit/client-sdk-android/pull/578

I included tests which mock the underlying data channel sending and receiving to test both sides in the same SDK app instance.


hiroshihorie (approved):
✅ Looks good !
Applied format & moved code to a separate file (extension) to organize.

Was just wondering if the Public APIs are "Swifty" enough 🧐

bcherry:
> Was just wondering if the Public APIs are "Swifty" enough 🧐

I could make it more swifty like this: `register(rpcMethod:"hello") { data in }` and `perform(rpcMethod: "hello", on: "some-identity", payload: "xxx", timeout: 8)`?

Only downside is it loses consistency with other sdks in parameter structure. i was also thinking that maybe it would be better to revise across all SDKs to a separate construct attached to room e.g. `room.rpc.register("xxx") { _ in }` and `room.rpc.perform("xxx", on: "yyy")`

bcherry:
> > Was just wondering if the Public APIs are "Swifty" enough 🧐
> 
> I could make it more swifty like this: `register(rpcMethod:"hello") { data in }` and `perform(rpcMethod: "hello", on: "some-identity", payload: "xxx", timeout: 8)`?
> 
> Only downside is it loses consistency with other sdks in parameter structure. i was also thinking that maybe it would be better to revise across all SDKs to a separate construct attached to room e.g. `room.rpc.register("xxx") { _ in }` and `room.rpc.perform("xxx", on: "yyy")`

@hiroshihorie I made those changes then reverted them while writing the docs because the way the docs are written it really is a lot nicer if all platforms have the same method names

PR #550, opened 2025-01-16
Improve handling of bundle settings

ladvoc:
Summary of changes:
- Report error when required bundle settings are not in place for broadcast screen capture.
- Introduce `BundleInfo` property wrapper to reduce code duplication.


hiroshihorie (approved):
✅ Looks neater !

PR #551, opened 2025-01-18
Properly handle broadcast capture state

### LINKED ISSUE #444: Dismissing RPSystemBroadcastPickerView without selecting Broadcast Extension leaves isScreenShareEnabled() true
**Describe the bug**
Dismissing RPSystemBroadcastPickerView without selecting Broadcast Extension leaves isScreenShareEnabled() true

**SDK Version**
2.0.12

**iOS/macOS Version**
iOS 17.5.1

**Xcode Version**
15.4
Swift 5

**Steps to Reproduce**
1. Setup broadcast extention
2. Call `room.localParticipant.setScreenShare(true)`, now `room.localParticipant.isScreenShareEnabled()` returns `true`
3. It shows RPSystemBroadcastPickerView, and press anywhere except Picker to dismiss it
4. I expect `room.localParticipant.isScreenShareEnabled()` returns `false` but it returns `true`

**Expected behavior**
Dismissing RPSystemBroadcastPickerView should update `isScreenShareEnabled()` returns `false`.

**Screenshots**

**Logs**
Please provide logs if you can.

### LINKED ISSUE #472: iOS screenshare room.localParticipant.isScreenShareEnabled() is returning true  even if start broadcast is not pressed.
When I trigger the start screen share function using the broadcast extension,    
  try await room.localParticipant.setScreenShare(enabled:true), but before even I select startBroadcast in the prompt view for screenshare, the didPublish callback is geting called and room.localParticipant.isScreenShareEnabled() is returning true.

**SDK Version**
2.0.13

**iOS Version** 17.5.1
The OS version which the issue occurs.

**Xcode Version**
15.4

**Steps to Reproduce**
Steps to reproduce the behavior.

**Expected behavior**

**Screenshots**

**Logs**
Please provide logs if you can.

ladvoc:
When using the broadcast capturer on iOS, the broadcast extension now drives whether or not screen sharing is enabled, publishing a screen sharing track when the extension begins broadcasting upon user approval. Additionally, the new `BroadcastManager` class gives developers control over the broadcast state and track publication.

Fixes #444 and fixes #472. 

**Minor breaking changes**
- Calling `setScreenShare(enabled: true)` or `set(source:enabled:captureOptions:publishOptions:)` on `LocalParticipant` currently returns a `LocalTrackPublication` representing the newly published screen share track. After this change, when using the broadcast capturer on iOS, this method will return `nil`, as the track is published asynchronously pending user approval. Developers should treat enabling screen share as a request that might not be fulfilled and should not interpret a `nil` return value from this method as an error.
- Since track publication is now asynchronous, capture options must be set as room defaults rather than being passed to `set(source:enabled:captureOptions:publishOptions:)` when enabling screen sharing.


hiroshihorie (approved):
✅

Sources/LiveKit/Participant/LocalParticipant.swift
```diff
@@ -229,6 +226,59 @@ public class LocalParticipant: Participant {
...
+
+    /// An optional hook called just after a broadcast starts.
+    ///
+    /// Returns a Boolean value indicating weather or not the broadcast should be
+    /// published as a screen share track. This could be useful to present a confirmation dialog,
+    /// returning `true` or `false` based on the user's response .
+    ///
+    public var broadcastStarted: (() async -> Bool)?
+
+    private var isBroadcasting = false {
```
bcherry, line 242: i think this or something like it needs to be publicly gettable, because developers may have UI that hinges on this state, in the case that they aren't starting publishing right away

bcherry (changes requested):
I think this looks pretty good and seems to be the right approach. 

One design question besides what's in the review comments is how does a developer "stop" a broadcast if they don't want it? Right now, their control is only over whether it is published or not. there needs to be a way to stop it, too... 

Another question is is this a breaking change? it looks like maybe it's actually not a breaking change and it will "just work"?

ladvoc:
> I think this looks pretty good and seems to be the right approach.
> 
> One design question besides what's in the review comments is how does a developer "stop" a broadcast if they don't want it? Right now, their control is only over whether it is published or not. there needs to be a way to stop it, too...
> 
> Another question is is this a breaking change? it looks like maybe it's actually not a breaking change and it will "just work"?



> I think this looks pretty good and seems to be the right approach.
> 
> One design question besides what's in the review comments is how does a developer "stop" a broadcast if they don't want it? Right now, their control is only over whether it is published or not. there needs to be a way to stop it, too...
> 
> Another question is is this a breaking change? it looks like maybe it's actually not a breaking change and it will "just work"?

Currently, there is no method to stop a broadcast other than by unpublishing the screen share track. To address your other comments, I am adding a method to stop the broadcast even if a track has not been published.

Regarding whether this constitutes a breaking change, there are a few points to consider. At present, when calling `setScreenShare(enabled: true)`, a `LocalTrackPublication?` is returned, representing the newly published screen share track. After this change, however, when using the broadcast capturer on iOS, the result of this call will be `nil`, as the track is published asynchronously. Since the return type is already optional, some developers may rely on this return value to determine whether the screen share was successful—even though in this case, `nil` does not indicate an error.

Please let me know your thoughts on this.

bcherry:
> Currently, there is no method to stop a broadcast other than by unpublishing the screen share track. To address your other comments, I am adding a method to stop the broadcast even if a track has not been published.

yeah it makes sense to add something new since broadcast is no longer explicitly tied to publication.

> Regarding whether this constitutes a breaking change, there are a few points to consider. At present, when calling setScreenShare(enabled: true), a LocalTrackPublication? is returned, representing the newly published screen share track. After this change, however, when using the broadcast capturer on iOS, the result of this call will be nil, as the track is published asynchronously. Since the return type is already optional, some developers may rely on this return value to determine whether the screen share was successful—even though in this case, nil does not indicate an error.

that does sound like a minor breaking change so we'll be sure to update the sdk minor version and include a clear note about it in the release notes.  since the default "detailed changelog" is just the git history, can you ensure the description on this PR includes notes about how this would affect existing apps and what they need to change to adapt? that will then go into the sqaushed commit as a description and be linked in the history as well

Sources/LiveKit/Participant/LocalParticipant.swift
```diff
@@ -229,6 +226,59 @@ public class LocalParticipant: Participant {
...
+
+        Task { [weak self] in
+            guard let self else { return }
+            let shouldPublish = await self.broadcastStarted?() ?? true
+            guard shouldPublish else {
+                logger.debug("Will not publish screen share track")
+                return
+            }
+            do {
+                try await self.setScreenShare(enabled: true)
```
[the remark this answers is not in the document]
ladvoc: `setScreenShare(enabled:)` calls down to `set(source:enabled:captureOptions:publishOptions:)` which in turn calls `requireRoom()`, performing a check to ensure the room isn't `nil`. I will need to perform some more examination to see if additional checks are necessary.

Sources/LiveKit/Participant/LocalParticipant.swift
ladvoc: Is it okay if we expose a Combine `AnyPublisher<Bool, Never>` in the public API? If so, I could see making the `isBroadcasting` property a `CurrentValueSubject` which is type erased into `AnyPublisher`. This would allow both getting the current value and listening for changes. We could then replace the `broadcastStarted` closure with a Boolean `publishOnBroadcast` property which allows the developer to disable the default behavior to manually control track publication. The public interface would then look something like this:

```swift
public var publishOnBroadcast: Bool
public var isBroadcasting: AnyPublisher<Bool, Never>
```

bcherry: it's an interesting idea and i'd defer to @hiroshihorie but my instinct is that we should consider introducing new patterns carefully and deliberately. i'm not sure it's worth using something novel here that is unlike other things in our SDK, especially because I'm not sure there's a really strong reason we'd need to besides that it might be a little nicer.

we use a delegate pattern everywhere else. maybe good to make a new BroadcastDelegate that has this and the onStarted method in it

bcherry:
looks like this will also close https://github.com/livekit/client-sdk-swift/issues/472

Sources/LiveKit/Participant/LocalParticipant.swift
hiroshihorie: I think @bcherry is right, I'm open to adding a AnyPublisher **in addition** to current design of reading value synchronously and notifying with delegate perhaps.

ladvoc: Just to make sure I’m following you, are you thinking the public interface would look something like this then?
```swift
public var isBroadcasting: Bool
public var isBroadcastingPublisher: AnyPublisher<Bool, Never>
```
This would be in addition to the delegate methods.

ladvoc: I have implemented the requested changes and added some tests. Additionally, I refactored the public-facing API into a dedicated class, `BroadcastManager`. The public interface for this class is as follows:

```swift
/// Manages the broadcast state and track publication for screen sharing on iOS.
public final class BroadcastManager : Sendable {

    public static let shared: BroadcastManager

    /// A delegate for handling broadcast state changes.
    public var delegate: BroadcastManagerDelegate? { get set }

    /// Indicates whether a broadcast is currently in progress.
    public var isBroadcasting: Bool { get }

    /// A publisher that emits the current broadcast state as a Boolean value.
    public var isBroadcastingPublisher: AnyPublisher<Bool, Never> { get }

    /// Determines whether a screen share track should be automatically published when broadcasting starts.
    ///
    /// Set this to `false` to manually manage track publication when the broadcast starts.
    ///
    public var shouldPublishTrack: Bool { get set }

    /// Displays the system broadcast picker, allowing the user to start the broadcast.
    ///
    /// - Note: This is merely a request and does not guarantee the user will choose to start the broadcast.
    ///
    public func requestActivation()

    /// Requests to stop the broadcast.
    ///
    /// If a screen share track is published, it will also be unpublished once the broadcast ends.
    /// This method has no effect if no broadcast is currently in progress.
    ///
    public func requestStop()
}

/// A delegate protocol for receiving updates about the broadcast state.
@objc public protocol BroadcastManagerDelegate {

    func broadcastManager(didChangeState isBroadcasting: Bool)
}
```
Let me know if you have any further questions or suggestions.

ladvoc:
I have implemented the requested changes and added some tests. Additionally, I refactored the public-facing API into a dedicated class, `BroadcastManager`. The public interface for this class is as follows:

```swift
/// Manages the broadcast state and track publication for screen sharing on iOS.
public final class BroadcastManager : Sendable {

    public static let shared: BroadcastManager

    /// A delegate for handling broadcast state changes.
    public var delegate: BroadcastManagerDelegate? { get set }

    /// Indicates whether a broadcast is currently in progress.
    public var isBroadcasting: Bool { get }

    /// A publisher that emits the current broadcast state as a Boolean value.
    public var isBroadcastingPublisher: AnyPublisher<Bool, Never> { get }

    /// Determines whether a screen share track should be automatically published when broadcasting starts.
    ///
    /// Set this to `false` to manually manage track publication when the broadcast starts.
    ///
    public var shouldPublishTrack: Bool { get set }

    /// Displays the system broadcast picker, allowing the user to start the broadcast.
    ///
    /// - Note: This is merely a request and does not guarantee the user will choose to start the broadcast.
    ///
    public func requestActivation()

    /// Requests to stop the broadcast.
    ///
    /// If a screen share track is published, it will also be unpublished once the broadcast ends.
    /// This method has no effect if no broadcast is currently in progress.
    ///
    public func requestStop()
}

/// A delegate protocol for receiving updates about the broadcast state.
@objc public protocol BroadcastManagerDelegate {

    func broadcastManager(didChangeState isBroadcasting: Bool)
}
```
Let me know if you have any further questions or suggestions.

Sources/LiveKit/Participant/LocalParticipant.swift
```diff
@@ -339,10 +381,11 @@ public extension LocalParticipant {
...
                     if options.useBroadcastExtension {
-                        await RPSystemBroadcastPickerView.show(
-                            for: BroadcastScreenCapturer.screenSharingExtension,
-                            showsMicrophoneButton: false
-                        )
+                        guard BroadcastManager.shared.isBroadcasting else {
+                            BroadcastManager.shared.requestActivation()
+                            return nil
+                        }
+                        // Wait until broadcasting to publish track
```
bcherry, line 388: you either need to store the passed options somewhere or a log a warning that custom options should _not_ be passed here when using broadcast extension (if the passed options differ from the default options on anything besides useBroadcast:true)

Sources/LiveKit/Types/Options/RoomOptions.swift
```diff
@@ -72,6 +72,10 @@ public final class RoomOptions: NSObject, Sendable {
     override public init() {
         defaultCameraCaptureOptions = CameraCaptureOptions()
         defaultScreenShareCaptureOptions = ScreenShareCaptureOptions()
+        #if os(iOS)
+            .settingUseBroadcastExtension(BroadcastBundleInfo.hasExtension)
+        #endif
```
bcherry, line 77: this would be better set in the default constructor for ScreenshareCaptureOptions than in the default constructor for Room only

Sources/LiveKit/Participant/LocalParticipant.swift
```diff
@@ -229,6 +226,51 @@ public class LocalParticipant: Participant {
...
+        Task { [weak self] in
+            guard let self else { return }
+            let room = try self.requireRoom()
+
+            guard BroadcastManager.shared.shouldPublishTrack else {
+                logger.debug("Will not publish screen share track")
+                return
+            }
+            let captureOptions = room._state.roomOptions.defaultScreenShareCaptureOptions
+                .settingUseBroadcastExtension(true)
```
bcherry, line 261: i'd just put the capture options in a var and change the value rather than add a helper method for this

bcherry (changes requested):
yeah looks very close!

Sources/LiveKit/Participant/LocalParticipant.swift
ladvoc: That was my first thought, but `ScreenShareCaptureOptions` is defined as a class for Objective-C compatibility, and all its properties are let constants—changing any to var would break `Sendable` conformance.

bcherry: you could still do something like this in ScreenShareCaptureOptions

```
    static let defaultToBroadcastExtension = () {
        #if os(iOS)
        return BroadcastBundleInfo.hasExtension
        #else
        return false
        #endif
    }()
    
    public init(dimensions: Dimensions = .h1080_169,
                fps: Int = 30,
                showCursor: Bool = true,
                useBroadcastExtension: Bool = defaultUseBroadcastExtension,
                includeCurrentApplication: Bool = false)
    {
     /// ...
    }
```

Sources/LiveKit/Participant/LocalParticipant.swift
```diff
@@ -378,16 +372,22 @@ public extension LocalParticipant {
...
                     #if os(iOS)
+                    
                     let localTrack: LocalVideoTrack
-                    let options = (captureOptions as? ScreenShareCaptureOptions) ?? room._state.roomOptions.defaultScreenShareCaptureOptions
-                    if options.useBroadcastExtension {
+                    let defaultOptions = room._state.roomOptions.defaultScreenShareCaptureOptions
+                    
+                    if defaultOptions.useBroadcastExtension {
+                        if captureOptions != nil {
+                            logger.warning("Capture options for broadcast screen share must be set as room defaults.")
```
bcherry, line 381: be more specific with method names etc here and say what will happen instead (which will be e.g. "using room defaults instead")

bcherry (approved):
just tested, this works great. last thing you should do before we merge it is can you go also include any updates to the doc you wrote that are needed? It seems like a very minimum it needs to be clear that you should prefer to set screenshare options on the room instead of in the call.

ladvoc:
> just tested, this works great. last thing you should do before we merge it is can you go also include any updates to the doc you wrote that are needed? It seems like a very minimum it needs to be clear that you should prefer to set screenshare options on the room instead of in the call.

That is a good idea. I am also documenting the use of `BroadcastManager`.

bcherry (approved):

PR #555, opened 2025-01-20
Simplify broadcast extension setup

ladvoc:
The broadcast extension [in the example app](https://github.com/livekit-examples/swift-example/tree/main/iOS/BroadcastExt) currently defines a log handler for OSLog and bootstraps the logging system using it. This setup allows log output to be viewed in Console for troubleshooting the broadcast extension.

Since this configuration is generally useful across applications, I have integrated this functionality into `LKSampleHandler` and added a property that subclasses can override to opt in. By default, this feature is disabled to ensure compatibility with apps that manually bootstrap the logging system, as the example app currently does. With this change, the boilerplate required to set up logging for the broadcast extension is significantly reduced:
```swift
import LiveKit

#if os(iOS)
@available(macCatalyst 13.1, *)
class SampleHandler: LKSampleHandler {
    override var enableLogging: Bool { true }
}
#endif
```
If this PR is approved, I will open a separate PR updating the code in the example app to reflect the example above.

### DIFF
--- a/.github/workflows/testing-matrix.yaml
+++ b/.github/workflows/testing-matrix.yaml
@@ -12,7 +12,7 @@ concurrency:
   cancel-in-progress: true
 
 env:
-  TESTS: "FunctionTests SDKTests RoomTests PublishDataTests PublishBufferCapturerTests VideoViewTests TrackTests"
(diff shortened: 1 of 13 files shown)


bcherry (approved):

hiroshihorie (approved):
✅

ladvoc:
I resolved two compilation issues for older versions of Swift. It looks like the remaining failure is due to the `RoomTests.testParticipantCleanUp` test failing on visionOS.

PR #561, opened 2025-01-23
macOS screen share audio

hiroshihorie:


ladvoc:
While testing this in the example app, I came across an issue where screen share audio still plays on the remote participant's device even after screen sharing is stopped. 

Steps to reproduce: 
1. Join a room using both a Mac and another device.
2. Play an audio source on the Mac.
3. Enable screen sharing on the Mac, then enable the microphone to publish an audio track.
4. Wait ~20 seconds, then disable screen sharing.
5. Stop the audio source on the Mac.
6. Re-enable screen sharing, and listen to the audio from the second device—you will still hear the previous screen share audio even though it is no longer playing.

hiroshihorie:
@ladvoc I think the audio from the screen share picker was being captured and causing confusion.
Can you try this branch: https://github.com/livekit-examples/swift-example/pull/64
I also changed capturing app audio to false by default.

ladvoc:
Yep, that fixed it

PR #565, opened 2025-01-29
Revamp broadcast IPC

ladvoc:
This PR revamps inter-process communication between the main app and the broadcast extension, replacing deprecated Core Foundation APIs. Additionally, it enhances testability, enables bi-directional communication, and introduces a more flexible message structure.

Summary of new types:
- `IPCChannel`: An abstraction for inter-process communication, built on top of the [Network](https://developer.apple.com/documentation/network) framework. Allows asynchronous sending and receiving of messages consisting of a dynamic header (any type conforming to `Codable`) and a data payload.
- `BroadcastImageCodec`: Encapsulates functionality for encoding/decoding image samples for transport. For now, this uses the same method of JPEG encoding/decoding as the current implementation.
- `BroadcastUploader`: Sends samples from ReplayKit to the main app, built on top of `IPCChannel`.
- `BroadcastReceiver`: Receives samples from the broadcast extension, built on top of `IPCChannel`.

This PR sets the groundwork for adding support for audio samples, but I decided to submit that in a separate PR (#576) as this is a large change (though not breaking).


bcherry (commented):
this looks pretty good, no comments on it from a first pass read-through. i'm going to come back once the other PR lands and you merge it in here and I will test it and then leave more comments if needed

Sources/LiveKit/Broadcast/BroadcastScreenCapturer.swift
```diff
@@ -53,22 +48,32 @@ class BroadcastScreenCapturer: BufferCapturer {
...
         }
-        frameReader.didEnd = { [weak self] in
-            guard let self else { return }
-            Task {
-                try await self.stopCapture()
+        Task { [weak self] in
+            do {
+                let receiver = try await BroadcastReceiver(socketPath: socketPath)
+                logger.debug("Broadcast receiver connected")
+                self?.receiver = receiver
```
bcherry, line 63: does this have to be done within the task?  it would be better if creating the receiver and assigning it to self could be done synchronously within createReceiver

Sources/LiveKit/Broadcast/BroadcastScreenCapturer.swift
```diff
@@ -53,22 +48,32 @@ class BroadcastScreenCapturer: BufferCapturer {
             .toEncodeSafeDimensions()
 
         set(dimensions: targetDimensions)
+        return createReceiver()
```
bcherry, line 51: might be worth adding guards somewhere that there isn't already a receiver set up

Sources/LiveKit/Broadcast/IPC/BroadcastImageCodec.swift
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    func decode(_ encodedData: Data, with metadata: Metadata) throws -> CVPixelBuffer {
+        try jpegDecode(encodedData, metadata)
+    }
+
+    private let imageContext = CIContext(options: nil)
+    private let colorSpace = CGColorSpaceCreateDeviceRGB()
+
+    private func jpegEncode(_ imageBuffer: CVImageBuffer) throws -> Data {
```
bcherry, line 62: why do we convert to jpeg at all here? why not just dump the sample buffers directly?

Sources/LiveKit/Broadcast/IPC/IPCChannel.swift
```diff
@@ -0,0 +1,269 @@
...
+    func incomingMessages<T: Decodable>(_: T.Type) -> AsyncMessageSequence<T> {
+        AsyncMessageSequence(upstream: connection.incomingMessages)
+    }
+}
+
+// MARK: - Extensions
+
+private extension Data {
+    func partition(bytesInFirst: Int) -> (Data, Data) {
+        guard bytesInFirst > 0 else { return (Data(), self) }
```
bcherry, line 151: wouldn't it make more sense to just check `bytesInFirst > self.length`?

Sources/LiveKit/Broadcast/IPC/IPCChannel.swift
```diff
@@ -0,0 +1,269 @@
...
+
+private extension NWConnection {
+    func waitUntilReady() async throws {
+        for await state in stateUpdates {
+            switch state {
+            case .ready: return
+            case .setup, .preparing: continue
+            case .waiting:
+                // Will enter this state when socket path does not exist yet
+                let restartDelay = UInt64(IPCChannel.restartDelay * 1_000_000_000)
```
bcherry, line 193: use [NSEC_PER_SEC](https://developer.apple.com/documentation/dispatch/nsec_per_sec)

bcherry (approved):
looks good, just a handful of small comments

Sources/LiveKit/Broadcast/IPC/BroadcastImageCodec.swift
ladvoc: This approach isn’t ideal, but for this PR I kept the focus solely on IPC. I used the same JPEG encoding method as the original implementation for consistency. In the future, I’d like to explore alternatives—directly dumping the sample buffer would be ideal if the socket bandwidth allows.

Sources/LiveKit/Broadcast/BroadcastScreenCapturer.swift
ladvoc: This is a result of a design decision I made with `IPCChannel`, which `BroadcastReceiver` uses under the hood. Specifically, I designed `IPCChannel` to use the RAII pattern, making its initialization asynchronous so that the consumer always receives an open channel after initialization. This approach also allows `IPCChannel` to conform to `Sendable` without additional synchronization, as the underlying `NWConnection` is both a constant (let) and `Sendable`. I am open to feedback about this design.

PR #569, opened 2025-01-29
Make some audio config extensions public

bcherry:
This makes it easier to implement a custom configure function by copying the default one, which relies on these non-public extensions to function. Now you can drop in the default and the only thing to change is the log lines.


hiroshihorie (approved):

hiroshihorie:
Oh, I think private symbols are being exposed so It's giving errors, I'll check this out later.

bcherry:
@hiroshihorie weird - just copy-pasting these into my own project is fine without errors 🤷🏻

ladvoc:
It looks like there are still some outstanding compilation issues.

PR #573, opened 2025-02-04
Simplify broadcast extension setup with standard format for identifiers

ladvoc:
When configuring a broadcast extension, manually setting the info keys `RTCAppGroupIdentifier` and `RTCScreenSharingExtension` is no longer required when using the standard format. The standard format is as follows:
- App group: `group.<main-app-bundle-id>`
- Broadcast extension: `<main-app-bundle-id>.broadcast`


Docs/ios-screen-sharing.md
```diff
@@ -129,3 +128,15 @@ class MyDelegate: BroadcastManagerDelegate {
 }
 BroadcastManager.shared.delegate = MyDelegate()
 ```
+
+#### Custom Identifiers
+
+By default, the app group identifier and broadcast extension bundle identifier must follow the format prescribed in the setup guide. If this is not suitable for your app, you can override the defaults:
```
bcherry, line 134: rather than say "the format prescribed in the setup guide", just describe it directly here. the whole thing could be a lot more succinct

> By default the app group identifier is expected to be `group.<main-app-bundle-id>` and the broadcast extension is expected to be `<main-app-bundle-id>.broadcast`. To override these values, set `RTCAppGroupIdentifier` in your Info.plist (both broadcast extension and main application) and `RTCScreenSharingExtension` in your main app's Info.plist.

Docs/ios-screen-sharing.md
```diff
@@ -44,15 +44,17 @@ flowchart LR
...
 
 <img src="Resources/new-target-options.png" width="500" />
 
 1. In Xcode, Choose "File" > "New > "Target"
 2. From the template chooser, select "Broadcast Upload Extension"
-3. Name the extension (e.g. "BroadcastExtension"). Take note of the extension's bundle identifier, as it will be needed later.
-4. Replace the default content of `SampleHandler.swift` in the new target with the following:
+3. Name the extension (e.g. "BroadcastExtension").
+4. Press "Finish"
+5. From the "Signing & Capabilities" tab of the new target, change the bundle identifier to `<BUNDLE_ID>.broadcast`, replacing `<BUNDLE_ID>` with the bundle identifier of your primary app target.
```
bcherry, line 56: I'd say "set it's bundle identifier to be the same as your main app, with `.broadcast` added at the end (to use a different identifier, see Custom Identifiers below)"

bcherry (approved):
great and thank you for updating the docs!

PR #575, opened 2025-02-04
Expose RPC types

hiroshihorie:


bcherry (approved):
thank you for fixing this! I'm really not sure how my tests worked at all...

hiroshihorie:
@bcherry Could you take a quick look if exposing these are sufficient ? 🙏

bcherry (approved):
yeah this seems to work

natecraft1:
> thank you for fixing this! I'm really not sure how my tests worked at all...

I think it's because you have @testable import LiveKit in the tests which exposes all the internal types. I wonder if removing @testable would have prevented the test from compiling

PR #576, opened 2025-02-05
iOS screen share audio

ladvoc:
This PR adds support for capturing application audio during screen share when using a broadcast extension.

Note: PR #598 should be merged first.

PR #579, opened 2025-02-06
Release automation

ladvoc:
- Add version and platform compatibility badges from [Swift Package Index](https://swiftpackageindex.com/) to README
  - Automatically updated on each release 
- Add [nanpa](https://github.com/nbsp/nanpa) configuration
  - Custom script bumps version across repo (currently Podspec, README, and LiveKitSDK class)
  - GitHub publish workflow (based on workflow from livekit/rust-sdks)
- Create workflow to push new releases to Cocopods when a release is published on GitHub

I have also added changeset files to my currently open PRs (#565, #576, and #573) that can be used to test this configuration.


.github/workflows/cocoapods.yaml
```diff
@@ -0,0 +1,20 @@
+name: Push to Cocoapods
+on:
+  workflow_dispatch:
+  push:
+    tags: ['*']
```
bcherry, line 5: better to use a release trigger? that way only actually publishing the release via github causes it to get published to cocoapods https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#release

hiroshihorie (approved):

Issue #566, opened 2025-01-29
LiveKitClient.podspec don't have `tvos.deployment_target`


joohae-kim:
**Describe the bug**
LiveKitClient.podspec don't have `tvos.deployment_target`, it's not possible to import from cocoapods

**SDK Version**
2.0.20

**iOS/macOS Version**
tvOS 18.2.1

**Xcode Version**
N/A

**Steps to Reproduce**
add `pod 'LiveKitClient', :git => 'https://github.com/livekit/client-sdk-swift', :branch=> 'main'` to the `Podfile` and run `pod install`

**Expected behavior**
LiveKitClient can be installed for tvOS

**Screenshots**
N/A

**Logs**
N/A

PR #587, opened 2025-02-13
Update Podspec

hiroshihorie:
Fixes: https://github.com/livekit/client-sdk-swift/issues/566

### DIFF
--- a/LiveKitClient.podspec
+++ b/LiveKitClient.podspec
@@ -14,7 +14,7 @@ Pod::Spec.new do |spec|
 
   spec.source_files = "Sources/**/*"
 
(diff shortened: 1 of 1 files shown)


ladvoc:
I found this change causes `pod spec lint` to fail with the following error:
```
- ERROR | [tvOS] unknown: Encountered an unknown error (The platform of the target `App` (tvOS 17.0) is not compatible with `LiveKitWebRTC (125.6422.18)`, which does not support `tvOS`.
```
I think adding the TV and VisionOS versions to the Podspec in webrtc-xcframework would fix this issue.

hiroshihorie:
Thanks @ladvoc Yes we should add it there also 🙂

PR #589, opened 2025-02-14
CocoaPods Workflows

ladvoc:
- Defines two workflows, lint and push, to streamline CocoaPods releases:
  - *cocoapods-lint*: Runs `pod lib lint` on every PR push and push to the main branch.
  - *cocoapods-push*: Runs `pod spec lint` and then `pod trunk push` whenever a new release is published.
- In case of linting failure, both workflows upload the generated Xcode project for manual inspection.

Note: `COCOAPODS_TRUNK_TOKEN` must be set in the repository’s secrets before using the push workflow.


hiroshihorie:
Unfortunately we might need to replace Swift-Log if we want to continue to support Cocoapods :
```
- ERROR | [visionOS] unknown: Encountered an unknown error (The platform of the target `App` (visionOS 1.0) is not compatible with `Logging (1.4.0)`, which does not support `visionOS`.
```

https://github.com/apple/swift-log/pull/271

ladvoc:
The issue stems from the fact that Cocoapods is [no longer officially supported](https://github.com/apple/swift-log/issues/320#issuecomment-2360691644) for swift-log. As a result, the Podspec for swift-log hasn’t been updated since the release of VisionOS, and when we recently added VisionOS as a supported target in our Podspec (#587), this introduced an error. Short of dropping support for Cocoapods or replacing swift-log, a potential solution is to build swift-log as a binary XCFramework which Cocoapods can depend on.

LiveKitClient.podspec
```diff
@@ -8,22 +8,23 @@ Pod::Spec.new do |spec|
 
   spec.ios.deployment_target = "13.0"
   spec.osx.deployment_target = "10.15"
-  spec.tvos.deployment_target = "17.0"
-  spec.visionos.deployment_target = "1.0"
```
hiroshihorie, line 12: Did we have to remove tvOS and visionOS ?

hiroshihorie (approved):

PR #593, opened 2025-02-18
Data streams

ladvoc:
**Summary of changes**
- Data streams support: full support for text and byte data streams.
- Data channel buffering: outgoing data packets are now buffered and sent asynchronously.

**Differences from other SDKs**
- Error reporting: most data streams methods are throwing, allowing the user to handle errors.
  - Open streams are coordinated with `IncomingStreamManager` and `OutgoingStreamManager` to allow detection of errors (e.g. duplicate stream ID).
- File APIs: byte streams come with additional APIs for sending/receiving files on disk with MIME type and file extension inferencing.


ladvoc:
I am encountering a race condition in testing between two rooms (see [DataStreamsTest](https://github.com/livekit/client-sdk-swift/blob/0154792aa15f0b1b09a068fa423d3079e4b6deea/Tests/LiveKitTests/DataStream/DataStreamTests.swift#L37)) and am trying to determine whether this could also occur in a non-test environment. Normally, a data stream transmission follows this sequence:

```mermaid
---
config:
  theme: neo
---
sequenceDiagram
    autonumber
    actor tx as Remote Participant
    participant rx as Stream Manager
    participant h as Topic Handler

    tx->>+rx: Header packet
    rx->>h: Invoke handler
    h-->>rx: Await reader sequence
    activate rx
    activate h
    loop Receive chunks
        tx->>rx: Chunk packet
        rx->>h: Yield chunk
    end
    tx->>rx: Trailer packet
    rx->>h:  Terminate sequence
    deactivate h
    deactivate rx
```
However, an error can occur when the first chunk packet arrives before the stream has been opened by awaiting the stream reader. In this case, the first chunk packet is dropped as there is not yet anywhere to route it:
```mermaid
---
config:
  theme: neo
---
sequenceDiagram
    title First chunk arrives before await
    autonumber
    actor tx as Remote Participant
    participant rx as Stream Manager
    participant h as Topic Handler

    tx->>rx: Header packet
    rx->>h: Invoke handler
    tx->>rx: Chunk packet
    rx-->rx: Error: unknown stream <br> [chunk dropped]
    h-->>rx: Await chunks
    activate rx
    activate h
    tx->>rx: Trailer packet
    rx->>h: Error: incomplete
    deactivate h
    deactivate rx
```

lukasIO:
naive question, why do we have to _await_ the opening of the stream reader? 
I think ideally `registerTextStreamHandler` would be a sync operation?

hiroshihorie (commented):
Looks like some APIs are not available and tests are failing 👀

lukasIO:
for JS the `start` method of the ReadableStream is invoked immediately when the stream is constructed, see https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream/ReadableStream#start

We don't want to depend on _when_ the reader is being awaited as a signal to accept/discard the stream. Registering the handler on the room is showing a user's intent to process the streams already.

ladvoc:
> Registering the handler on the room is showing a user's intent to process the streams already.

That makes sense, thanks for the clarification. I’ll update the Swift implementation to match this behavior.

ladvoc:
The issue is now fixed. @lukasIO, I would appreciate your feedback on the new method I added to stream reader. This method allows the user to incrementally write a file to disk as it is received, returning its path once the entire file is received:
```swift
try await room.localParticipant
    .registerByteStreamHandler(for: "my-topic") { reader, participantIdentity in
        let fileURL = try await reader.readToFile()
        print("Wrote file to: \(fileURL)")
    }
```
By default, this method writes to the system temporary directory, and chooses the file name and extension by inspecting the name and MIME type fields from the stream header. Additional overloads also exist which allow the user to customize the destination directory and override the file name.

lukasIO:
that's a nice addition! 
`readToFile` sounds a bit confusing, `.writeToFile` or `.toFile` makes more sense in my head

hiroshihorie (approved):
Nice, tested it with the example app. ✅

PR #598, opened 2025-02-20
In-app audio capture format converter

hiroshihorie:
Runtime audio format converting instead of hard-coding the format.
ReplayKit uses different format from macOS screen share app audio.


ladvoc:
I just tested this with audio samples from ReplayKit; no issues.

ladvoc:
Actually, I also hear crackling sounds now. Does it sound like this for you too?

https://github.com/user-attachments/assets/2e60c3f0-934c-4619-872f-195ea7b9c33e

hiroshihorie:
@ladvoc Yes sounds pretty terrible, not sure what's going on. The convert part seems to be ok. Maybe something related to scheduling.

ladvoc:
@hiroshihorie, I checked the output status of `converter.convert(to:error:)`, and it is returning `inputRanDry`, which according to the docs indicates "the method doesn’t have enough input available to satisfy the request." Maybe buffering app audio samples would fix the issue?

hiroshihorie:
@ladvoc I think it always returns `.inputRanDry` when setting `outStatus.pointee = .noDataNow`, I think it should be ok. Test seems to be ok 🤔

hiroshihorie:
@ladvoc I think it's fixed now

ladvoc:
Yes, it works perfectly now!

hiroshihorie:
@ladvoc Nice! finally we can merge these PRs. 😁

PR #606, opened 2025-02-24
Tests: Audio recorder

hiroshihorie:


Tests/LiveKitTests/AudioEngineTests.swift
```diff
@@ -297,6 +304,40 @@ class AudioEngineTests: LKTestCase {
...
+        let format16k: [String: Any] = [
+            AVFormatIDKey: Int(kAudioFormatMPEG4AAC),
+            AVSampleRateKey: 16000,
+            AVNumberOfChannelsKey: 1,
+            AVEncoderAudioQualityKey: AVAudioQuality.high.rawValue,
+        ]
+
+        let tempLocalUrl = FileManager.default.temporaryDirectory.appendingPathComponent(UUID().uuidString).appendingPathExtension("aac")
+        let recorder = try AVAudioRecorder(url: tempLocalUrl, settings: format16k)
+        recorder.record()
```
ladvoc, line 329: One small suggestion—wrapping `recorder.record()` and `player.play()` in `XCTAssertTrue()` could help catch any failures in recording or playback since both return false if something goes wrong.

PR #618, opened 2025-03-06
Unified CI workflow

ladvoc:
This PR creates a unified CI workflow that performs building, testing, linting, and documentation generation. This brings several advantages:
- **Multi-platform docs**: previously, docs were built only for macOS—meaning any platform-specific symbols would not be included in the built documentation. Moving to a unified CI workflow allows generating symbol graphs across multiple platforms to be used in doc generation without redundant builds.
    - In the platform matrix, any platform that has `symbol-graph` set to true will trigger a release build with symbol graphs emitted—these will then be used to build docs.
- **Doc correctness**: warnings generated by DocC should be visible before PRs are merged to main.
- **Single run**: only one workflow run is triggered per push.


hiroshihorie (approved):

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,172 @@
...
+    branches: [main]
+  pull_request:
+    branches: [main]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+env:
+  TESTS: "FunctionTests SDKTests RoomTests PublishDataTests PublishBufferCapturerTests VideoViewTests TrackTests"
```
bcherry, line 15: is this really necessary? seems like an easy trap that we'll add new tests but forget to add them here and thus CI won't actually run them

ladvoc: This part was copied from the existing test workflow.

hiroshihorie: Yes Ben is right, but some tests don't run on CI/Simulator needs device etc. We can organize this in another PR.

bcherry (approved):

PR #620, opened 2025-03-07
Chunk text along UTF-8 boundaries

ladvoc:
This PR addresses a limitation in the current data streams implementation as noted by @lukasIO where text is not chunked along UTF-8 boundaries.


Sources/LiveKit/DataStream/StreamData.swift
```diff
@@ -0,0 +1,52 @@
...
+
+extension String: Chunkable {
+    func chunks(of size: Int) -> [SubSequence] {
+        guard size > 0, !isEmpty else { return [] }
+
+        var result: [SubSequence] = []
+        var currentIndex = startIndex
+
+        while currentIndex < endIndex {
+            let nextIndex = index(currentIndex, offsetBy: size, limitedBy: endIndex) ?? endIndex
```
lukasIO, line 46: this appears to chunk the string by character length instead of by byte size?

hiroshihorie (approved):
LGTM ✅

PR #622, opened 2025-03-10
Send pre-connect audio

pblazej:
- Adds infrastructure and public API for the agent pre-connect feature based on data channels

https://github.com/livekit-examples/multimodal-agent-python/pull/18

### Building blocks:
- `LocalAudioTrackRecorder` - a simple stream-based wrapper for `startLocalRecording`, can be used in isolation
- `PreConnectAudioBuffer` - handles the `Room` state automatically (if a `Room?` is passed) by setting participant attributes and flushing the recording into the data stream 🚰, can be used manually via `start/stop` methods
- `Room.startCapturingBeforeConnecting` - a convenience one-liner to get the default behavior; start automatically, stop when the agent joins

### Usage

```swift
.onAppear {
    try? AudioManager.shared.setRecordingAlwaysPreparedMode(true)
}

+

private func connect() {
    Task(priority: .userInitiated) {
        isConnecting = true

        if enablePreConnect {
            try? await room.startCapturingBeforeConnecting()
            isConnecting = false
        }
```


Sources/LiveKit/Track/Recorders/LocalAudioTrackRecorder.swift
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    public func start(maxSize: Int = 0) -> Stream {
+        let buffer: Stream.Continuation.BufferingPolicy = maxSize > 0 ? .bufferingNewest(maxSize) : .unbounded
+        let stream = Stream(bufferingPolicy: buffer) { continuation in
+            self.continuation = continuation
+        }
+
+        track.add(audioRenderer: self)
+        AudioManager.shared.startLocalRecording()
```
pblazej, line 44: @hiroshihorie any idea why it flags the non-existing `throws`? I tried older tool chains but no luck 🤔

hiroshihorie: This method can throw now, since it can fail when audio session configuration fails... should be marked as throws.

pblazej: Just needs rebase 💯

pblazej: TODO: `LocalAudioTrack.startCapture()` may be a better fit.

Sources/LiveKit/Track/Recorders/LocalAudioTrackRecorder.swift
pblazej, line 1: This could conform to `AsyncSequence` as well (as elsewhere), but IMO `start()->Stream` is more logical here.

Sources/LiveKit/Track/Recorders/LocalAudioTrackRecorder.swift
```diff
@@ -0,0 +1,138 @@
...
+
+    /// Starts capturing audio from the local track and returns a stream of audio data.
+    /// - Returns: A stream of audio data.
+    /// - Throws: An error if the audio track cannot be started.
+    public func start() async throws -> Stream {
+        guard _state.continuation == nil else {
+            throw LiveKitError(.invalidState, message: "Cannot start the recorder multiple times.")
+        }
+
+        try await track.startCapture()
```
pblazej, line 74: Forking that into a new task would give us some marginal speedup, for the sake of discarding errors (and converting it to `Stream?` as a did before).

pblazej:
TODO: `LiveKitKrispNoiseFilter Process failed, channel: 0`

Sources/LiveKit/Track/Recorders/LocalAudioTrackRecorder.swift
ladvoc: I think this makes sense, but I would recommend logging any errors thrown by `startCapture`.

ladvoc (approved):
Looks good to me!

Sources/LiveKit/Core/PreConnectAudioBuffer.swift
```diff
@@ -0,0 +1,147 @@
...
+    /// - Parameters:
+    ///   - room: The room instance to listen for events.
+    ///   - recorder: The audio recorder to use for capturing.
+    @objc
+    public init(room: Room?,
+                recorder: LocalAudioTrackRecorder = LocalAudioTrackRecorder(
+                    track: LocalAudioTrack.createTrack(),
+                    format: .pcmFormatInt16,
+                    sampleRate: 24000,
+                    maxSize: 10_000_000
```
bcherry, line 55: can you add a comment for where 10M comes from?

bcherry (approved):
this is really cool

Sources/LiveKit/Track/Recorders/LocalAudioTrackRecorder.swift
pblazej: Yes, I had e.g. stream readers in mind here.

Good point about the `consuming` nature, but you quickly run into:
image

Sources/LiveKit/Core/PreConnectAudioBuffer.swift
pblazej: yes, we can convert numbers like `private static let chunkSize = 15000` as well 👍

pblazej: ec6b27d66b95d21c23f2b086bb887bbf348fd51e

Sources/LiveKit/Track/Recorders/LocalAudioTrackRecorder.swift
pblazej: Plus ofc `struct` vs `class` vs ObjC

PR #624, opened 2025-03-11
Default AudioCaptureOptions for Simulator

hiroshihorie:
On simulator, Apple's Voice-Processing I/O is not available. Use WebRTC's voice processing instead.
CLT-683


ladvoc (approved):
LGTM ✅

PR #632, opened 2025-03-13
Audio mix recorder

hiroshihorie:
Easily mix and record multiple audio sources (AudioRenderer protocol).
Can add additional source while recording.

TODO:
- [x] Customize recording format
- [x] Optimize code

## Example usage
```swift
let audioSettings: [String: Any] = [
    AVFormatIDKey: kAudioFormatMPEG4AAC,
    AVSampleRateKey: 48_000,
    AVNumberOfChannelsKey: 1,
    AVLinearPCMBitDepthKey: 32,
    AVLinearPCMIsFloatKey: true,
    AVLinearPCMIsNonInterleaved: false,
    AVLinearPCMIsBigEndianKey: false,
]

let recordFilePath = URL(fileURLWithPath: NSTemporaryDirectory()).appendingPathComponent(UUID().uuidString + ".aac")
print("Recording to \(recordFilePath)...")

let recorder = try AudioMixRecorder(filePath: recordFilePath, audioSettings: audioSettings)
try recorder.start()

// Record local audio
AudioManager.shared.add(localAudioRenderer: recorder.addSource())
// And all remote audio
AudioManager.shared.add(remoteAudioRenderer: recorder.addSource())
// Or individual remote tracks
// RemoteAudioTrack.add(audioRenderer:)
```


bcherry (changes requested):
this is awesome and it works!

I found one bug, not sure what the cause is, but if I do (pseudo-code)
 
```
room.connect()
audioRecorder = audiomixrecorder(filePath:"../..")
audiomanager.shared.add(localRenderer: audioRecorder.newSource())
audiomanager.shared.add(remoteRendered: audioRecorder.newSource())
```

then at disconnect

```
room.disconnect()
audioRecorder.stop()
audioRecorder = nil
```

then if I try to connect _again_, i get a crash. I can fix the crash by manually removing the old sources from the audiomanager delegates. so it suggests that the sources (which should be stored weakly in audiomanager) aren't getting deallocated and maybe there is a retain cycle inside AudioMixRecorder

otherwise this is perfect

hiroshihorie:
@bcherry Can you try to reproduce the crash again ?

ladvoc (approved):

hiroshihorie:
@ladvoc I agree we can improve the AudioSettings but maybe we can add that later. I think it's a bit hard to properly cover all  settings.

Issue #528, opened 2024-12-10
WebSocket "SWIFT TASK CONTINUATION MISUSE" crash


dfed:
**Describe the bug**
I have a crash report from LiveKit's WebSocket cancelling a continuation twice. I've only seen this happen once.

**SDK Version**
2.0.17

**iOS/macOS Version**
iOS 18.1.1

**Xcode Version**
Xcode 16.1

**Steps to Reproduce**
Unknown

**Expected behavior**
Does not crash.

**Screenshots**

**Logs**
Crash log below. I've removed some threads for clarity.
```
OS Version: iOS 18.1.1 (22B91)
Report Version: 104

Exception Type: EXC_BREAKPOINT (SIGTRAP)
Exception Codes: 0 at 0x00000001824645e0
Crashed Thread: 7

Application Specific Information:
_Concurrency/CheckedContinuation.swift > Fatal error > SWIFT TASK CONTINUATION MISUSE: init(url:) tried to resume its continuation more than once, throwing Error Domain=io.livekit.swift-sdk Code=100 "Cancelled" UserInfo={NSLocalizedDescription=Cancelled}!

Thread 7 Crashed:
0   libswiftCore.dylib              0x3038555e0         _assertionFailure
1   libswift_Concurrency.dylib      0x31d9c5820         CheckedContinuation.resume
2   Tolan                           0x2036b57b0         WebSocket.close (WebSocket.swift:73)
3   Tolan                           0x2036b7390         [inlined] closure in WebSocket.init (WebSocket.swift:63)
4   Tolan                           0x2036b7390         closure in WebSocket.init
5   libswift_Concurrency.dylib      0x31da2d014         __swift::__runtime::llvm::function_ref<T>::callback_fn<T>
6   libswift_Concurrency.dylib      0x31da2bb50         withStatusRecordLock
7   libswift_Concurrency.dylib      0x31da2c700         swift_task_cancelImpl
8   Tolan                           0x203b0a948         LiveKitCall.disconnect (ChatService.swift:1558)
9   libswift_Concurrency.dylib      0x31da21100         swift::runJobInEstablishedExecutorContext

Thread 0
0   UnityFramework                  0x1052e5ac4         delete[]
1   UnityFramework                  0x10511bbc4         MultiplyMatrixArrayWithBase4x4_NEON
2   UnityFramework                  0x107a7bba0         DllCanUnloadNow
3   UnityFramework                  0x1062c08fc         DllCanUnloadNow
[...]

PR #633, opened 2025-03-14
Sync access to web socket continuations

pblazej:
Resolves https://github.com/livekit/client-sdk-swift/issues/528

Turning `WebSocket` into an `actor` would do more harm than good (nonisolated `deinit`, nonisolated requirements). Nullifying in the critical section is the only way to prevent double `resume()`.


hiroshihorie (approved):
Nice ✅

PR #637, opened 2025-03-19
Convert stream timestamps from/to ms

pblazej:
Noticed that while working with text streams 🕙


ladvoc (approved):
That was an oversight on my part, thank you for fixing!

Issue #636, opened 2025-03-18
Error while trying to unmute localParticipant in LiveKit


BerkanBlindid:
![Image](https://github.com/user-attachments/assets/769f9ce8-3a27-4528-822c-1261baff3a8f)
![Image](https://github.com/user-attachments/assets/8df5235d-1deb-4131-b6a3-0a4856617a1b)

I'm just trying to unmute the local participant and I get this error.

SDK Version: 2.3.1

iPhone 15 Pro version: 18.4

Xcode version: 16.1

jpmcglone:
This is exactly the same error I'm getting. I will try to downgrade 1 version at a time to see where it doesn't happen and let you know which version is stable.

jpmcglone:
@BerkanBlindid I had to downgrade all the way to 2.1.0 to not have the bug 👀

hiroshihorie:
This is odd, I can't reproduce this.

It seems that somehow the nodes aren't attached.
```
  engine.attach(appNode)
  engine.attach(appMixerNode)
```

Can you try adding a break point if attach logic gets called at all ?
![Image](https://github.com/user-attachments/assets/4ef66173-1b9c-44fb-a8ec-9e20fef09de0)

istevanovic:
I get the same crash on 2.3.1 version. @hiroshihorie 

![Image](https://github.com/user-attachments/assets/881dc2e2-bba0-4d39-901e-f88fb8fbca99)

I added the breakpoint as you suggested in the previous comment, it doesnt get called.
I tried with both custom and default AudioEngineObserver.

BerkanBlindid:
I added the break point you mentioned and there was no break when there was an error in version 2.3.1. Then I downgraded the version to 2.1.0 and the problem was solved. @jpmcglone Thanks for your help. I kept getting the problem in higher versions. It fixed it in version 2.1.0 for me too.

jpmcglone:
@BerkanBlindid Glad that helped!

When I created a LiveKit room and joined, everything worked fine with about three participants for a short time. At some point, I left the room and couldn’t rejoin without encountering a crash when unmuting my mic (very consistent). I'm not sure if that context helps whoever is working on this bug.

Also, I’ve noticed participant video issues in the latest LiveKit version that weren’t present in 2.1.0. This might be due to changes in how participant video views should be used, but I just wanted to flag that versions beyond 2.1.0 seem rather buggy.

BerkanBlindid:
[...]

PR #638, opened 2025-03-20
Ensure AudioManager is prepared

hiroshihorie:
The lazy initialization of `AudioManager` was causing inconsistencies in the `AudioEngineObserver` event lifecycle.
`AVAudioNode` instances were being connected without being attached, which could lead to crashes.
Ensure that the shared instance of `AudioManager` is ready whenever any Room object is instantiated.

Related: https://github.com/livekit/client-sdk-swift/issues/636


ladvoc (approved):
Looks good to me. I will test this on device

PR #641, opened 2025-03-21
Remove unnecessary logs from streams

pblazej:
In some cases these were the only form of error propagation (although not public), so we need to be careful with that 👍


ladvoc (approved):
LGTM ✅

PR #645, opened 2025-03-26
Swift 6: Delegates and global functions

pblazej:
- Adds `Sendable` requirement for all delegate protocols
  - it's an equivalent of "you can be called on any thread"
  - is it a breaking change? https://forums.swift.org/t/is-adding-a-sendable-requirement-a-breaking-change/69957/7
- Adds `Sendable` to `MulticastDelegate` et al.
- Converts simpler internal classes to actors (`Sendable` by default)
- Marks UI(Kit)-related stuff as `@MainActor`
- Fixes misc warnings and deprecations

Test: enable `.enableExperimentalFeature("StrictConcurrency")` for 5.9 package.

### DIFF
--- a/Sources/LiveKit/Core/RTC.swift
+++ b/Sources/LiveKit/Core/RTC.swift
@@ -49,7 +49,7 @@ private class VideoEncoderFactorySimulcast: LKRTCVideoEncoderFactorySimulcast {
     }
 }
 
-class RTC {
+actor RTC {
     static let h264BaselineLevel5CodecInfo: LKRTCVideoCodecInfo = {
         // this should never happen
         guard let profileLevelId = LKRTCH264ProfileLevelId(profile: .constrainedBaseline, level: .level5) else {
--- a/Sources/LiveKit/Core/Room.swift
+++ b/Sources/LiveKit/Core/Room.swift
@@ -218,7 +218,9 @@ public class Room: NSObject, ObservableObject, Loggable {
         // log sdk & os versions
         log("sdk: \(LiveKitSDK.version), os: \(String(describing: Utils.os()))(\(Utils.osVersionString())), modelId: \(String(describing: Utils.modelIdentifier() ?? "unknown"))")
 
-        signalClient._delegate.set(delegate: self)
+        Task {
+            await signalClient._delegate.set(delegate: self)
+        }
 
         log()
 
--- a/Sources/LiveKit/Core/Room.swift
+++ b/Sources/LiveKit/Core/Room.swift
@@ -218,9 +218,7 @@ public class Room: NSObject, ObservableObject, Loggable {
         // log sdk & os versions
         log("sdk: \(LiveKitSDK.version), os: \(String(describing: Utils.os()))(\(Utils.osVersionString())), modelId: \(String(describing: Utils.modelIdentifier() ?? "unknown"))")
 
-        Task {
-            await signalClient._delegate.set(delegate: self)
-        }
+        signalClient._delegate.set(delegate: self)
 
         log()
 
--- a/Sources/LiveKit/Support/AppStateListener.swift
+++ b/Sources/LiveKit/Support/AppStateListener.swift
@@ -35,8 +35,8 @@ protocol AppStateDelegate: AnyObject, Sendable {
     func appDidWake()
 }
 
-@globalActor
-actor AppStateListener: Loggable {
+@MainActor
+class AppStateListener: Loggable {
     static let shared = AppStateListener()
 
     private let _queue = OperationQueue()
--- a/Sources/LiveKit/Support/AppStateListener.swift
+++ b/Sources/LiveKit/Support/AppStateListener.swift
@@ -25,7 +25,7 @@ import AppKit
 #endif
 
 @objc
-protocol AppStateDelegate: AnyObject {
+protocol AppStateDelegate: AnyObject, Sendable {
     func appDidEnterBackground()
     func appWillEnterForeground()
     func appWillTerminate()
--- a/Sources/LiveKit/Support/AppStateListener.swift
+++ b/Sources/LiveKit/Support/AppStateListener.swift
@@ -35,7 +35,8 @@ protocol AppStateDelegate: AnyObject, Sendable {
     func appDidWake()
 }
 
-class AppStateListener: Loggable {
+@globalActor
+actor AppStateListener: Loggable {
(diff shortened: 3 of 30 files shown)


Sources/LiveKit/Core/RTC.swift
```diff
@@ -49,7 +49,7 @@ private class VideoEncoderFactorySimulcast: LKRTCVideoEncoderFactorySimulcast {
     }
 }
 
-class RTC {
+actor RTC {
```
pblazej, line 52: This is mostly for static `LKRTCVideoEncoderFactory` etc. inside this object. IMO it's internal and does not require further changes, so good to convert 👍

Sources/LiveKit/Support/AsyncSerialDelegate.swift
```diff
@@ -16,26 +16,22 @@
...
 import Foundation
 
-class AsyncSerialDelegate<T> {
-    private struct State {
-        weak var delegate: AnyObject?
-    }
-
-    private let _state = StateSync(State())
-    private let _serialRunner = SerialRunnerActor<Void>()
+actor AsyncSerialDelegate<T: Sendable> {
```
pblazej, line 19: It's also internal and easy enough to remove `StateSync`.

Sources/LiveKit/Support/MulticastDelegate.swift
```diff
@@ -28,7 +28,7 @@ public protocol MulticastDelegateProtocol {
 ///
 /// Uses `NSHashTable` internally to maintain a set of weak delegates.
 ///
-public class MulticastDelegate<T>: NSObject, Loggable {
+public class MulticastDelegate<T: Sendable>: NSObject, @unchecked Sendable, Loggable {
```
pblazej, line 31: This one cannot be easily actorized:
- it's public
- there's some inheritance going on - needs deeper refactor

Sources/LiveKit/Support/AppStateListener.swift
```diff
@@ -35,7 +35,8 @@ protocol AppStateDelegate: AnyObject {
     func appDidWake()
 }
 
-class AppStateListener: Loggable {
+@globalActor
+actor AppStateListener: Loggable {
```
pblazej, line 39: This could be a `final class ... : Sendable` as well as every prop is sendable, that was just a preference.

pblazej:
I introduced some warnings to unit tests, but I'm going to tackle them in a separate ticket.

Sources/LiveKit/Core/Room.swift
```diff
@@ -218,7 +218,9 @@ public class Room: NSObject, ObservableObject, Loggable {
         // log sdk & os versions
         log("sdk: \(LiveKitSDK.version), os: \(String(describing: Utils.os()))(\(Utils.osVersionString())), modelId: \(String(describing: Utils.modelIdentifier() ?? "unknown"))")
 
-        signalClient._delegate.set(delegate: self)
+        Task {
+            await signalClient._delegate.set(delegate: self)
+        }
```
hiroshihorie, line 223: I wonder if this is guaranteed to be early enough if we call connect() right after init().

pblazej: Yeah, this is kinda painful - to be on the safe side, we can keep it as `final class: Sendable` with state sync.

pblazej: 246ae901945b8a480b81a6b78c5eb73bbe8b6241

ladvoc (commented):
From the Swift [docs](https://developer.apple.com/documentation/swift/sendable):
> To satisfy the requirements of the Sendable protocol, a class must:
> - Be marked final
> - Contain only stored properties that are immutable and sendable
> - Have no superclass or have NSObject as the superclass

Yeah, it looks like in order to make `Participant` Sendable without `@unchecked`, there would need to be some breaking changes—short of introducing "a ton of proxying." Most notably, Objective-C support would need to be dropped (as `@objc` requires inheritance from `NSObject`) and there could be no inheritance. 

I think for now, the best option is to used `@unchecked`. Another option is to use the `@preconcurrency` attribute on specific types. This seems to be the approach Apple takes for some public APIs, for example:
```swift
@preconcurrency public struct EnvironmentObject<ObjectType>
```

Side note: later on when we move to safe Sendable conformance across the SDK, `Participant` could be modeled as an enum as it is in Rust to avoid inheritance:
```rust
pub enum Participant {
    Local(LocalParticipant),
    Remote(RemoteParticipant),
}
```

pblazej:
> From the Swift [docs](https://developer.apple.com/documentation/swift/sendable):
> 
> > To satisfy the requirements of the Sendable protocol, a class must:
> > 
> > * Be marked final
> > * Contain only stored properties that are immutable and sendable
> > * Have no superclass or have NSObject as the superclass
> 
> Yeah, it looks like in order to make `Participant` Sendable without `@unchecked`, there would need to be some breaking changes—short of introducing "a ton of proxying." Most notably, Objective-C support would need to be dropped (as `@objc` requires inheritance from `NSObject`) and there could be no inheritance.
> 
> I think for now, the best option is to used `@unchecked`. Another option is to use the `@preconcurrency` attribute on specific types. This seems to be the approach Apple takes for some public APIs, for example:
> 
> ```swift
> @preconcurrency public struct EnvironmentObject<ObjectType>
> ```
> 
> Side note: later on when we move to safe Sendable conformance across the SDK, `Participant` could be modeled as an enum as it is in Rust to avoid inheritance:
> 
> ```rust
> pub enum Participant {
>     Local(LocalParticipant),
>     Remote(RemoteParticipant),
> }
> ```

I agree with `struct`/`enum` (aka embrace immutability), but keep in mind Obj-C/RN compatibility 😭

Sources/LiveKit/Track/Capturers/VideoCapturer.swift
```diff
@@ -333,7 +333,7 @@ extension VideoCapturer {
             self.delegate?.capturer(capturer, didCapture: rtcFrame)
 
             if self.rendererDelegates.isDelegatesNotEmpty {
-                self.rendererDelegates.notify { renderer in
+                self.rendererDelegates.notify { [lkFrame] renderer in
```
pblazej, line 336: @hiroshihorie this one bothers me as well (for older Swift version, there's a false positive warning here)

hiroshihorie (approved):

PR #646, opened 2025-03-27
Swift 6: Participants and streams

pblazej:
- Marks `Participant` et al. as `@unchecked Sendable` (per last discussions)
- Fixes `Sendable` warnings for stream-related methods/closures

### DIFF
--- a/Sources/LiveKit/Participant/Participant.swift
+++ b/Sources/LiveKit/Participant/Participant.swift
@@ -23,7 +23,7 @@ internal import LiveKitWebRTC
 #endif
 
 @objc
-public class Participant: NSObject, ObservableObject, Loggable {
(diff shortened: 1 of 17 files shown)


Sources/LiveKit/Participant/Participant.swift
```diff
@@ -179,7 +179,7 @@ public class Participant: NSObject, ObservableObject, Loggable {
             }
 
             // Notify when state mutates
-            Task.detached { @MainActor in
+            Task { @MainActor in
```
pblazej, line 182: I created another ticket to revisit that `detached` thing, in most cases (if not all) it's just bypassing current task priority and actor context.

pblazej: In some cases like `VideoView` it seems unnecessary (at least for now?)

hiroshihorie (approved):

PR #650, opened 2025-03-27
Hiroshi/set audiomodule type

hiroshihorie:


Sources/LiveKit/Audio/Manager/AudioManager.swift
```diff
@@ -163,21 +163,49 @@ public class AudioManager: Loggable {
     public let defaultInputDevice = AudioDevice(ioDevice: LKRTCIODevice.defaultDevice(with: .input))
 
     public var outputDevices: [AudioDevice] {
+        #if os(macOS)
         RTC.audioDeviceModule.outputDevices.map { AudioDevice(ioDevice: $0) }
+        #else
+        []
```
pblazej, line 169: Out of curiosity, this is a no-op because of RTC limitations for iOS?

Sources/LiveKit/Core/RTC.swift
```diff
@@ -50,6 +50,14 @@ private class VideoEncoderFactorySimulcast: LKRTCVideoEncoderFactorySimulcast {
 }
 
 actor RTC {
+    struct PeerConnectionFactoryState {
+        var isInitialized: Bool = false
+        var admType: AudioDeviceModuleType = .audioEngine
+        var bypassVoiceProcessing: Bool = false
+    }
+
+    static let pcFactoryState = StateSync(PeerConnectionFactoryState())
```
pblazej, line 59: Nit: probably you can get rid of `StateSync` vs `actor` here if the mutations are local.

pblazej (approved):
LGTM

Sources/LiveKit/Core/RTC.swift
hiroshihorie: Will it require to be async ?

Sources/LiveKit/Audio/Manager/AudioManager.swift
hiroshihorie: Yes this is a limitation at the moment.
I think we can simulate this to manipulate the AVAudioSession output port, but I'm not sure.

Sources/LiveKit/Audio/Manager/AudioManager+ModuleType.swift
```diff
@@ -0,0 +1,51 @@
...
+        }
+    }
+}
+
+public extension AudioManager {
+    /// Sets the desired `AudioDeviceModuleType` to be used which handles all audio input / output.
+    ///
+    /// This method must be called before the peer connection is initialized. Changing the module type after
+    /// initialization is not supported and will result in an error.
+    static func set(audioDeviceModuleType: AudioDeviceModuleType) throws {
```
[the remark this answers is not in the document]
hiroshihorie: Yes it only works when called early.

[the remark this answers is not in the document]
pblazej: I mean e.g. in our example app even if you put it before `.shared` e.g. here:
```swift
        // here
        AudioManager.shared.onDeviceUpdate = { [weak self] _ in
```
it's not enough as SwiftUI may create `.shared` for you (e.g. in form of computed props).

Sources/LiveKit/Audio/Manager/AudioManager.swift
pblazej: We can comment it - just for the future generations.

Sources/LiveKit/Core/RTC.swift
pblazej: Yes

Sources/LiveKit/Audio/Manager/AudioManager+ModuleType.swift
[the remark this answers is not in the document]
hiroshihorie: Yes it needs to be earlier at the moment. For our example app, it works here.
スクリーンショット 2025-04-03 14 36 10

PR #651, opened 2025-03-28
VideoView

pblazej:
- Fixes all warnings around `VideoView` + related classes
- It does NOT change the isolation of `VideoRenderer` et al. which is **clearly non-main-thread**
  - the switch still happens in `VideoView.render(frame:)`
  - however, it removes the use of `Task.detached` - IMO keeping the task priority as described [here](https://www.donnywals.com/wp-content/uploads/Screen-Shot-2021-06-08-at-18.33.50-1024x449.png) is not harmful, otherwise we should define it explicitly
- The alternative would be the opposite:
  - break the isolation of `LKRTCVideoRenderer` with `@preconcurrency`
  - then move most of this stuff to `@MainActor` - with some performance implications
  - that wasn't the intention of RTC implementers to have only sync rendering ⌚ 

⚠️ This one needs deeper testing with example apps, I did macOS + iOS with tsan/Metal warnings enabled, no assertions triggered so far 🦺 

Take your time 🧪 and feel free to break it

### DIFF
--- a/Sources/LiveKit/Views/VideoView.swift
+++ b/Sources/LiveKit/Views/VideoView.swift
@@ -34,7 +34,7 @@ public class VideoView: NativeView, Loggable {
 
     // MARK: - Static
 
-    static let mirrorTransform = CATransform3DMakeScale(-1.0, 1.0, 1.0)
+    static let mirrorTransform = CATransform3D.mirror
     private static let _freezeDetectThreshold = 2.0
 
     /// Specifies how to render the video withing the ``VideoView``'s bounds.
--- a/Sources/LiveKit/Views/VideoView.swift
+++ b/Sources/LiveKit/Views/VideoView.swift
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import AVFoundation
+@preconcurrency import AVFoundation
 import MetalKit
 
 #if swift(>=5.9)
@@ -96,7 +96,7 @@ public class VideoView: NativeView, Loggable {
 
     /// Calls addRenderer and/or removeRenderer internally for convenience.
     @objc
-    public weak var track: VideoTrack? {
+    public nonisolated weak var track: VideoTrack? {
         get { _state.track as? VideoTrack }
         set {
             _state.mutate {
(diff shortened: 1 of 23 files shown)


Sources/LiveKit/Views/VideoView.swift
```diff
@@ -124,9 +124,7 @@ public class VideoView: NativeView, Loggable {
         get { _state.isHidden }
         set {
             _state.mutate { $0.isHidden = newValue }
-            Task.detached { @MainActor in
```
pblazej, line 127: I don't think the `Task` is even needed for main actor isolated prop, probably a leftover.

Sources/LiveKit/Views/VideoView.swift
```diff
@@ -584,22 +582,22 @@ private extension VideoView {
...
     }
 
-    public var adaptiveStreamSize: CGSize {
+    public nonisolated var adaptiveStreamSize: CGSize {
         _state.rendererSize ?? .zero
     }
 
-    public func set(size: CGSize) {
+    public nonisolated func set(size: CGSize) {
         DispatchQueue.main.async { [weak self] in
```
pblazej, line 594: We can enforce the use of e.g. `MainActor.run` via lint, it's mostly a matter of consistency now.

Sources/LiveKit/Views/VideoView.swift
```diff
@@ -227,12 +225,20 @@ public class VideoView: NativeView, Loggable {
         }
     }
 
-    var _state: StateSync<State>
+    let _state: StateSync<State>
 
     // MARK: - Private
 
+    #if compiler(>=6.0)
+    private nonisolated(unsafe) var _primaryRenderer: NativeRendererView?
```
pblazej, line 233: There's a slight inconsistency here, it worked fine until the `render` method wasn't explicitly marked as `nonisolated`.

pblazej: Hm, looking at 5.7 version, there's a little contradiction here: we cannot express the idea that renderers are mutated only from the main thread but accessed from nonisolated render methods.

In older versions, everything will be main actor by default and the compiler is still happy - changing the actual behavior.

hiroshihorie (approved):
LGTM ✅

PR #656, opened 2025-04-02
Minor visionOS patch

hiroshihorie:
@pblazej I had to add `final` for it to compile with visionOS.
Also `Sendable` is already defined in `CaptureOptions` but should we still add it for child classes ?


pblazej:
We can also fix the remaining one from old Xcode:
```
capture of 'sampleBuffer' with non-sendable type 'CMSampleBuffer' in a `@Sendable` closure
```

Sources/LiveKit/Types/Options/ARCameraCaptureOptions.swift
```diff
@@ -24,7 +24,7 @@ internal import LiveKitWebRTC
 #endif
 
 @objc
-public class ARCameraCaptureOptions: NSObject, VideoCaptureOptions {
+public final class ARCameraCaptureOptions: NSObject, VideoCaptureOptions, Sendable {
```
pblazej, line 27: Yep, this is just a protocol `protocol VideoCaptureOptions: CaptureOptions, Sendable`, we need to mark conforming classes (as it was done for `CameraCaptureOptions` et al.)

pblazej:
@hiroshihorie any blockers here?

hiroshihorie:
> @hiroshihorie any blockers here?

@pblazej  No blockers, I'll merge it.

PR #657, opened 2025-04-03
Create platform-dependent locks

pblazej:
Just a PoC:
- How can we use more recent APIs (without dynamic dispatch)
- How can we fix this kind of [warning](https://developer.apple.com/documentation/os/osallocatedunfairlock#overview)

`Mutex<>` still in progress, as non-copyable requirement breaks the build.


pblazej:
I think the idea is generally good, maybe would be nice to run another benchmark on our new (and concrete) implementation here.

pblazej:
There's no clear performance winner here (iOS ⬅️ macOS ➡️), but I think it's still beneficial to fix the warning and consume future improvements in the newer APIs.

Screenshot 2025-04-04 at 3 33 39 PM

hiroshihorie:
Nice. Even though there isn’t a significant gain at the moment, it’s good to transition to the safer locks.

pblazej:
> Nice. Even though there isn’t a significant gain at the moment, it’s good to transition to the safer locks.

👍 Haven't noticed any perf degradation with our new measurements 📐

PR #658, opened 2025-04-03
chore(docs): relocate docc to /reference/client-sdk-swift base path

rektdeckard:
See: https://github.com/livekit/web/pull/1054


ladvoc (approved):
LGTM ✅

PR #659, opened 2025-04-04
Main thread isolation in VideoView

pblazej:
This is a tiny change, but prevents a crash in `.v6` mode by fixing an important inconsistency.

- `OnDidMutate` is called within the critical section - so from _any_ thread
- for `Sendable` classes, it doesn't make much difference though
- for `@MainActor` methods like `mainSyncOrAsync` it does, as they:
  - were still `@MainActor` isolated
  - were called from bg threads - which is checked at runtime [now](https://forums.swift.org/t/crash-when-running-in-swift-6-language-mode/72431/2)

So the idea is `mainSyncOrAsync` signature now reflects what *really* happens here:
- it must be `nonisolated` itself as it's called from a `@Sendable` `OnDidMutate` 💚 
- the body is compile-time `@MainActor` 💚 
- the "switch" happens inside the method, not "accidentally outside"

The crash will happen during video track publication, the stack trace is pretty clear:

Screenshot 2025-04-04 at 11 18 53 AM

### DIFF
--- a/Sources/LiveKit/Views/VideoView.swift
+++ b/Sources/LiveKit/Views/VideoView.swift
@@ -862,6 +862,7 @@ extension LKRTCMTLVideoView: Mirrorable {
 #endif
 
 private extension VideoView {
+    #if compiler(>=5.9)
     nonisolated func mainSyncOrAsync(operation: @MainActor @escaping () -> Void) {
         if Thread.current.isMainThread {
             MainActor.assumeIsolated(operation)
@@ -871,6 +872,17 @@ private extension VideoView {
             }
         }
     }
+    #else
+    nonisolated func mainSyncOrAsync(operation: @escaping () -> Void) {
+        if Thread.current.isMainThread {
+            operation()
+        } else {
+            Task { @MainActor in
+                operation()
+            }
+        }
+    }
+    #endif
 }
 
 #if os(iOS)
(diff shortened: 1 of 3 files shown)


Sources/LiveKit/Views/VideoView.swift
```diff
@@ -285,7 +285,7 @@ public class VideoView: NativeView, Loggable {
 
             // Enter .main only if UI updates are required
             if trackDidUpdate || shouldRenderDidUpdate || renderModeDidUpdate {
-                self.mainSyncOrAsync {
+                self.mainSyncOrAsync { @MainActor in
```
pblazej, line 288: This `MainActor` is necessary only for <5.9 due to lack of `assumeIsolate` as you see below 🫳

hiroshihorie (approved):
I couldn't repo the crash with main, how can I repo the crash ?

pblazej:
> I couldn't repo the crash with main, how can I repo the crash ?

I used Xcode 16.2 with Package 6.0 - every video track publish will fail 💣

pblazej:
@hiroshihorie still unable to pull that one `spec.dependency("LiveKitWebRTC", "= 125.6422.28")`

hiroshihorie:
> @hiroshihorie still unable to pull that one `spec.dependency("LiveKitWebRTC", "= 125.6422.28")`

I haven't pushed that pod lib yet

PR #661, opened 2025-04-07
Measure state sync congestion

pblazej:
- Adds `os.signpost` capabilities
- Adds `LK_SIGNPOSTS` flag that you need to `.define` in your SPM vs redirecting logs to `/dev/null` (`OSLog.disabled`) - to avoid unexpected behavior/perf penalty
- The only awkward thing is `subscript()` as we cannot alter the signature here
- Dumps the results per file/area like so (macOS):

image

### DIFF
--- a/Sources/LiveKit/Support/StateSync.swift
+++ b/Sources/LiveKit/Support/StateSync.swift
@@ -14,11 +14,23 @@
  * limitations under the License.
  */
 
-import Combine
 import Foundation
 
+#if LK_SIGNPOSTS
+import os.signpost
+#endif
+
 @dynamicMemberLookup
-public final class StateSync<State>: @unchecked Sendable {
+public final class StateSync<State>: @unchecked Sendable, Loggable {
+    // MARK: - Logging
+
+    #if LK_SIGNPOSTS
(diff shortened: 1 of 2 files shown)


pblazej:
On iOS, it's "slower", shows some more anomalies, like:

image

pblazej:
Eventual "fixes" should come in separate PRs 😢

Sources/LiveKit/Support/StateSync.swift
```diff
@@ -14,11 +14,23 @@
...
+#if LK_SIGNPOSTS
+import os.signpost
+#endif
+
 @dynamicMemberLookup
-public final class StateSync<State>: @unchecked Sendable {
+public final class StateSync<State>: @unchecked Sendable, Loggable {
+    // MARK: - Logging
+
+    #if LK_SIGNPOSTS
```
pblazej, line 27: If that's too awkward to read, we can use the `.disabled` pattern as suggested here https://developer.apple.com/videos/play/wwdc2018/405/ + launch args.

hiroshihorie (approved):

PR #665, opened 2025-04-08
Improve negotiation

hiroshihorie:
This one needs to be tested well.

### DIFF
--- a/Sources/LiveKit/Support/AsyncDebounce.swift
+++ b/Sources/LiveKit/Support/AsyncDebounce.swift
@@ -16,10 +16,17 @@
 
 import Foundation
 
+/// Actor that implements debouncing of async operations
+///
+/// Debouncing ensures that rapid, repeated calls to an operation will only execute
+/// after a specified delay has elapsed since the most recent call.
 actor Debounce {
     private var _task: Task<Void, Never>?
     private let _delay: TimeInterval
+    private var _nonce = 0 // Used to track the latest scheduled task
 
+    /// Initialize with a specified delay
+    /// - Parameter delay: The time to wait after the last call before executing
     init(delay: TimeInterval) {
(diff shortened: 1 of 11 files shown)


Sources/LiveKit/Support/TaskQueue.swift
pblazej, line 1: Similarities to `QueueActor`?

Sources/LiveKit/Support/AsyncDebounce.swift
```diff
@@ -16,10 +16,17 @@
 
 import Foundation
 
+/// Actor that implements debouncing of async operations
+///
+/// Debouncing ensures that rapid, repeated calls to an operation will only execute
+/// after a specified delay has elapsed since the most recent call.
 actor Debounce {
```
pblazej, line 23: Visiting these actors recently, the internals get pretty complex quickly, have you considered simply using `Combine`/`AsyncAlgorithms` for that?

Sources/LiveKit/Support/TaskQueue.swift
hiroshihorie: Yes... I should merge / use one of them..

Sources/LiveKit/Support/AsyncDebounce.swift
hiroshihorie: I agree... are you talking about the [swift-async-algorithms](https://github.com/apple/swift-async-algorithms) package ?

pblazej: It would require more upstream changes, but using event-driven approach (streams as signals) can simplify the reasoning (especially long-term).

hiroshihorie:
Closing this one, I think I better make more smaller incremental changes not to break it.

PR #672, opened 2025-04-14
Update protobuf to v1.37.0

hiroshihorie:
Bump protobuf to v1.37.0


Package@swift-5.9.swift
pblazej, line 1: 5.7 package not supported?

pblazej (approved):

PR #682, opened 2025-04-23
Background blur video processor

pblazej:
### Overview

This is a revisited and slightly optimized version of Vision background blur.

While the basics are fairly easy to implement, the devil lies in the details:
- segmenting each frame synchronously is not an option, even on macOS
  - being unable to cancel ongoing requests, it's a simple recipe for getting the whole pipeline stuck and causing more frame drops
  - the same issue is visible in Apple's own [example](https://developer.apple.com/documentation/vision/applying-matte-effects-to-people-in-images-and-video)
  - there are 2 (?) basic approaches here:
    - handling the mask asynchronously so it's "slightly outdated"
    - delaying the whole pipeline - I'm afraid in the worst case there will be no easy way to recover, so I prioritized the main "sync" stream
- using `CI` Metal-backed stuff really helps here
- trying to reuse as many building blocks as possible

Example PR to test: https://github.com/livekit-examples/swift-example/pull/72

#### CoreImage 🟢
- requires yuv → rgba conversion which will increase general throughput (cloud as well)
- down-scaling + limiting the blur size should get us to 4K (on most devices at least)
- looks naive, but the key is Metal-backed `CIContext` + the choice of params

#### Metal (added for comparison) 🟠
- can work with yuv (or anything) directly
- but slow for large kernel sizes (anything above `8` will drop frames, especially on iOS)
- "proper" solution would require:
  - down-scaling (essentially level 1 mipmap)
  - smarter Gaussian blur decoupled into H and V pass
  - thread group optimization
    - size
    - caching
  - half-precision maybe?
  - etc. and there's no guarantee we can get faster than CI/MPS optimized impl

For now, my estimate is that the proper shader will be a significant investment, without any guarantee it's really optimal.

### DIFF
--- a/Sources/LiveKit/VideoProcessors/BackgroundBlurVideoProcessor.swift
+++ b/Sources/LiveKit/VideoProcessors/BackgroundBlurVideoProcessor.swift
@@ -24,9 +24,10 @@ import os.signpost
 
 /// A ``VideoProcessor`` that blurs the background of a video stream.
 ///
-/// This processor uses Vision to generate a mask of the person in the video stream,
-/// downscales the background by 4x, applies a blur, and then blends it back with the foreground.
+/// This processor uses Vision framework to generate a mask of the person in the video stream,
+/// downscales the background, applies a blur, and then blends it back with the foreground.
 ///
+/// - Important: This class is not thread safe and will be called on a dedicated serial `processingQueue`.
 @available(iOS 15.0, macOS 12.0, tvOS 15.0, visionOS 1.0, *)
 @objc
 public final class BackgroundBlurVideoProcessor: NSObject, @unchecked Sendable, VideoProcessor, Loggable {
@@ -150,7 +151,6 @@ public final class BackgroundBlurVideoProcessor: NSObject, @unchecked Sendable, 
             let scaleY = inputDimensions.height / maskDimensions.height
             let scaleTransform = CGAffineTransform(scaleX: scaleX, y: scaleY)
 
-            // Invert the mask so that the person is not blurred, just the background
             self.cachedMaskImage = maskImage.transformed(by: scaleTransform)
         }
     }


pblazej:
`Frame processing hasn't completed yet, skipping frame` happens on iOS @ 4K, will focus on that now. One of the options is to downscale the mask, however the segmentation time does not scale linearly.

Sources/LiveKit/Metal/BackgroundBlur.metal
pblazej, line 1: This is just the basic gist, without mipmaps - tried them but looks like the generation itself is slower than simple CI downscale...

pblazej:
To the testers: look for `Frame processing hasn't completed yet, skipping frame...`

Sources/LiveKit/Extensions/CoreImage.swift
```diff
@@ -0,0 +1,68 @@
...
+            .workingColorSpace: NSNull(),
+            .useSoftwareRenderer: false,
+            .priorityRequestLow: false,
+            .allowLowPower: true,
+        ])
+    }
+}
+
+public extension CIImage {
+    func croppedAndScaled(to rect: CGRect, highQuality: Bool = true) -> CIImage {
```
pblazej, line 34: Unfortunately, we need to recreate the crop information as `CVPixelBuffer` is still the source one...

Sources/LiveKit/VideoProcessors/BackgroundBlurVideoProcessor.swift
```diff
@@ -0,0 +1,178 @@
...
+    private let segmentationFrameInterval = 1
+    #else
+    private let segmentationFrameInterval = 3
+    #endif
+
+    // MARK: Vision
+
+    private let segmentationRequest = {
+        let segmentationRequest = VNGeneratePersonSegmentationRequest()
+        segmentationRequest.qualityLevel = .balanced
```
lukasIO, line 55: maybe we'd want to expose the segmentation quality level as a user setting?

Sources/LiveKit/VideoProcessors/BackgroundBlurVideoProcessor.swift
```diff
@@ -0,0 +1,178 @@
...
+        // Blur
+
+        let downscaleTransform = getDownscaleTransform(relativeTo: inputDimensions)
+        let downscaledImage = inputImage.transformed(by: downscaleTransform, highQualityDownsample: false)
+
+        blurFilter.inputImage = downscaledImage.clampedToExtent()
+        blurFilter.radius = blurRadius
+
+        guard let blurredImage = blurFilter.outputImage else { return frame }
+        let upscaledBlurredImage = blurredImage.transformed(by: downscaleTransform.inverted(), highQualityDownsample: false)
```
lukasIO, line 106: do we even need to manually upscale it? would have hoped the texture simply get's mapped onto the blendFilter in the correct dimension

Sources/LiveKit/VideoProcessors/BackgroundBlurVideoProcessor.swift
pblazej: `.accurate` won't work in most scenarios (including macOS), but I'm happy to do a `highQuality: Bool` between `.fast` and `.balanced`

Sources/LiveKit/VideoProcessors/BackgroundBlurVideoProcessor.swift
```diff
@@ -0,0 +1,178 @@
...
+@objc
+public final class BackgroundBlurVideoProcessor: NSObject, @unchecked Sendable, VideoProcessor, Loggable {
+    #if LK_SIGNPOSTS
+    private let signpostLog = OSLog(subsystem: Bundle.main.bundleIdentifier ?? "", category: "BackgroundBlur")
+    #endif
+
+    // MARK: Parameters
+
+    private let downscaleFactor: CGFloat = 2 // Downscale before blurring, upscale before blending
+    private let blurRadius: Float = 3 // Keep the kernel size small O(n^2)
```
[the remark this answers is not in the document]
lukasIO: I like the idea, just want to point out that in my testing on JS overly downscaling the background led to noticeably worse blurring, it didn't have the same visual "quality" anymore. Maybe CI is smarter about it though and this is not an issue here?

Sources/LiveKit/VideoProcessors/BackgroundBlurVideoProcessor.swift
pblazej: Screenshot 2025-04-25 at 1 28 29 PM

lukasIO: :D thanks for that!

Sources/LiveKit/VideoProcessors/BackgroundBlurVideoProcessor.swift
lukasIO: sounds good to me! yeah, accurate doesn't sound like it should be used for video in the first place

pblazej:
@lukasIO looks like it balances itself nicely (the cost of scaling vs total):

This is iPhone 14, Release:
image

Now we're totally safe even >30 FPS.

Sources/LiveKit/VideoProcessors/BackgroundBlurVideoProcessor.swift
```diff
@@ -0,0 +1,178 @@
...
+    private let relativeSize: CGFloat = 1080 // Blur effect optimized for HD, extrapolate for other video sizes
+
+    // Skip segmentation every N frames for slower devices
+    private var frameCount = 0
+    #if os(macOS)
+    private let segmentationFrameInterval = 1
```
pblazej, line 46: This could be calculated using `ProcessInfo.processInfo.isLowPowerModeEnabled`

pblazej:
CI errors are because of old cocoapods thing, will resolve that after finalizing comments 👍

hiroshihorie (approved):
✅ Nice
Apple's built-in bg blur also has segmentation delay so I think it's fine.

Sources/LiveKit/Track/Capturers/VideoCapturer.swift
```diff
@@ -51,7 +51,7 @@ public class VideoCapturer: NSObject, @unchecked Sendable, Loggable, VideoCaptur
     public let delegates = MulticastDelegate<VideoCapturerDelegate>(label: "VideoCapturerDelegate")
     public let rendererDelegates = MulticastDelegate<VideoRenderer>(label: "VideoCapturerRendererDelegate")
 
-    private let processingQueue = DispatchQueue(label: "io.livekit.videocapturer.processing")
+    private let processingQueue = DispatchQueue(label: "io.livekit.videocapturer.processing", autoreleaseFrequency: .workItem)
```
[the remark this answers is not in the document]
pblazej: Just learned that from Apple's ancient examples.

pblazej:
I'm merging despite:
- CocoaPods error - missing package `LiveKitWebRTC (= 125.6422.29)`
- random Room test flakiness (unrelated)

PR #686, opened 2025-04-30
Add Sendable for internal callbacks

pblazej:
So that consumers like:

```swift
AudioManager.shared.onDeviceUpdate = { [weak self] _ in
            guard let self else { return }
            // force UI update for outputDevice / inputDevice
            Task { @MainActor [weak self] in
```

have proper isolation.


hiroshihorie (approved):
lgtm

PR #687, opened 2025-05-01
Neater logging

hiroshihorie:
Improve logging to make it more consistent and descriptive.


pblazej:
Can we add a custom SwiftLint rule to force people to use `metaData: ScopedMetadataContainer = ScopedMetadataContainer()` vs `\(myProperty)` - this should be relatively simple (essentially prohibiting `\(...)` with a regex

PR #688, opened 2025-05-02
Recording permission check at SDK level

hiroshihorie:


pblazej (commented):
- Starting a fresh app with `try! AudioManager.shared.setRecordingAlwaysPreparedMode(true)` looks like a deadlock (or the popup appears underneath)
- What's the reason for this change?

hiroshihorie:
@pblazej I think you need to test with the LKTestHost app, since the permissions dialog blocks the test. However, I believe I can disable this check for the Simulator.

The purpose of this check is that in some cases, the audio engine appears to start but only produces empty frames. I want to ensure the condition is valid, otherwise, an error should be thrown.

pblazej:
> @pblazej I think you need to test with the LKTestHost app, it might be blocking at the permissions dialog. However, I believe I can disable this check for the Simulator.

Hm, I'm testing with `LiveKitExample` (device) - I think it can be reproduced with any other (consumer) app.

pblazej:
You can add the changelog for these older prs.

PR #690, opened 2025-05-05
Ensure audio frame when publishing

hiroshihorie:
Ensure at least one audio frame was generated before returning publish().


pblazej (approved):
LGTM 🟢 maybe a little comment, why this dependency is even needed? What if we publish before rendering?

hiroshihorie:
If I remember correctly, if no audio frames arrive, the server fails to create the track. The publish() call still returns but silently fails to publish.

boks1971:
> If I remember correctly, if no audio frames arrive, the server fails to create the RTC track. The publish() call still returns but silently fails to publish.
> 
> Am I right ? @boks1971 😅

Yes and no @hiroshihorie .

Yes - server needs at least one packet to declare `onTrack`. 
No - going forward, pion can fire on track based on negotiated SDP and does not need a packet (still needs it for simulcast video tracks, but if the SDP contains the SSRC of the track, it can fire onTrack without an RTP packet). But, that feature is not enabled yet in cloud production deploy.

hiroshihorie:
Thanks for the details @boks1971 🙏

Ideally, if track.start() succeeds, we can guarantee that frames will be generated. However, in some edge cases it appears to fail silently, so I think we should add something like this PR as an interim additional check.

ladvoc (approved):
LGTM ✅

PR #691, opened 2025-05-06
Fix external webcam visibility on macOS

pblazej:
Handles `.unspecified` position

### DIFF
--- a/Sources/LiveKit/Extensions/AVCaptureDevice.swift
+++ b/Sources/LiveKit/Extensions/AVCaptureDevice.swift
@@ -38,8 +38,12 @@ public extension AVCaptureDevice {
 public extension Collection where Element: AVCaptureDevice {
     /// Helper extension to return only a single suggested device for each position.
     func singleDeviceforEachPosition() -> [AVCaptureDevice] {
+        #if os(macOS)
+        return Array(self) // no-op, allows external webcams
+        #else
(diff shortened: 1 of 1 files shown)


hiroshihorie:
I've added `facingPosition` as an extension on `AVCaptureDevice` so please take a look there (if you haven't already) 🙂

pblazej:
> I've added `facingPosition` as an extension on `AVCaptureDevice` so please take a look there (if you haven't already) 🙂

I think it's too restrictive tho `if deviceType == .builtInWideAngleCamera` - it does not handle external cams. We can remove this part instead.

pblazej:
> > I've added `facingPosition` as an extension on `AVCaptureDevice` so please take a look there (if you haven't already) 🙂
> 
> I think it's too restrictive tho `if deviceType == .builtInWideAngleCamera` - it does not handle external cams. We can remove this part instead.

Hm, but it comes at the cost of filtering out built-in camera. Are we sure we don't wanna handle all 3 positions instead?

hiroshihorie:
@pblazej I'm open to changes, as long as we don't break camera publish & flip logic. 🙂

pblazej:
Closing in favor of https://github.com/livekit-examples/swift-example/pull/73 - the only idea is easier testing on macOS without breaking any contracts 🤞

PR #692, opened 2025-05-06
Echo agent tests

hiroshihorie:
WIP, needs echo agent to be running.

### DIFF
--- a/Tests/LiveKitTests/Audio/EchoTesting.swift
+++ b/Tests/LiveKitTests/Audio/EchoTesting.swift
@@ -1,0 +1,113 @@
+/*
+ * Copyright 2025 LiveKit
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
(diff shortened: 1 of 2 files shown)


pblazej:
Can we just deploy the agent with `agents-hosting`?

hiroshihorie:
@pblazej That would be much better 🙂

pblazej:
> @pblazej That would be much better 🙂

Almost there @hiroshihorie, just waiting for beta access to hosting [for the project]

pblazej:
Screenshot 2025-08-20 at 2 13 49 PM

Thanks to cloud updates, the agent is deployed 🎉

pblazej (approved):

pblazej:
@hiroshihorie if it's meant to run as an integration test (on device) I'd rather create a token server and leave empty credentials in the test file.

pblazej:
I mean in the same way you did here:

```swift
class PublishDeviceOptimizationTests: LKTestCase {
    // For testing remote server:
    let url: String? = nil
    let token: String? = nil
```

PR #693, opened 2025-05-07
Extract agent components

pblazej:


Sources/LiveKit/Agent/Participant+Agent.swift
```diff
@@ -0,0 +1,35 @@
...
+        switch kind {
+        case .agent: return true
+        default: return false
+        }
+    }
+
+    @objc
+    var agentState: AgentState {
+        guard isAgent else { return .unknown }
+        guard let attrString = _state.attributes[agentStateAttributeKey] else { return .connecting }
```
pblazej, line 31: I'm more into breaking the contract and returning `.unknown` here and below (makes more sense?)

Sources/LiveKit/Agent/Room+Agent.swift
```diff
@@ -0,0 +1,39 @@
...
+import Foundation
+
+let publishOnBehalfAttributeKey = "lk.publish_on_behalf"
+
+public extension Room {
+    /// Returns a dictionary containing all agent participants.
+    @objc
+    var agentParticipants: [Participant.Identity: Participant] {
+        // Filter out agents that are replaced by another agent e.g. avatar worker
+        let onBehalfIdentities = Set(allParticipants.compactMap {
```
pblazej, line 26: I decided to hide this kind of complexity - plays nicely with UI, advanced users will filter with their custom logic anyway.

Sources/LiveKit/Agent/Room+Agent.swift
```diff
@@ -0,0 +1,39 @@
...
+            $0.value.attributes[publishOnBehalfAttributeKey]
+        })
+        return allParticipants.filter {
+            $0.value.isAgent && !onBehalfIdentities.contains($0.key.stringValue)
+        }
+    }
+
+    /// Returns the first agent participant or `nil` if there are no agent participants.
+    @objc
+    var agentParticipant: Participant? {
```
pblazej, line 36: IMO returning `Participant` vs `RemoteParticipant` is okayish as there's no rigid relationship (every `Participant` can be `.agent` and have `agentState` - already public)

Sources/LiveKit/Agent/Room+Agent.swift
pblazej: BTW, this is helpful for avatars mainly.

hiroshihorie (approved):

Sources/LiveKit/Agent/Participant+Agent.swift
hiroshihorie: I agree `.unknown` makes more sense but I think I got this default from the JS SDK, can you take a look (search) ?

Sources/LiveKit/Agent/Room+Agent.swift
bcherry: are you also planning to add the avatar worker tracks into the agent participant's track collections?

bcherry (approved):

Sources/LiveKit/Agent/Participant+Agent.swift
bcherry: if you're using the new instant listening feature then you'd probably want your agent UI to be in the "listening" state right away. we should think about how to make that easy to do as we work on this. you wouldn't have an agent and they wouldn't have an attribute, but the component should still render

pblazej: Yeah definitely it can make our lives harder in certain (or many) scenarios... the problem is I'd rather fix the source of truth so this value `_state.attributes[agentStateAttributeKey]` rather than adding some magic to the enum (like above).

In other words, the agent should report its initial state as `.listening` if possible. Let me double-check how it looks like in a real-world scenario (real agent).

Sources/LiveKit/Agent/Room+Agent.swift
pblazej: good q, initially I wanted to "replace the participant" without moving the tracks around - as it will be less surprising.

also the recovery is kinda easier - assuming that this API is inconvenient for some advanced cases, you can just re-filter participants; with tracks it's kinda unpredictable.

@lukasIO any opinion on this one? you'll face it as well with avatars (or it's already there)

lukasIO: yeah, that's going to be tricky to handle transparently. 
replacing the participant raises a bunch of questions around eg attributes and metadata set on the "original" agent. What happens with those?

@bcherry added support for avatar tracks on the JS side for the components framework only right now. 
The thinking being, that users using the client SDK directly will be comfortable filtering this themselves and it's ok to make high-level assumptions on the components level.

pblazej: > eg attributes and metadata

yes, discovering e.g. available input modes for the agent becomes impossible.

The alternative (that I reverted) was introducing an explicit concept of worker with some drawbacks - you need to handle it explicitly in the UI e.g.

```swift
    /// All avatar workers in the Room
    var avatarWorkers: [Participant.Identity: RemoteParticipant] {
        remoteParticipants.filter { $0.value.isAgent && $0.value.isAvatarWorker }
    }

    /// The first avatar worker in the Room
    var avatarWorker: RemoteParticipant? {
        avatarWorkers.values.first
    }
```

then

```swift
if let participant = room.avatarWorker {
...
} else if let agent = room.agentParticipant {
...
}
```

hiding that down in the UI component itself is not great - as it also obfuscates the idea of `Participant`...

pblazej: @lukasIO conceptually, we can do sth like that - move the "ordering" to the SDK
241e394f99de72dabcd1bbfd33845e34d002271a

bcherry: worker is also dependent on agent anyways. so maybe good to start with being able to lookup workers from the agent participant object and then let the components library automatically work out the implications for tracks: `room.agentParticipants.first.avatarWorker`

Sources/LiveKit/Agent/Room+Agent.swift
```diff
@@ -0,0 +1,37 @@
...
+
+    @objc
+    var agentParticipant: Participant? {
+        agentParticipants.values.first
+    }
+
+    @objc
+    var avatarOrAgentParticipant: Participant? {
+        agentParticipants.values.filter { $0.attributes[publishOnBehalfAttributeKey] != nil }.first ?? agentParticipants.values.first
+    }
```
bcherry, line 36: i don't see the need for this at the room level, and i'd make it so agentParticipants explicitly does not include avatar workers

pblazej: yes I think nesting is the smartest option, will go this route once finishing preconnect stuff

pblazej: Done 6695e2f7cce9055739cec2852204eee8fbc83cc7

PR #642, opened 2025-03-21
Expose reconnect mode

pblazej:
The solution is a bit awkward, as `connectionState` does not cover `.quick`, but probably the only non-breaking one.

Resolves #368

PR #694, opened 2025-05-12
Expose complete reconnect mode cycle

bombruch:
**Issue**
@pblazej did an amazing job in https://github.com/livekit/client-sdk-swift/pull/642 with exposing reconnect mode.
But it happens to give info when SDK goes into `.quick` or `.full` reconnect but not back to normal connect state.
So with current logic there is no way to get info when SDK establishes connection back and it is time to hide reconnect UI elements.

**Solution**
Expose full cycle for `ReconnectMode`. Since SDK uses Swift optional enum `ReconnectMode?` there is no way to notify ObjC delegate with changes. Idea is to present `.none` ReconnectMode option and use it instead of optional. With such approach it is feasible to notify ObjC delegates. With such logic SDK can tell clients that reconnection process is over. Still not sure wether `.none` is the best name for such option so may be reviewers could suggest any better.

### DIFF
--- a/Sources/LiveKit/Core/Room+Engine.swift
+++ b/Sources/LiveKit/Core/Room+Engine.swift
@@ -268,7 +268,7 @@ extension Room {
         log("\(_state.connectStopwatch)")
     }
 
-    func startReconnect(reason: StartReconnectReason, nextReconnectMode: ReconnectMode? = nil) async throws {
+    func startReconnect(reason: StartReconnectReason, nextReconnectMode: ReconnectMode = .none) async throws {
         log("[Connect] Starting, reason: \(reason)")
 
         guard case .connected = _state.connectionState else {
@@ -286,7 +286,7 @@ extension Room {
             throw LiveKitError(.invalidState)
         }
 
-        guard _state.isReconnectingWithMode == nil else {
+        guard _state.isReconnectingWithMode == .none else {
             log("[Connect] Reconnect already in progress...", .warning)
             throw LiveKitError(.invalidState)
         }
@@ -368,7 +368,8 @@ extension Room {
                                         return delay
                                     }) { currentAttempt, totalAttempts in
                 // Not reconnecting state anymore
-                guard let currentMode = self._state.isReconnectingWithMode else {
+                let currentMode = self._state.isReconnectingWithMode
+                guard currentMode != .none else {
                     self.log("[Connect] Not in reconnect state anymore, exiting retry cycle.")
                     return
                 }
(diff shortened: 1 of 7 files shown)


pblazej:
Just fyi my initial idea was to use it in conjunction with the global [state](https://github.com/livekit/client-sdk-swift/blob/4b582a1e6811b7e629a9bd6d21e2ffe0809c6359/Sources/LiveKit/Protocols/RoomDelegate.swift#L39C11-L39C12) so that we avoid changes in the reconnect state machine.

I'm open to re-discussing this one cc @hiroshihorie

pblazej:
@bombruch can you confirm that the above solution may work for you first?

bombruch:
@pblazej  of corse it does.  All below is based on assumption  that `ConnectOptions` is created with:

```
 ConnectOptions(
    reconnectAttempts: 30,
    reconnectAttemptDelay: 1
)
```

Consider 2 cases:

**First case**) WiFi is turned off for a long period  (~20 seconds). This case is covered with existing logic in LK SDK. As shortened logs look like:

> 18:22:54.629  didUpdateReconnectMode, reconnectMode: .quick

Time to show reconnect UI. 
Wait for ~20 seconds 
Turn on WiFi  (it takes time to recover WiFi)

> 18:23:36.961  didUpdateReconnectMode, reconnectMode: .full
> 18:23:36.963  didUpdateConnectionState: .reconnecting, oldValue: .connected, error: empty error
> 18:23:36.966 roomIsReconnecting
> 18:23:38.452 didUpdateConnectionState: .connected, oldValue: .reconnecting, error: empty error
> 18:23:38.462 roomDidReconnect

Based on last 2 lines in logs above - there is info to hide reconnect UI

**Second case**) WiFi is turned off for a short period  (~5 seconds).  

> 11:29:59.375 didUpdateReconnectMode, reconnectMode: .quick

After WiFi  connection is restored it can  tell about `didUpdateConnectionQuality ` (it happens mostly if camera is active) but that is basically it. I mean no logs as no delegate methods are fired.

With fix for second case  I can clearly see logs like:

> 11:47:28.638 didUpdateReconnectMode, reconnectMode: .quick
> 11:47:35.871  didUpdateReconnectMode, reconnectMode: .none

During preparation of this comment I noticed that I've broken your logic with `reconnect` vs `connect` state. For sure I can't use Xcode search tool. So I've updated PR with second commit and got a bit embarrassed - seems like too many changes to get full cycle in delegate. May be something like separate Enum for ObjC can be presented and used for delegate. While in SDK still original is used. In other words this PR could be like in screen shot below:

Screenshot 2025-05-20 at 12 04 32


Pls tell me your thoughts on all above. Thanks!

PR #696, opened 2025-05-13
Update audio session logic

hiroshihorie:
Fixes `isSpeakerOutputPreferred = false` not working.


pblazej:
Tested with BT speaker, for me this flag still does nothing (until I just switch the device via AirPlay menu).

hiroshihorie:
Thanks for testing 🙏

PR #702, opened 2025-05-16
Wait until active (remote participant)

pblazej:
### Problem

Our data-related APIs do not (or indirectly) validate the state of their `destinationIdentity` which seriously impacts the reliability of such channels. It moves the responsibility to handle these state transitions to the consumer, respond to proper `Participant` lifecycle events, etc.

Reasoning about the order of events is impossible without server-side knowledge, handling all the combinations - highly impractical, and counter-intuitive (like participant already subscribed to a track, but cannot receive data from a data channel).

Consumers shouldn't be exposed to transport internals and their distributed nature. In most practical applications, we can just assume that if the `Participant.Identity` is returned somewhere, it's also "valid" for communication (regardless of the channel).

### Swift Status Quo

Sending messages to inactive/non-existing participants will lead to:

- 🟡 `room.localParticipant.performRpc` - will `timeout` after 10s, but the API suggests:
```swift
    public enum BuiltInError {
        case connectionTimeout
        case responseTimeout // this is triggered
        case recipientDisconnected
        case sendFailed
        case recipientNotFound
```
- 🔴 `room.localParticipant.publish(data:)` - only checks publisher's own transport via `publisherTransportConnectedCompleter` - will return immediately
- 🔴 `room.room.localParticipant.sendText` and other streams - will just return immediately

#### Publisher

Always check for local transport, then `await`, reuse this part across 3+ APIs. This change is purely internal and should not break any existing behaviors.

#### Receivers

Add public option(s) to wait for the receivers with a consistent timeout, reuse this part across 3+ APIs. Technically, it's not super hard, as all the APIs are already `async`. New errors for opt-iners.

RPC example attached.

### Questions

- Which modes do we really need?
  - The most practical one is just `waitUntilActive` (mirroring publisher side), forming `Participant.Identity` from `String` should not be that popular, so we assume the `Participant` exists somewhere among `allParticipants` already. If not, just throw immediately.

- "As soon as someone joins"
  - To handle that implicitly, we'd need an extremely long timeout, that could be covered by "joined event" → `waitUntilActive` sequence.

- How to handle multiple destinations (any/all = optimistic/pessimistic)?
  - Can we even predict the use cases here?


pblazej:
Per our discussion with @lukasIO the bare minimum:
- wait until publisher active (that's more SDK bug than actual change)
- wait until receivers active
  - wait concurrently for each and every receiver → send
  - fail/throw if no one activates until timeout
  - success if at least one activates within timeout

pblazej:
@bcherry @lukasIO we can introduce `waitUntilActive()` for the local participant, but:
- it must be called explicitly (no implicit benefits)
- it does not solve the problem of `[Participant.Identity]` - for multiple receivers, unless we choose the convention here (don't wait for all e.g.)

lukasIO:
what's the benefit on having `waitUntilActive` on the _local_ participant?

pblazej:
> what's the benefit on having `waitUntilActive` on the _local_ participant?

No benefit, apart from what's actually implemented for data packets:
```swift
    func send(dataPacket packet: Livekit_DataPacket) async throws {
        func ensurePublisherConnected() async throws {
```

For remotes, as I said doing `waitUntilActive()` on a single `Participant.Identity` is also cumbersome, as all the APIs are based on `[Participant.Identity] - consumer needs to do the heavy lifting (fork, async stuff, etc.).

.swift-version
pblazej, line 1: This must be reverted as the formatters/linters will respect that and break the old-compatible version.

Does not make much difference for the package/other infra though.

pblazej:
@lukasIO I tried to simplify my initial vision a bit, mimicking your `waitUntilActive()` approach, I appreciate further feedback (if any).

lukasIO (approved):
this looks good to me! 

for `waitForAllActive` there's somewhat of an edge case when participants join while you're waiting for others to become active. 

The question would be what's expected here for users, is it obvious enough that it will only ensure activeness of the participants present at method invocation time? 

would probably make sense to have one of our other designated swift experts to have a look at the PR

pblazej:
> for `waitForAllActive` there's somewhat of an edge case when participants join while you're waiting for others to become active.
> 
> The question would be what's expected here for users, is it obvious enough that it will only ensure activeness of the participants present at method invocation time?

Not sure if I understand the 1st part properly, the 2nd cannot be resolved easily, even when moving the call to the "inside" of `sendSomething` method - it's not "atomic" in this sense, the only guarantee is "at the time of calling" not "with active".

lukasIO:
haha it's supposed to be only one point describing the same "edge" case, I think it's fine to leave it as is.

pblazej:
@hiroshihorie are you able to take a look? It's a pretty isolated change.

pblazej:
Looks like pods are failing due to the missing sim as well, skipping that for now 👍

PR #703, opened 2025-05-19
macOS audio device improvements

hiroshihorie:
* Better mic starting for macOS when external device is used.
* Better handling of default device switching.

Patch https://github.com/webrtc-sdk/webrtc/commit/a58a087f6909d49d1a24dcfea1a66752ee8f6e5e


hiroshihorie:
It's a weird issue in `AVAudioEngine` that despite setting it up correctly, it fails to start. I think it's related to initializing the voice-processing internally. AVAudioEngine has [weird bugs](https://stackoverflow.com/a/39936853)...

I can occasionally reproduce it while another video call is running in the background, the audio engine fails to start sometimes:
https://github.com/user-attachments/assets/1207e211-492d-47d2-aa67-d4ec0d354cdb

After patch, more reliably starts:
https://github.com/user-attachments/assets/845895b5-6810-4525-80c9-7fff35a05453

I will add a small test.

hiroshihorie:
@pblazej I manually tested this for macOS, I think it's good to merge.

Things to look on macOS are :
* If mic starts smoothly with/without bluetooth earphones etc.
* Switches smoothly when changing default audio device from control center.

pblazej (approved):

PR #706, opened 2025-05-21
Fix race condition during remote track deinit

### LINKED ISSUE #704: A deadlock can occur when joining a new room after disconnection
**Describe the bug**
Disconnecting from a room sometimes hangs. Then when joining a new room, creating a new peer connection hangs, locking the liveKitWebRTC queue. It doesn't happen every time, but we have reproduced it with a fork of the example project.

**SDK Version**
2.5.1

**iOS/macOS Version**
iOS 18 (various, but most recently 18.4.1)

**Xcode Version**
16.3, Swift 5 or 6

**Steps to Reproduce**
- Run a [token-server](https://github.com/livekit-examples/token-server-node?tab=readme-ov-file) sandbox
- Clone our fork of the example project: https://github.com/Our-Kinship/livekit-swift-example
- Replace YOUR_SANDBOX_ID in ConnectView.swift with the ID of the above sandbox
- Build and install it on at least 2 devices
- Run them concurrently for a few minutes

All connection and disconnection code remains the same, but the app automatically joins a room and disconnects every 10 seconds.

**Expected behavior**
All participants should be able to connect and disconnect without deadlocks.

As far as I can tell, this is the sequence of events:

1. `_pc.close` hangs indefinitely when closing the subscriber connection.

```swift
    func close() async {
        // prevent debounced negotiate firing
        await _debounce.cancel()

        // Stop listening to delegate
        _pc.delegate = nil
        // Remove all senders (if any)
        for sender in _pc.senders {
            _pc.removeTrack(sender)
        }

        _pc.close()
    }
```

2. When initiating a new subscriber connection, PeerConnectionFactory.peerConnection hangs because the previous one was never closed.

```
    static func createVideoTrack(source: LKRTCVideoSource) -> LKRTCVideoTrack {
        DispatchQueue.liveKitWebRTC.sync { peerConnectionFactory.videoTrack(with: source,
                                                                            trackId: UUID().uuidString) }
    }
```

At this point, all usages of `DispatchQueue.liveKitWebRTC` will wait indefinitely. Some of these calls happen from the main thread, so this can also result in freezing.

I suspect `_pc.close` is waiting to run on the signaling thread, but so far I haven't been able to identify what the signaling thread is busy with.

pblazej:
Resolves #704 

Removes a mini-optimization while holding empty delegates, simplifying the delegate chain a bit.

- `deinit` is `nonisolated`
- operations on `LKRTCAudioTrack` are not thread-safe
- removing "the last delegate" can happen from an arbitrary thread (e.g. SwiftUI main if I'm the visualizer)
- `deinit` may run concurrently, e.g. when a whole remote track is destroyed during disconnection
- in the worst-case scenario, a deadlock may occur inside rtc, around:
```
frame #0: 0x0000000103140014 libsystem_kernel.dylib`__psynch_cvwait + 8
frame #1: 0x000000010302aab8 libsystem_pthread.dylib`_pthread_cond_wait + 976
frame #2: 0x00000001048442b4 LiveKitWebRTC`___lldb_unnamed_symbol9036 + 528
frame #3: 0x0000000104841920 LiveKitWebRTC`rtc::Thread::BlockingCallImpl(rtc::FunctionView<void ()>, webrtc::Location const&) + 384
frame #4: 0x00000001047e7bdc LiveKitWebRTC`___lldb_unnamed_symbol7471 + 284
frame #5: 0x00000001061c2dd4 LiveKitExample.debug.dylib`RemoteAudioTrack.remove(audioRenderer=(object = 0x000060000177dd40 -> 0x000000010677cc18 type metadata for LiveKitComponents.AudioProcessor)) at RemoteAudioTrack.swift:75:24
frame #7: 0x00000001065f49a4 LiveKitExample.debug.dylib`AudioProcessor.deinit() at AudioProcessor.swift:47:16
```

### DIFF
--- a/Sources/LiveKit/Track/Remote/RemoteAudioTrack.swift
+++ b/Sources/LiveKit/Track/Remote/RemoteAudioTrack.swift
@@ -69,10 +69,5 @@ public class RemoteAudioTrack: Track, RemoteTrack, AudioTrack, @unchecked Sendab
 
     public func remove(audioRenderer: AudioRenderer) {
         _adapter.remove(delegate: audioRenderer)
-        // Remove adapter only if there are no more delegates
-        if _adapter.countDelegates == 0 {
-            guard let audioTrack = mediaTrack as? LKRTCAudioTrack else { return }
-            audioTrack.remove(_adapter)
-        }
     }
 }
--- a/Sources/LiveKit/Track/Remote/RemoteAudioTrack.swift
+++ b/Sources/LiveKit/Track/Remote/RemoteAudioTrack.swift
@@ -49,22 +49,20 @@ public class RemoteAudioTrack: Track, RemoteTrack, AudioTrack, @unchecked Sendab
                    source: source,
                    track: track,
                    reportStatistics: reportStatistics)
+
+        if let audioTrack = mediaTrack as? LKRTCAudioTrack {
+            audioTrack.add(_adapter)
+        }
     }
 
     deinit {
-        // Directly remove the adapter without unnecessary checks
-        guard let audioTrack = mediaTrack as? LKRTCAudioTrack else { return }
-        audioTrack.remove(_adapter)
+        if let audioTrack = mediaTrack as? LKRTCAudioTrack {
+            audioTrack.remove(_adapter)
+        }
     }
 
     public func add(audioRenderer: AudioRenderer) {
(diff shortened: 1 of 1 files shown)


hiroshihorie (approved):

schenkty:
Excited to get this merged! Thank you @pblazej for the quick response

PR #715, opened 2025-06-06
Version script

pblazej:
- Keeps the old nanpa format and parses that (moves to the `.changes` directory)
- Replaces the old bash script for versioning
- Everything else is opinionated, so feel free to comment/change if needed

IMO, it's _good enough_ for our simple use case


pblazej:
> Wow written in Swift 🙂 LGTM, but I think @ladvoc had some ideas/plans .

Yeah, it's not a ton of work, so feel free to discard it if any better/global solution.

ladvoc (approved):
Looks good to me! I’ve been working on a more feature-rich Rust-based tool on the weekends, but it’s still a work in progress. For this repository, this is a solid solution.

PR #723, opened 2025-06-17
v2.6.1

hiroshihorie:


CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## [2.6.1] - 2025-06-17
+
+### Fixed
+
+- Fixed Xcode 26 build errors with Swift 6.2
```
pblazej, line 7: We can add missing entries "in between" nanpa and the new script here.

PR #737, opened 2025-07-14
Reliable data channel

pblazej:
Ports basic concepts from https://github.com/livekit/client-sdk-js/pull/1546

The differences are mostly syntactic, due to how backpressure is handled via events in Swift. Thus, I decided to separate `SendBuffer` that's tightly coupled to RTC and `RetryBuffer` with slightly different semantics.

### DIFF
--- a/Sources/LiveKit/Core/DataChannelPair.swift
+++ b/Sources/LiveKit/Core/DataChannelPair.swift
@@ -40,6 +40,10 @@ class DataChannelPair: NSObject, @unchecked Sendable, Loggable {
         var lossy: LKRTCDataChannel?
         var reliable: LKRTCDataChannel?
 
+        var reliableMessageBuffer: [Livekit_DataPacket] = []
+        var reliableDataSequence: UInt32 = 1
+        var reliableReceivedState: [String: UInt32] = [:]
+
         var isOpen: Bool {
             guard let lossy, let reliable else { return false }
             return reliable.readyState == .open && lossy.readyState == .open
@@ -225,7 +229,7 @@ class DataChannelPair: NSObject, @unchecked Sendable, Loggable {
     }
 
     public func send(dataPacket packet: Livekit_DataPacket) async throws {
-        let serializedData = try packet.serializedData()
+        let serializedData = try resendable(packet).serializedData()
         let rtcData = RTC.createDataBuffer(data: serializedData)
 
         try await withCheckedThrowingContinuation { continuation in
@@ -241,10 +245,43 @@ class DataChannelPair: NSObject, @unchecked Sendable, Loggable {
         }
     }
 
+    private func resendable(_ packet: Livekit_DataPacket) -> Livekit_DataPacket {
+        guard packet.kind == .reliable, packet.sequence == 0 else { return packet }
+        var packet = packet
+        _state.mutate {
+            packet.sequence = $0.reliableDataSequence
+            $0.reliableDataSequence += 1
+            $0.reliableMessageBuffer.append(packet)
+        }
+        return packet
+    }
+
+    public func resendReliable(resumingFrom lastSeq: UInt32) {
+        print("🔥 resending from", lastSeq)
+        _state.mutate {
+            $0.reliableMessageBuffer.removeAll(where: { $0.sequence <= lastSeq })
+            for packet in $0.reliableMessageBuffer {
+                Task {
+                    print("resending", packet.sequence)
+                    try? await send(dataPacket: packet)
+                }
+            }
+        }
+    }
+
     public func infos() -> [Livekit_DataChannelInfo] {
         _state.read { [$0.lossy, $0.reliable] }
             .compactMap { $0 }
             .map { $0.toLKInfoType() }
+    }
+
+    public func receiveStates() -> [Livekit_DataChannelReceiveState] {
+        _state.reliableReceivedState.map { sid, seq in
+            Livekit_DataChannelReceiveState.with {
+                $0.publisherSid = sid
+                $0.lastSeq = seq
+            }
+        }
     }
 
     private static let reliableLowThreshold: UInt64 = 2 * 1024 * 1024 // 2 MB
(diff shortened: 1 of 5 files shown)


Sources/LiveKit/Core/DataChannelPair.swift
```diff
@@ -207,6 +293,8 @@ class DataChannelPair: NSObject, @unchecked Sendable, Loggable {
         let (lossy, reliable) = _state.mutate {
             let result = ($0.lossy, $0.reliable)
             $0.reliable = nil
+            $0.reliableDataSequence = 1
+            $0.reliableReceivedState = [:]
```
pblazej: Will reset on full reconnect, similarly to JS

Sources/LiveKit/Core/DataChannelPair.swift
```diff
@@ -241,15 +333,47 @@ class DataChannelPair: NSObject, @unchecked Sendable, Loggable {
...
+    }
+
+    // MARK: - Constants
+
     private static let reliableLowThreshold: UInt64 = 2 * 1024 * 1024 // 2 MB
     private static let lossyLowThreshold: UInt64 = reliableLowThreshold
 
+    // Keep at least N reliable packets while draining webrtc buffers.
+    // Should prevent losing packets - requesting retry from lastSeq that's already dequeued.
+    private static let reliableRetryCapacity = 32
```
pblazej: This is kinda conservative, but during device testing I discovered that a gap in the history may occur:

Screenshot 2025-07-14 at 10 12 57 AM

When we drain our buffer to `0` as soon as RTC part does it... So drained != sent 😭 

We could also keep a fixed-size queue of let's say 1MB - the RAM sacrifice is negligible if we want real consistency here.

Sources/LiveKit/Core/DataChannelPair.swift
cnderrauber: js clean the state with a ttl to prevent it from expanding too large, is is required in iOS?

cnderrauber (approved):
looks good!

Sources/LiveKit/Core/DataChannelPair.swift
pblazej: we discussed that with @lukasIO, hypothetically we can do that, or even use `NSCache` (just reacting to memory warnings) with no aux data structures.

pblazej: sorry we cannot do `NSCache` because we need to enumerate keys (while sending to backend).

pblazej: imo huge number of entries probably would mean other problems as well (huge number of `RemoteParticipant`s in general)

lukasIO: sorry to make a bit of a turnaround here, but a potentially huge number of remote participants is something we should  be designing for/keeping in mind, it's definitely a valid use case

pblazej: yes, it just means: everything should be capped then 🥲

pblazej: here's a more-or-less-JS version: [`d36df08` (#737)](https://github.com/livekit/client-sdk-swift/pull/737/commits/d36df082444d7fb226fb15747150e58d63f9d4aa)

Sources/LiveKit/Core/DataChannelPair.swift
lukasIO: in what test cases did this happen? 
curious if you have repro steps so that we can ensure we don't have the same problem on JS

pblazej: It happened with real n/w (real latency). Local server (used for unit tests) is so fast that the reconnection delay may not even interfere with the transfer, so it mostly checks the "duplicate logic".

To test it "live" I'm just sending a large packet indefinitely, firing a `Quick reconnect` in the meantime, then the server requests retry...

```swift
while !Task.isCancelled {
  try? await room.localParticipant.streamText(for: "asd").write(String(repeating: "a", count: 1024 * 100))
  try? await Task.sleep(100 * NSEC_PER_MSEC)
}
```

pblazej: Maybe still specific to the webrtc internals or even iOS n/w stack, but worth checking 💣

pblazej: @lukasIO here's an example for bigger test data 4f55e6fb42d0acd346f45e8b46a00935d186b6b5

However, keep in mind that:
- everything may depend on timing here, from `yield`ing the cooperative concurrent stuff to n/w delays, chunk size, etc.
- cli requests retry from `0` - looks like it does not respect the "receive state" part yet

So this is just an attempt to replicate the real-world condition.

pblazej:
@cnderrauber @lukasIO for the record:

I did a lot of semi-automated monkey-testing on iOS and macOS (like reconnecting every N s while streaming data). It turns out that *it depends on how you test it* aka *how hard you stress it*.

By *stress*, I mostly mean whether you `sleep` 🛌 or not in between calls to `stream...`.

If you keep streaming **continuously**, you gotta buffer at least the full webrtc capacity (backpressure amount) + a few packets (I just did 2x backpressure). Otherwise, you get:

```
2025-07-15T15:37:37+0200 warning LiveKitSDK : [LiveKit] DataChannelPair.retry(buffer:from:) Wrong packet sequence while retrying: 913 > 910, 3 packets missing
2025-07-15T15:37:43+0200 warning LiveKitSDK : [LiveKit] DataChannelPair.retry(buffer:from:) Wrong packet sequence while retrying: 1698 > 1694, 4 packets missing
2025-07-15T15:37:48+0200 warning LiveKitSDK : [LiveKit] DataChannelPair.retry(buffer:from:) Wrong packet sequence while retrying: 2508 > 2502, 6 packets missing
2025-07-15T15:37:53+0200 warning LiveKitSDK : [LiveKit] DataChannelPair.retry(buffer:from:) Wrong packet sequence while retrying: 3346 > 3342, 4 packets missing
2025-07-15T15:37:58+0200 warning LiveKitSDK : [LiveKit] DataChannelPair.retry(buffer:from:) Wrong packet sequence while retrying: 4178 > 4176, 2 packets missing
2025-07-15T15:38:04+0200 warning LiveKitSDK : [LiveKit] DataChannelPair.retry(buffer:from:) Wrong packet sequence while retrying: 5015 > 5011, 4 packets missing
2025-07-15T15:38:14+0200 warning LiveKitSDK : [LiveKit] DataChannelPair.retry(buffer:from:) Wrong packet sequence while retrying: 6682 > 6680, 2 packets missing
2025-07-15T15:38:25+0200 warning LiveKitSDK : [LiveKit] DataChannelPair.retry(buffer:from:) Wrong packet sequence while retrying: 8113 > 8108, 5 packets missing
```

Once introducing `sleep`, this requirement is relaxed, and you can let it drain.

So in my case, it looks like what entered webrtc **can be lost** (as well as the in-flight data, these few packets).

As I said, it may depend on everything, webrtc stack, Apple's own POSIX impl, etc.

Sources/LiveKit/Core/DataChannelPair.swift
cnderrauber: The ttl map looks good! Since each participant who sends a message will add a entry to the map so it  will expand to a large size even if the room size is not large but there are many participants entering and leaving.

PR #738, opened 2025-07-15
iOS 26 Simulator crash workaround

hiroshihorie:
Workaround: https://developer.apple.com/forums/thread/787365?answerId=842938022#842938022

I think this is a temporary bug in the Beta...


Sources/LiveKit/Support/WebSocket.swift
```diff
@@ -32,6 +33,7 @@ final class WebSocket: NSObject, @unchecked Sendable, Loggable, AsyncSequence, U
     private let request: URLRequest
 
     private lazy var urlSession: URLSession = {
+        nw_tls_create_options()
```
pblazej: We can wrap it in `#swift` for now

pblazej: Out of curiosity, still reproducible in `Xcode 26.0 Beta 3 (17A5276g)`?

hiroshihorie: Yes happens in `iOS 26.0 (23A5287g)` Beta 3.

PR #739, opened 2025-07-15
Add `outputVolume` property to audio mixer

hiroshihorie:


Sources/LiveKit/Audio/MixerEngineObserver.swift
```diff
@@ -50,15 +61,19 @@ public final class MixerEngineObserver: AudioEngineObserver, Loggable {
...
 
         // Not connected for device rendering mode.
-        public let micNode = AVAudioPlayerNode()
-        public let micMixerNode = AVAudioMixerNode()
+        let micNode = AVAudioPlayerNode()
+        let micMixerNode = AVAudioMixerNode()
+
+        // Reference to mainMixerNode
+        weak var mainMixerNode: AVAudioMixerNode?
+        var outputVolume: Float = 1.0
```
pblazej, line 73: Hypothetically we could just read it from `mainMixerNode` defaulting `??` to `0` (as a computed prop). Less state, fewer errors.

pblazej (approved):
LGTM, works 👍

Sources/LiveKit/Audio/MixerEngineObserver.swift
hiroshihorie: Thanks for noticing that, but my intention was that the user might want to mutate outputVolume before mainMixerNode is assigned to that variable, since it’s only first assigned when audio is used.

PR #743, opened 2025-07-23
Consume generated attributes

pblazej:
- Adds generated attributes https://github.com/livekit/attribute-definitions
- Consumes agent attributes via `Participant.State` (cache)
- Transcription attributes will come soon with moving the logic from the VoiceAgent → SDK

A basic test case:
- See if the agent (visualizer) state is handled properly
- Avatars are also fine
- You may need https://github.com/livekit/components-swift/pull/27


Sources/LiveKit/Agent/Participant+Agent.swift
```diff
@@ -25,19 +25,29 @@ public extension Participant {
...
     var agentState: AgentState {
-        guard isAgent else { return .unknown }
-        guard let attrString = attributes[agentStateAttributeKey] else { return .unknown }
-        guard let state = AgentState.fromString(attrString) else { return .unknown }
-        return state
+        agentAttributes?.lkAgentState ?? .idle
+    }
+
+    @objc
+    var agentStateString: String {
```
pblazej, line 33: It sort of breaks Obj-C contract, but IMO it's better than changing the schema for every single language, just for the sake of having `Int` raw value here...

Sources/LiveKit/Types/Attributes/AttributeTypings.swift
```diff
@@ -0,0 +1,71 @@
...
+    case text
+    case video
+}
+
+enum AgentOutput: String, Codable, Sendable {
+    case audio
+    case transcription
+}
+
+public enum AgentState: String, Codable, Sendable {
```
pblazej, line 50: This `public` is a manual change, I don't think it's possible to expose a single one vs global `"access-level": "public"`.

Sources/LiveKit/Agent/AgentState+.swift
```diff
@@ -0,0 +1,21 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+extension AgentState: CustomStringConvertible {
+    public var description: String {
+        rawValue.capitalized
```
pblazej, line 19: Happy to remove this one, looks like leftover that could be handled by the frontend.

Sources/LiveKit/Participant/LocalParticipant.swift
```diff
@@ -185,7 +185,10 @@ public class LocalParticipant: Participant, @unchecked Sendable {
     public func set(attributes: [String: String]) async throws {
         let room = try requireRoom()
         try await room.signalClient.sendUpdateParticipant(attributes: attributes)
-        _state.mutate { $0.attributes = attributes }
+        _state.mutate {
+            $0.attributes = attributes
+            $0.agentAttributes = attributes.transcoded(to: AgentAttributes.self)
```
pblazej: Theoretically we could skip that for `LocalParticipant`, but it's not enforced anywhere else...

pblazej: Seems unused (but `public`), so I reverted that...

PR #745, opened 2025-07-25
Update webrtc to m137

pblazej:


pblazej:
@hiroshihorie this one still requires another update, right? https://github.com/webrtc-sdk/webrtc/pull/186/files

hiroshihorie:
I think we can ship that one with the h265 patch 🙂

PR #749, opened 2025-08-05
Remove lazy init for incoming streams

pblazej:
It may return a brand-new instance of `IncomingStreamManager` when run concurrently:

```
// Before:
self 0x0000600003352b50 ["cover": (Function)]
self 0x0000600003341d40 ["now_playing": (Function)]
// After:
self 0x0000600001eb81b0 ["now_playing": (Function)]
self 0x0000600001eb81b0 ["cover": (Function), "now_playing": (Function)]
```

Thus leading to UB - missing registration (random).

TL;DR be careful with `lazy` consumed via `async` methods.

### DIFF
--- a/.changes/concurrent-streams
+++ b/.changes/concurrent-streams
@@ -1,0 +1,1 @@
+patch type="fixed" "Fixed concurrent registration for text/byte streams"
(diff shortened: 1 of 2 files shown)


hiroshihorie (approved):

hiroshihorie:
Wow interesting..

PR #761, opened 2025-08-20
Improve disconnect vs reconnect reliability

### LINKED ISSUE #757: Disconnect is unreliable
**Describe the bug**
Room.disconnect sometimes fails. In one degenerate case, disconnect failed for another user and myself, _and we could still hear each other!_

**SDK Version**
2.6.1

**iOS/macOS Version**
iOS 18

**Xcode Version**
Xcode 16.4, Swift 5

**Steps to Reproduce**
Call `Room.disconnect` when leaving a room. It may occur more frequently with a poor connection.

**Actual behavior**
Room can remain connected with an active server session even after `disconnect()` returns.

**Expected behavior**
`disconnect()` should always succeed in cleaning up local resources and, when possible, notify the server. The room should never remain connected after `disconnect()` completes.

Here's one possible flow that could result in failure to disconnect:

- `Room.connectionState = .connected`
- Network issue (e.g. failed to receive message from websocket) → `SignalClient.cleanUp`
- Quick reconnect is started, calling `SignalClient.connect`
- `SignalClient.cleanUp`
- `_state.socket` is set to `nil`
- Application calls `Room.disconnect()`

```
public func disconnect() async {
      if case .disconnected = connectionState { return } 

      do {
          try await signalClient.sendLeave()
      } catch {
          log("Failed to send leave with error: \(error)")
      }

      await cleanUp()
  }
```

- The room state is `.connected` (not yet updated), so we attempt to send leave
- `sendLeave() → _sendRequest() → _requestQueue.processIfResumed()`
- `SignalClient.requireWebSocket` throws because `socket` was set to `nil` during `SignalClient.cleanUp()`, and the new socket hasn't been created yet
- The error propagates, `sendLeave()` fails, and `Room.cleanUp()` is called
- The reconnection **completes** and the room is resurrected

I'm not completely sure about this. There's definitely the possibility that `Room.disconnect` can be called in the middle of `SignalClient.cleanUp`, resulting in some weird combinations of states.

For now, we'll work around it by repeatedly calling `disconnect` until it succeeds. But I think this has been reproducing for a while - maybe the same as #729, #256, #583, #642

pblazej:
Resolves #757 

- Adds cancellation to reconnect, making sure that it cannot continue after `disconnect()` and bring the room back to life
- ⚠️ Optionally, we can introduce a breaking `.disconnecting` state to prevent new tasks from being started (vs cancelling ongoing tasks - a little _less elegant_)
  - Theoretically it should not affect other state transitions, as `disconnect()` is basically "destructive" - the only valid path leads to `.disconnected` during `cleanup()`, however that's not statically enforced in any way
  - On the other hand, it does not provide any added value for other parts of the SDK, just covering this tiny time window

### DIFF
--- a/.changes/disconnect-reconnect
+++ b/.changes/disconnect-reconnect
@@ -1,0 +1,1 @@
+patch type="fixed" "Fixed race condition between reconnect and disconnect leading to failed disconnects"
--- a/.changes/disconnected
+++ b/.changes/disconnected
@@ -1,0 +1,1 @@
+patch type="added" "Added .disconnected connection state"


saarva:
@pblazej this fixed some disconnect issues that we were seeing, do you know if it'll go into the main branch soon?

pblazej:
> @pblazej this fixed some disconnect issues that we were seeing, do you know if it'll go into the main branch soon?

Should be merged soon, the risk is _moderate_. We need to finalize the ongoing discussion about catching more edge cases (see the issue).

pblazej:
> LGTM but If possible, I would rather avoid exposing the `.disconnecting` connection state for developer experience / simplicity. Removing/deprecating this would be hard in the future.

- For the current (major) version - yes, that's a bit cumbersome to introduce it atm
- For the next version - why do we wanna remove that? I think it's a good addition to the (incomplete) state machine, without that you're basically unable to resolve the root cause of this one completely ⬆️ tl;dr now it looks complete (with all async state transitions)

@hiroshihorie I'm fine with reverting the state right now, if we're able to retest it.

pblazej:
```mermaid
stateDiagram-v2
    [*] --> Disconnected

    Disconnected --> Connecting: start connection
    Connecting --> Connected: success
    Connecting --> Disconnected: failed/canceled

    Connected --> Disconnecting: user closes
    Disconnecting --> Disconnected: closed

    Connected --> Reconnecting: network loss / ICE restart
    Reconnecting --> Connected: recovered
    Reconnecting --> Disconnected: failed to reconnect
    Reconnecting --> Reconnecting: retrying
```

PR #769, opened 2025-08-25
`VideoView` flickering

### LINKED ISSUE #722: VideoView Flickers when opponent user rejoins video call
These are the logs :

2025-06-11T14:02:08+0530 debug LiveKitSDK : [LiveKit] RemoteVideoTrack.init(name:kind:source:track:reportStatistics:) Track.dimensions Optional(Dimensions(216x384)) -> Optional(Dimensions(320x180))
2025-06-11T14:02:08+0530 debug LiveKitSDK : [LiveKit] RemoteVideoTrack.init(name:kind:source:track:reportStatistics:) Track.dimensions Optional(Dimensions(320x180)) -> Optional(Dimensions(216x384))
2025-06-11T14:02:08+0530 debug LiveKitSDK : [LiveKit] RemoteVideoTrack.init(name:kind:source:track:reportStatistics:) Track.dimensions Optional(Dimensions(320x180)) -> Optional(Dimensions(216x384))
2025-06-11T14:02:08+0530 debug LiveKitSDK : [LiveKit] VideoView.render(frame:captureDevice:captureOptions:) canRender is false, skipping render...

pblazej:
Resolves #722 

- Fixes `VideoRendererAdapter` never actually removed from the RTC track, causing multiple tracks concurrently rendering into the same view
  - The root cause is tricks like `override func isEqual(_ object: Any?)` won't work (anymore?) on the Obj-C++ side as it uses `==` directly; it's also pretty fragile
  - `uikit-minimal` was a good example as the number of adapters exploded after reusing cells
  - Reproduction steps:
    - create multiple cells in `uikit-minimal` - multiple participants (some of them can be muted)
    - shuffle
    - observe cell behavior and FPS
- Moves the side effects a bit, as forcing them to be synchronous:
  - was kinda unexpected (and unnecessary) when running from the main actor `Handle track removal outside of main queue`
  - can lead to deadlocks in practice

Screenshot 2025-08-28 at 1 37 53 PM


pblazej:
This is no longer true with async renderer creation ofc, we can either reconsider this case or remove...

```swift
// avSampleBufferDisplayLayer should not be nil at this point
XCTAssert(view.avSampleBufferDisplayLayer != nil)
```

Sources/LiveKit/Track/Track.swift
```diff
@@ -127,8 +127,8 @@ public class Track: NSObject, @unchecked Sendable, Loggable {
         var rtpSenderForCodec: [VideoCodec: LKRTCRtpSender] = [:] // simulcastSender
         var rtpReceiver: LKRTCRtpReceiver?
 
-        // Weak reference to all VideoRenderers attached to this track.
-        var videoRenderers = NSHashTable<VideoRenderer>.weakObjects()
+        // All VideoRendererAdapters attached to this track, key/value for direct removal.
+        var videoRendererAdapters = NSMapTable<VideoRenderer, VideoRendererAdapter>.weakToStrongObjects()
```
pblazej, line 131: Hypothetically, we could use `weakToWeak` as well (let them be retained only by RTC side), but assuming add/remove cycle is balanced, we won't get "leaks".

hiroshihorie (approved):
✅ Nice

hiroshihorie:
I think the testAVSampleBufferDisplayLayer is failing, we need view.avSampleBufferDisplayLayer to be available when view.renderMode = .sampleBuffer.

PR #770, opened 2025-08-28
Transceiver crash

pblazej:
Reverts https://github.com/livekit/client-sdk-swift/pull/751

Let's revisit transceiver pools - it requires deeper testing though.


hiroshihorie (approved):
✅ Were you able to reproduce ?

pblazej:
> ✅ Were you able to reproduce ?

Nope, but it looks identical to the one reported in the original issue. Also, theoretically - it may happen.

I'm experimenting with _better_ solutions, will report progress.

PR #773, opened 2025-08-29
Reuse transceiver

pblazej:
Let's see what we can do without patching.

As an alternative: we can create a webrtc patch and see what's really allocated, then try to dump it once the transceiver is stopped.

### DIFF
--- a/Sources/LiveKit/Core/Transport.swift
+++ b/Sources/LiveKit/Core/Transport.swift
@@ -315,11 +315,16 @@ extension Transport {
     func addTransceiver(with track: LKRTCMediaStreamTrack,
                         transceiverInit: LKRTCRtpTransceiverInit) throws -> LKRTCRtpTransceiver
     {
-        guard let transceiver = _pc.addTransceiver(with: track, init: transceiverInit) else {
+        if let existingTransceiver = _pc.transceivers.first(where: { $0.mediaType == .init(trackKind: track.kind) && $0.direction == .inactive }) {
+            existingTransceiver.sender.track = track
+            log("Reusing existing transceiver (out of \(_pc.transceivers.count)) for \(track.kind) track", .debug)
+            return existingTransceiver
+        } else if let newTransceiver = _pc.addTransceiver(with: track, init: transceiverInit) {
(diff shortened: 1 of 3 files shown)


pblazej:
The only failure I encountered is this one has no effect on `.inactive` transceivers, while I cannot do `transInit.direction = .sendOnly` (again).

```swift
if let preferredCodec = publishOptions.preferredCodec {
  try transceiver.set(preferredVideoCodec: preferredCodec)
}
```

This is not tragically bad tbh.

Sources/LiveKit/Core/Transport.swift
```diff
@@ -315,11 +315,16 @@ extension Transport {
     func addTransceiver(with track: LKRTCMediaStreamTrack,
                         transceiverInit: LKRTCRtpTransceiverInit) throws -> LKRTCRtpTransceiver
     {
-        guard let transceiver = _pc.addTransceiver(with: track, init: transceiverInit) else {
+        if let existingTransceiver = _pc.transceivers.first(where: { $0.mediaType == .init(trackKind: track.kind) && $0.direction == .inactive }) {
+            existingTransceiver.sender.track = track
```
boks1971, line 319: Is `existingTransceiver` guaranteed to have a `sender`? If there is no `sender`, does this crash? (sorry, I do not know Swift, so dumb questions :-) )

cloudwebrtc: yeah, maybe some transceivers with `.recvonly => .inactive` will have their `transceiver.sender` empty ?

boks1971: True @cloudwebrtc . In LK use case it won't be as we use publisher and subscriber peer connection separately. So, all publisher peer connection m-lines would have been either `sendrecv` or `sendonly` at some point and would have become `inactive` after unpublish.

If we do switch to single peer connection, there might be m-lines used for subscribed tracks which could go from `recvonly` -> `inactive`.

pblazej:
I'm more into making this one work https://github.com/webrtc-sdk/webrtc/pull/194 - no architectural changes

pblazej:
Closing in favor of #775

PR #774, opened 2025-08-31
Audio capturing for manual rendering mode

hiroshihorie:


hiroshihorie:
We need to test if I didn't break anything here 🙂

pblazej:
Vibe-coded a quick prototype according to README https://github.com/livekit-examples/swift-example/pull/77

It works well with one mode, but switching between manual/mic gives me either `Engine is not running` or a crash (on iOS). What am I doing wrong?

BTW maybe it's worth converting into a "proper panel" for future testing - feel free to destroy it 👍

hiroshihorie:
We don't really support switching manual / device rendering mode at runtime, it's a complicated internal switch... although ideally it should work.

pblazej:
> We don't really support switching manual / device rendering mode at runtime, it's a complicated internal switch... although ideally it should work.

Maybe `assertionFailure` would be better in this case?

pblazej:
BTW, are you gonna integrate https://github.com/livekit/webrtc-xcframework/releases/tag/137.7151.04?

hiroshihorie:
Yes I think the options are:
1. Make it settable once only like ADM type.
2. Simply fix the AVAudioEngine ADM to be able to switch at run time. I think the errors you see are from the attached nodes in MixerEngineObserver.swift. I think the node attach/cleanup lifecycle is not working.

hiroshihorie:
@pblazej Yes I think 0.4 is already being used.

pblazej:
> Make it settable once only like ADM type.

I think this is perfectly fine for now.

pblazej:
Started getting `Room.signalClient(_:didSubscribeTrack:) Could not find local track publication for subscribed event` in manual mode, anything wrong with the example?

hiroshihorie:
This seems to fix run-time toggle: https://github.com/webrtc-sdk/webrtc/pull/199/commits/0d9f8a7530c09e07f8bc193d242c06c00e491374

pblazej:
@hiroshihorie great, I'll revisit that as it's updated.

hiroshihorie:
@pblazej lib updated, shouldn't crash when toggling modes now 🤞🙂

pblazej (approved):
Works like a charm now 🔈

PR #781, opened 2025-09-08
Encrypted data channel

pblazej:
- Adds data channel encryption based on https://github.com/livekit/client-sdk-js/pull/1595
- Adds `currentKey` state to `BaseKeyProvider`
  - It was missing comparing to other platforms, and we cannot explicitly default to `0`
- Adds some integration tests, as most of the _crypto_ stuff is tested by webrtc
  - Existing hi-level tests e.g. data streams will get a `sharedKey` by default 🎉 

### Migration/deprecations
- Replace `E2EEOptions` with `EncryptionOptions` (1:1)
- Add `EncryptionType` parameter to `Room/ParticipantDelegate` methods

### DIFF
--- a/Sources/LiveKit/Core/DataChannelPair.swift
+++ b/Sources/LiveKit/Core/DataChannelPair.swift
@@ -90,7 +90,7 @@ class DataChannelPair: NSObject, @unchecked Sendable, Loggable {
         func peek() -> PublishDataRequest? { queue.first }
 
         mutating func enqueue(_ request: PublishDataRequest) {
-            queue.append(request.retry())
+            queue.append(request.withoutContinuation())
             currentAmount += UInt64(request.data.data.count)
         }
 
@@ -114,7 +114,7 @@ class DataChannelPair: NSObject, @unchecked Sendable, Loggable {
         let sequence: UInt32
         let continuation: CheckedContinuation<Void, any Error>?
 
-        func retry() -> Self {
+        func withoutContinuation() -> Self {
             .init(data: data, sequence: sequence, continuation: nil)
         }
     }
@@ -241,6 +241,7 @@ class DataChannelPair: NSObject, @unchecked Sendable, Loggable {
             log("Wrong packet sequence while retrying: \(first.sequence) > \(lastSeq + 1), \(first.sequence - lastSeq - 1) packets missing", .warning)
         }
         while let request = buffer.dequeue() {
+            assert(request.continuation == nil, "Continuation may fire multiple times while retrying causing crash")
             if request.sequence > lastSeq {
                 let event = ChannelEvent(channelKind: .reliable, detail: .publishData(request))
                 _state.eventContinuation?.yield(event)
@@ -434,11 +435,11 @@ extension DataChannelPair: LKRTCDataChannelDelegate {
             }
         }
 
-        if let encrypted = dataPacket.decrypt,
+        if let encryptedPacket = dataPacket.encryptedPacketOrNil,
            let e2eeManager, e2eeManager.isDataChannelEncryptionEnabled
         {
             do {
-                let decryptedData = try e2eeManager.handle(encryptedData: encrypted.toRTCEncryptedPacket(), participantIdentity: dataPacket.participantIdentity)
+                let decryptedData = try e2eeManager.handle(encryptedData: encryptedPacket.toRTCEncryptedPacket(), participantIdentity: dataPacket.participantIdentity)
                 let decryptedPayload = try Livekit_EncryptedPacketPayload(serializedBytes: decryptedData)
 
                 var dataPacket = dataPacket
--- a/Sources/LiveKit/Core/DataChannelPair.swift
+++ b/Sources/LiveKit/Core/DataChannelPair.swift
@@ -23,6 +23,7 @@ internal import LiveKitWebRTC
 
 protocol DataChannelDelegate: Sendable {
     func dataChannel(_ dataChannelPair: DataChannelPair, didReceiveDataPacket dataPacket: Livekit_DataPacket)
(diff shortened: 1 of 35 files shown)


Sources/LiveKit/Core/Room.swift
```diff
@@ -339,15 +341,26 @@ public class Room: NSObject, @unchecked Sendable, ObservableObject, Loggable {
             _state.mutate { $0.connectOptions = connectOptions }
         }
 
+        await cleanUp()
```
pblazej, line 345: Moved due to `E2EEManager.cleanUp` called after `setup`

pblazej:
@lukasIO would be nice to perform some x-platform test (proto compatibility)

pblazej:
@lukasIO as for x-platform test:
- ratcheting from JS → native works fine after [fixing](https://github.com/livekit/client-sdk-swift/pull/781/commits/d569de4aeccea5fca7c2a58e972b00a1f521715c) default window size
```
🆔 CHAT:  Hello new participant participant-qwfdzf1mmbr. This is just an progressively updating chat message from me, participant participant-oghni0gl54h.
🆔 CHAT:  rat1
🆔 CHAT:  rat2
🆔 CHAT:  rat3
```
- ratcheting from the native client → JS does not work - tried to listen for `KeyProviderEvent.RatchetRequest` but nothing...

image


cc @cloudwebrtc maybe it's worth to do another x-check with flutter

lukasIO:
> ratcheting from the native client → JS does not work 

does audio/video keep working when ratcheting on the native client ?

pblazej:
> > ratcheting from the native client → JS does not work
> 
> does audio/video keep working when ratcheting on the native client ?

nope, keys get out of sync.

```
Error encrypting track data: InvalidKey: Decryption failed: The operation failed for an operation-specific reason 
```

pblazej:
@lukasIO if you wanna dig deeper, I added one option to the example (2.8.0). Seems like the frame logic may be broken, unrelated to this PR (I haven't tested `main` tho).

Screenshot 2025-09-12 at 8 22 38 AM

BTW, macOS chat is still using old infra, but you should be able to get there with frame encryption. I can help you real-time if needed.

For me:
- if JS ratchets, native is able to decrypt, JS is unable
- if native ratchets, both are unable

so looks like the issue is on JS side (custom logic vs rtc)?

pblazej:
Looks like the above behavior is by [design](https://github.com/livekit/client-sdk-js/blob/239448f1fb7790a95a1ef69bb9a899850baeaa40/src/e2ee/KeyProvider.ts#L83-L85), for the shared key it should still be: `public let defaultRatchetWindowSize: Int32 = 0` so no auto-ratcheting.

pblazej:
Ended up reverting that [extended ratchet window] ⬆️

Sources/LiveKit/Core/DataChannelPair.swift
```diff
@@ -413,8 +435,29 @@ extension DataChannelPair: LKRTCDataChannelDelegate {
             }
         }
 
-        delegates.notify {
-            $0.dataChannel(self, didReceiveDataPacket: dataPacket)
+        if let encryptedPacket = dataPacket.encryptedPacketOrNil,
+           let e2eeManager, e2eeManager.isDataChannelEncryptionEnabled
```
lukasIO: decrypting should also work when data channel _encryption_ is not explicitly enabled.

pblazej: thought about that, yep it makes more sense 👍

pblazej: 6070ec26e2b117e61bcf37f700a2b781e473b9f1

It does not apply to situations where e2ee is explicitly disabled (no options passed).

lukasIO: nice!

Sources/LiveKit/Protocols/RoomDelegate.swift
```diff
@@ -151,7 +151,11 @@ public protocol RoomDelegate: AnyObject, Sendable {
 
     /// Received data from from a user or server. `participant` will be nil if broadcasted from server.
     @objc optional
-    func room(_ room: Room, participant: RemoteParticipant?, didReceiveData data: Data, forTopic topic: String)
+    func room(_ room: Room, participant: RemoteParticipant?, didReceiveData data: Data, forTopic topic: String, encryptionType: EncryptionType)
+
+    /// Failed to decrypt a data packet when encryption is enabled.
+    @objc optional
+    func room(_ room: Room, didFailToDecryptDataWithEror error: LiveKitError)
```
pblazej, line 158: Maybe it would be good to pull participant id here?

pblazej:
@hiroshihorie feel free to take a look at the Swift side, I don't see any obvious improvements atm.

Sources/LiveKit/Core/DataChannelPair.swift
```diff
@@ -34,6 +35,8 @@ class DataChannelPair: NSObject, @unchecked Sendable, Loggable {
 
     var isOpen: Bool { _state.isOpen }
 
+    var e2eeManager: E2EEManager?
```
hiroshihorie, line 38: Maybe this can be `weak` ? I think the `Room` holds the instance. 🤔

hiroshihorie (approved):

Sources/LiveKit/Core/DataChannelPair.swift
pblazej: I'd be careful with that - I mean, there's no point in making sure that there's _only_ 1 strong ref... if there's no cycle here... Let's double-check if the DC is deallocated properly (if needed).

Sources/LiveKit/Protos/livekit_metrics.pb.swift
pblazej, line 1: We gotta move'm to submodules finally, it's hard to track these subtle changes.

pblazej:
Merging as macOS build must be fixed on `main` anyway.

PR #782, opened 2025-09-11
Minor validation logic improvements

hiroshihorie:
Throws different type of error if server validation failed


pblazej:
No more flaky tests? 😄

Hm, looks like this is still related? Is it just ACK (202)?

```
testParticipantCleanUp, failed: caught error: "Error Domain=io.livekit.swift-sdk Code=202 "Network error(Unknown)" UserInfo={NSLocalizedDescription=Network error(Unknown)}"
```

Sources/LiveKit/Core/SignalClient.swift
```diff
@@ -170,29 +170,34 @@ actor SignalClient: Loggable {
             // Skip validation if user cancelled
             if error is CancellationError {
                 await cleanUp(withError: error)
-                throw error
+                throw LiveKitError(.cancelled, internalError: error)
```
pblazej: Without typed throws (6.0) it's hard to tell globally, but I'm not sure if wrapping `CancellationError` is a good idea.

Sources/LiveKit/Support/HTTP.swift
```diff
@@ -16,27 +16,51 @@
 
 import Foundation
 
+enum ServerValidationResponse {
+    case valid
+    case invalid(message: String)
+    // Network error etc.
+    case unknown(error: Error)
```
pblazej: IMO either `throws`/`Result`/`Error` where `throws` should be prioritized

Sources/LiveKit/Core/SignalClient.swift
hiroshihorie: ooh yes this isn't good

pblazej:
Let's unblock this PR, 2 nitpicks:
- `ServerValidationResponse` should be valid/invalid + the method `throws`
- let's maybe revert this line `throw LiveKitError(.cancelled, internalError: error)`

Othwerwise, LGTM 🌮

Sources/LiveKit/Errors.swift
```diff
@@ -30,6 +30,7 @@ public enum LiveKitErrorType: Int, Sendable {
     case webRTC = 201
 
     case network // Network issue
+    case validation // Network issue
```
pblazej: Nit: I'd rename that to `authorization` (or better?), as this is very generic (not sure about other SDKs).

Otherwise LGTM^2

hiroshihorie: I think we used to call this "validation", the endpoint (path is /validate) https://github.com/livekit/client-sdk-swift/blob/fc43f1e39db57ad012b9acc395d5251b4134f6cf/Sources/LiveKit/Support/Utils.swift#L170
So maybe validation is better ?

pblazej: If that matches backend, then why not

PR #784, opened 2025-09-15
Fix audio processing delegate lifecycle

hiroshihorie:


Sources/LiveKit/Audio/Manager/AudioManager.swift
```diff
@@ -117,28 +117,28 @@ public class AudioManager: Loggable {
...
 
     /// Add a delegate to modify the local audio buffer before it is sent to the network
     /// - Note: Only one delegate can be set at a time, but you can create one to wrap others if needed
     /// - Note: If you only need to observe the buffer (rather than modify it), use ``add(localAudioRenderer:)`` instead
-    public var capturePostProcessingDelegate: AudioCustomProcessingDelegate? {
-        get { capturePostProcessingDelegateAdapter.target }
-        set {
-            capturePostProcessingDelegateAdapter.set(target: newValue)
-            capturePostProcessingDelegateSubject.send(newValue)
+    public weak var capturePostProcessingDelegate: AudioCustomProcessingDelegate? {
```
xianshijing-lk: do we really need to set it weak ? or we make sure it will be set to nil when AudioRenderer get removed from the track ?

pblazej: I generally agree with this comment as posted elsewhere, for delegates it's the "default" approach, but for "top-down" ownership model (no cycles) we should use `weak` sparingly.

hiroshihorie: Either option works for me, as long as we’re okay with having inconsistencies related to the delegate weakness in the SDK, since anything using `MulticastDelegate` is weak now, e.g. `room.add(delegate:)`, `AudioManager.shared.add(localAudioRenderer:)`, etc.

pblazej: Unfortunately I don't see any option to "automate" it e.g. via swiftlint as every use case is vastly different and must be evaluated separately 😿

hiroshihorie:
Hmm... looks like something broke with this PR 😢

PR #787, opened 2025-09-16
Credential providers

pblazej:
Incarnation of https://github.com/livekit/client-sdk-js/pull/1645

Key differences:
- Mostly leverages Swift-native approach (protocols + default impl)
- Token `fetch` is stateless, caching is opt-in wrapper (which could be extended to Keychain tho)
- JWT support is limited e.g. exposing cached token payload
  - we have a 3rd party dependency, but only used for testing - IMO it's not worth to compile it for every single use case, but I'm happy to include it based on feedback

### DIFF
--- a/Sources/LiveKit/Auth/ConnectionCredentials.swift
+++ b/Sources/LiveKit/Auth/ConnectionCredentials.swift
@@ -1,0 +1,160 @@
+/*
+ * Copyright 2025 LiveKit
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import Foundation
+
+public enum ConnectionCredentials {
+    public struct Request: Encodable, Equatable, Sendable {
+        let roomName: String? = nil
+        let participantName: String? = nil
+        let participantIdentity: String? = nil
+        let participantMetadata: String? = nil
+        let participantAttributes: [String: String]? = nil
+//        let roomConfiguration: RoomConfiguration? = nil
+    }
+
+    public struct Response: Decodable, Sendable {
+        let serverUrl: URL
+        let participantToken: String
+    }
+
+    public typealias Literal = Response
+}
+
+// MARK: - Provider
+
+public protocol CredentialsProvider: Sendable {
+    func fetch(_ request: ConnectionCredentials.Request) async throws -> ConnectionCredentials.Response
+}
+
+// MARK: - Implementation
+
+extension ConnectionCredentials.Literal: CredentialsProvider {
+    public func fetch(_: ConnectionCredentials.Request) async throws -> ConnectionCredentials.Response {
+        self
+    }
+}
+
+public struct SandboxTokenServer: CredentialsProvider, Loggable {
+    private static let baseURL = URL(string: "https://cloud-api.livekit.io")!
+
+    public struct Options: Sendable {
+        let id: String
+        let baseURL: URL? = nil
+    }
+
+    private let options: Options
+
+    public init(options: Options) {
+        self.options = options
+    }
+
+    public init(id: String) {
+        options = .init(id: id)
+    }
+
+    public func fetch(_ request: ConnectionCredentials.Request) async throws -> ConnectionCredentials.Response {
+        log("Using sandbox token server is not applicable for production environemnt", .info)
+
+        let baseURL = options.baseURL ?? Self.baseURL
+        var urlRequest = URLRequest(url: baseURL.appendingPathComponent("api/sandbox/connection-details"))
+
+        urlRequest.httpMethod = "POST"
+        urlRequest.addValue(options.id.trimmingCharacters(in: CharacterSet(charactersIn: "\"")), forHTTPHeaderField: "X-Sandbox-ID")
+        urlRequest.httpBody = try JSONEncoder().encode(request)
+
+        let (data, response) = try await URLSession.shared.data(for: urlRequest)
+
+        guard let httpResponse = response as? HTTPURLResponse else {
+            throw LiveKitError(.network, message: "Error generating token from sandbox token server, no response")
+        }
+
+        guard (200 ... 299).contains(httpResponse.statusCode) else {
+            throw LiveKitError(.network, message: "Error generating token from sandbox token server, received \(httpResponse)")
+        }
+
+        return try JSONDecoder().decode(ConnectionCredentials.Response.self, from: data)
+    }
+}
+
+// MARK: - Cache
+
+public actor CachingCredentialsProvider: CredentialsProvider, Loggable {
+    private let provider: CredentialsProvider
+    private let validator: (ConnectionCredentials.Request, ConnectionCredentials.Response) -> Bool
+
+    private var cached: (ConnectionCredentials.Request, ConnectionCredentials.Response)?
+
+    public init(_ provider: CredentialsProvider, validator: @escaping (ConnectionCredentials.Request, ConnectionCredentials.Response) -> Bool = { _, res in res.hasValidToken() }) {
+        self.provider = provider
+        self.validator = validator
+    }
+
+    public func fetch(_ request: ConnectionCredentials.Request) async throws -> ConnectionCredentials.Response {
+        if let (cachedRequest, cachedResponse) = cached, cachedRequest == request, validator(cachedRequest, cachedResponse) {
+            log("Using cached credentials", .debug)
+            return cachedResponse
+        }
+
+        let response = try await provider.fetch(request)
+        cached = (request, response)
+        return response
+    }
+
+    public func invalidate() {
+        cached = nil
+    }
+}
+
+// MARK: - Validation
+
+public extension ConnectionCredentials.Response {
+    func hasValidToken(withTolerance tolerance: TimeInterval = 60) -> Bool {
+        let parts = participantToken.components(separatedBy: ".")
+        guard parts.count == 3 else {
+            return false
+        }
+
+        let payloadData = parts[1]
+
+        struct JWTPayload: Decodable {
+            let exp: Double
+        }
+
+        guard let payloadJSON = payloadData.base64URLDecode(),
+              let payload = try? JSONDecoder().decode(JWTPayload.self, from: payloadJSON)
+        else {
+            return false
+        }
+
+        let now = Date().timeIntervalSince1970
+        return payload.exp > now - tolerance
+    }
+}
+
+private extension String {
+    func base64URLDecode() -> Data? {
+        var base64 = self
+        base64 = base64.replacingOccurrences(of: "-", with: "+")
+        base64 = base64.replacingOccurrences(of: "_", with: "/")
+
+        while base64.count % 4 != 0 {
+            base64.append("=")
+        }
+
+        return Data(base64Encoded: base64)
+    }
+}
--- a/Sources/LiveKit/Auth/ConnectionCredentials.swift
+++ b/Sources/LiveKit/Auth/ConnectionCredentials.swift
@@ -146,7 +146,7 @@ public struct SandboxTokenServer: TokenServer {
 
 // MARK: - Cache
 
-/// `CachingCredentialsProvider` handles in-memory caching of credentials from any other `CredentialsProvider`.
+/// `CachingCredentialsProvider` handles caching of credentials from any other `CredentialsProvider` using configurable storage.
 public actor CachingCredentialsProvider: CredentialsProvider, Loggable {
     /// A tuple containing the request and response that were cached.
     public typealias Cached = (ConnectionCredentials.Request, ConnectionCredentials.Response)
@@ -155,31 +155,74 @@ public actor CachingCredentialsProvider: CredentialsProvider, Loggable {
 
     private let provider: CredentialsProvider
     private let validator: Validator
-
-    private var cached: Cached?
+    private let storage: CredentialsStorage
 
     /// Initialize a caching wrapper around any credentials provider.
     /// - Parameters:
     ///   - provider: The underlying credentials provider to wrap
+    ///   - storage: The storage implementation to use for caching (defaults to in-memory storage)
     ///   - validator: A closure to determine if cached credentials are still valid (defaults to JWT expiration check)
-    public init(_ provider: CredentialsProvider, validator: @escaping Validator = { _, res in res.hasValidToken() }) {
+    public init(
+        _ provider: CredentialsProvider,
+        storage: CredentialsStorage = InMemoryCredentialsStorage(),
+        validator: @escaping Validator = { _, res in res.hasValidToken() }
+    ) {
         self.provider = provider
+        self.storage = storage
         self.validator = validator
     }
 
     public func fetch(_ request: ConnectionCredentials.Request) async throws -> ConnectionCredentials.Response {
-        if let (cachedRequest, cachedResponse) = cached, cachedRequest == request, validator(cachedRequest, cachedResponse) {
+        if let (cachedRequest, cachedResponse) = await storage.retrieve(),
+           cachedRequest == request,
+           validator(cachedRequest, cachedResponse)
+        {
             log("Using cached credentials", .debug)
             return cachedResponse
         }
 
         let response = try await provider.fetch(request)
-        cached = (request, response)
+        try await storage.store((request, response))
         return response
     }
 
     /// Invalidate the cached credentials, forcing a fresh fetch on the next request.
-    public func invalidate() {
+    public func invalidate() async {
+        await storage.clear()
+    }
+}
+
+// MARK: - Storage
+
+/// Protocol for abstract storage that can persist and retrieve a single cached credential pair.
+/// Implement this protocol to create custom storage implementations e.g. for Keychain.
+public protocol CredentialsStorage: Sendable {
+    /// Store credentials in the storage (replaces any existing credentials)
+    func store(_ credentials: CachingCredentialsProvider.Cached) async throws
+
+    /// Retrieve the cached credentials
+    /// - Returns: The cached credentials if found, nil otherwise
+    func retrieve() async -> CachingCredentialsProvider.Cached?
+
+    /// Clear the stored credentials
+    func clear() async
+}
+
+/// Simple in-memory storage implementation
+public actor InMemoryCredentialsStorage: CredentialsStorage {
+    private var cached: CachingCredentialsProvider.Cached?
+
+    public init() {}
+
+    public func store(_ credentials: CachingCredentialsProvider.Cached) async throws {
+        cached = credentials
+    }
+
+    public func retrieve() async -> CachingCredentialsProvider.Cached? {
+        cached
+    }
+
+    public func clear() async {
         cached = nil
     }
 }
--- a/Sources/LiveKit/Auth/ConnectionCredentials.swift
+++ b/Sources/LiveKit/Auth/ConnectionCredentials.swift
@@ -40,40 +40,31 @@ public protocol CredentialsProvider: Sendable {
     func fetch(_ request: ConnectionCredentials.Request) async throws -> ConnectionCredentials.Response
 }
 
-// MARK: - Implementation
-
 extension ConnectionCredentials.Literal: CredentialsProvider {
     public func fetch(_: ConnectionCredentials.Request) async throws -> ConnectionCredentials.Response {
         self
     }
 }
 
-public struct SandboxTokenServer: CredentialsProvider, Loggable {
-    private static let baseURL = URL(string: "https://cloud-api.livekit.io")!
+// MARK: - Token Server
 
-    public struct Options: Sendable {
-        let id: String
-        let baseURL: URL? = nil
-    }
+public protocol TokenServer: CredentialsProvider {
+    var url: URL { get }
+    var method: String { get }
+    var headers: [String: String] { get }
+}
 
-    private let options: Options
+public extension TokenServer {
+    var method: String { "POST" }
+    var headers: [String: String] { [:] }
 
-    public init(options: Options) {
-        self.options = options
-    }
+    func fetch(_ request: ConnectionCredentials.Request) async throws -> ConnectionCredentials.Response {
+        var urlRequest = URLRequest(url: url)
 
-    public init(id: String) {
-        options = .init(id: id)
-    }
-
-    public func fetch(_ request: ConnectionCredentials.Request) async throws -> ConnectionCredentials.Response {
-        log("Using sandbox token server is not applicable for production environemnt", .info)
-
-        let baseURL = options.baseURL ?? Self.baseURL
-        var urlRequest = URLRequest(url: baseURL.appendingPathComponent("api/sandbox/connection-details"))
-
-        urlRequest.httpMethod = "POST"
-        urlRequest.addValue(options.id.trimmingCharacters(in: CharacterSet(charactersIn: "\"")), forHTTPHeaderField: "X-Sandbox-ID")
+        urlRequest.httpMethod = method
+        for (key, value) in headers {
+            urlRequest.addValue(value, forHTTPHeaderField: key)
+        }
         urlRequest.httpBody = try JSONEncoder().encode(request)
 
         let (data, response) = try await URLSession.shared.data(for: urlRequest)
@@ -87,6 +78,19 @@ public struct SandboxTokenServer: CredentialsProvider, Loggable {
         }
 
         return try JSONDecoder().decode(ConnectionCredentials.Response.self, from: data)
+    }
+}
+
+public struct SandboxTokenServer: TokenServer {
+    public let url = URL(string: "https://cloud-api.livekit.io/api/sandbox/connection-details")!
+    public var headers: [String: String] {
+        ["X-Sandbox-ID": id.trimmingCharacters(in: CharacterSet(charactersIn: "\""))]
+    }
+
+    public let id: String
+
+    public init(id: String) {
+        self.id = id
     }
 }
 
--- a/Sources/LiveKit/Auth/ConnectionCredentials.swift
+++ b/Sources/LiveKit/Auth/ConnectionCredentials.swift
@@ -17,20 +17,28 @@
 import Foundation
 
 public enum ConnectionCredentials {
-    public struct Request: Encodable, Equatable, Sendable {
+    public struct Request: Encodable, Sendable, Equatable {
         let roomName: String?
         let participantName: String?
         let participantIdentity: String?
         let participantMetadata: String?
         let participantAttributes: [String: String]?
-//        let roomConfiguration: RoomConfiguration?
+        let roomConfiguration: RoomConfiguration?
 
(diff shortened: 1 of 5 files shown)


Sources/LiveKit/Types/RoomConfiguration.swift
pblazej, line 1: I haven't pushed the respective proto, it does not provide any kind of "validation" for the payload keys, etc.

Sources/LiveKit/Types/RoomConfiguration.swift
```diff
@@ -0,0 +1,102 @@
...
+        case departureTimeout = "departure_timeout"
+        case maxParticipants = "max_participants"
+        case metadata
+        case minPlayoutDelay = "min_playout_delay"
+        case maxPlayoutDelay = "max_playout_delay"
+        case syncStreams = "sync_streams"
+        case agents
+    }
+
+    public init(
```
pblazej, line 62: It may be good to provide (later) a simpler overload with `agentName` for the purpose of agent abstractions.

Sources/LiveKit/Core/Room.swift
```diff
@@ -408,6 +408,15 @@ public class Room: NSObject, @unchecked Sendable, ObservableObject, Loggable {
...
     }
 
+    public func connect(credentialsProvider: CredentialsProvider,
+                        credentialsOptions: ConnectionCredentials.Options = .init(),
+                        connectOptions: ConnectOptions? = nil,
+                        roomOptions: RoomOptions? = nil) async throws
+    {
+        let credentials = try await credentialsProvider.fetch(credentialsOptions)
+        try await connect(url: credentials.serverUrl.absoluteString, token: credentials.participantToken, connectOptions: connectOptions, roomOptions: roomOptions)
+    }
```
1egoman: A few notes:

- At least in the javascript implementation, I had started in a similar direction as what it looks like you are doing here (passing "options" alongside `connect`) but @lukasIO had objected and  preferred that the request parameters be set directly on the `ConnectionCredentials` with a method. I'll let him rearticulate his exact argument but I think it was in his mind a cleaner separation of concerns. It's maybe something to consider, idk.

- Just like I did in the javascript version, It might be a good idea to cache the `credentialsProvider` instance on `room` in the `connect` call and then in `disconnect` call `self.credentialsProvider.fetch(...)` to regenerate the token if required so that reconnecting after a disconnect could be a little faster since the token would already be valid.

- Github won't let me select the old `connect` function signature, but is there some way it can officially be deprecated in swift in a similar way as what I did in typescript? Given there's also a `ConnectionCredentials.Literal` it seems like there's not really an explicit use case for the old signature anymore?

Sources/LiveKit/Auth/ConnectionCredentials.swift
```diff
@@ -0,0 +1,269 @@
...
+        /// The WebSocket URL for the LiveKit server.
+        let serverUrl: URL
+        /// The JWT token containing participant permissions and metadata.
+        let participantToken: String
+
+        public init(serverUrl: URL, participantToken: String) {
+            self.serverUrl = serverUrl
+            self.participantToken = participantToken
+        }
+    }
```
1egoman: I know you had said you didn't want to link in a jwt library in your main pull request comment. That being said, given you already have the jwt decoding logic implemented in `hasValidToken`, it might be good to expose some of the same additional fields with some getters as I did in the web version - [see here](https://github.com/livekit/client-sdk-js/pull/1645#discussion_r2349197825). Or even just a method / getter to access the raw jwt payload I think would be sufficient, just so those additional (potentially autogenerated) field values are actually accessible somehow.

Sources/LiveKit/Auth/ConnectionCredentials.swift
```diff
@@ -0,0 +1,269 @@
...
+    /// The URL endpoint for token generation.
+    var url: URL { get }
+    /// The HTTP method to use (defaults to "POST").
+    var method: String { get }
+    /// Additional HTTP headers to include with the request.
+    var headers: [String: String] { get }
+}
+
+public extension TokenServer {
+    var method: String { "POST" }
```
1egoman: Something like this might be worth doing on the web too - Lukas had been referring to this as `ConnectionCredentials.Standard` in some of our discussions. I realize it's a protocol and not a real object definition, but I guess what I'm wondering is if maybe it's worth upgrading it given ben's vision of "one request / response standard that can be used everywhere".

pblazej: > worth upgrading

Then you need inheritance to define custom URL (with "empty" URL in the base class).

Now, a minimal server is not harder than that (but avoiding this problem):
```swift
actor MyTokenServer: TokenServer {
    var url: URL = ...
}
```

Sources/LiveKit/Auth/ConnectionCredentials.swift
pblazej: Yeah, I'm aware you did it, I think exposing the cached response is enough (without decoding - which could be a wrapper).

Sources/LiveKit/Core/Room.swift
pblazej: - Yes, passing that is a little awkward here - but the mutable version is equally awkward to me - what if you don't set a `request`/options at all and `fetch`?
- Then the caching part makes less sense to me - if you invalidate on every disconnect, am I right?
- Yes, the deprecation is perfectly possible, buuuut it will appear on the ObjC side as well - and the new method essentially breaks ObjC compat (as `AgentSession`) does

Sources/LiveKit/Auth/ConnectionCredentials.swift
pblazej: 709949fb90bc7c2a3b84a8d9054c3bbb24b24f08

Sources/LiveKit/Core/Room.swift
pblazej: @lukasIO making everything stateful by default also creates some challenges, e.g. breaks the composition (who should cache the request - the cache or the underlying provider), I cannot force `Sandbox` to be initialized with both options and ID, etc.

Essentially, it fits one use case - when you have one concrete (caching) provider and switch the lambda under the hood.

IMO, that's the same problem that exists in `Room` - you can pass options on both `connect()` and `init()`.

I think the top-level question we need to answer is how and when to pass these token-related options here:
```swift
private let session = AgentSession(credentials: CachingCredentialsProvider(SandboxTokenServer(id: Self.sandboxId)),
                                       // agentName: ...
```
My initial concept was as above ⬆️ then we create credential "options" under the hood and pass to the room on connect. The session reuses `agentName(s)` for its internal filtering.

1egoman: > Yes, passing that is a little awkward here - but the mutable version is equally awkward to me - what if you don't set a request/options at all and fetch?

I'd think in that case it would default to an empty request? That's what the typescript implementation does at least.

> Then the caching part makes less sense to me - if you invalidate on every disconnect, am I right?

No, it shouldn't - it should refetch the token only if it is expired, and effectively be a no-op if it isn't expired. Though since the caching logic is opt in maybe the tradeoffs differ a bit and it's a less good idea if there isn't caching in the picture.

> Yes, the deprecation is perfectly possible, buuuut it will appear on the ObjC side as well - and the new method essentially breaks ObjC compat (as AgentSession) does

Hmm, I see. Sounds like there's some nuance / not a clear thing to do there...

Sources/LiveKit/Auth/ConnectionCredentials.swift
1egoman: Ah, I see. Yea that seems decent then, good point.

pblazej:
There will be some renaming to `TokenSource` and `TokenSource.Endpoint` as discussed with JS.

Sources/LiveKit/Auth/Sandbox.swift
pblazej, line 1: Moved it to a separate file for visibility.

Sources/LiveKit/Auth/ConnectionCredentials.swift
pblazej: Full JWT support b89ab16ff86c065e39ac38f24aba0090442d7dd2

pblazej:
Added JWT support (if we really want it 👍)

Sources/LiveKit/Auth/TokenSource.swift
```diff
@@ -0,0 +1,271 @@
...
+            case participantToken
+        }
+
+        public init(serverURL: URL, participantToken: String) {
+            self.serverURL = serverURL
+            self.participantToken = participantToken
+        }
+    }
+
+    public typealias Options = Request
```
bcherry: i'm not sure i understand the point of this alias

Sources/LiveKit/Auth/TokenSource.swift
```diff
@@ -0,0 +1,271 @@
...
+    public typealias Literal = Response
```
bcherry: why not make `Token` itself the literal type so it can just be `Token(...)`?

Sources/LiveKit/Auth/TokenSource.swift
```diff
@@ -0,0 +1,271 @@
...
+    /// - Returns: A token response containing the server URL and participant token
+    /// - Throws: An error if the token generation fails
+    func fetch(_ request: Token.Request) async throws -> Token.Response
+}
+
+/// `Token.Literal` contains a single set of credentials, hard-coded or acquired from a static source.
+extension Token.Literal: TokenSource {
+    public func fetch(_: Token.Request) async throws -> Token.Response {
+        self
+    }
```
bcherry: doing this as a protocol is pretty clever, but isn't a little messy that you end up with literal tokens implementing a fetch method and then discarding the request? it seems abstraction-breaking. if for some reason you sent a request to a literal token which doesn't match (e.g. literal token is for identity "foo" but request is for identity "bar") it would make more sense to throw an error than to return self.

Sources/LiveKit/Auth/Sandbox.swift
```diff
@@ -0,0 +1,35 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import Foundation
+
+/// `Sandbox` queries LiveKit Sandbox token server for credentials,
+/// which supports quick prototyping/getting started types of use cases.
+/// - Warning: This token endpoint is **INSECURE** and should **NOT** be used in production.
```
bcherry: add the url https://cloud.livekit.io/projects/p_/sandbox/templates/token-server

Sources/LiveKit/Auth/TokenSource.swift
```diff
@@ -0,0 +1,271 @@
...
+
+    public typealias Options = Request
+    public typealias Literal = Response
+}
+
+// MARK: - Source
+
+/// Protocol for types that can provide connection credentials.
+/// Implement this protocol to create custom credential providers (e.g., fetching from your backend API).
+public protocol TokenSource: Sendable {
```
bcherry: i think we'll need to expand this docstring to explain the whole system, since it seem like the entrypoint where people will encounter tokens for the first time (on the params to the room connect method)

Sources/LiveKit/Auth/TokenSource.swift
pblazej: It's 100% to pass it to the new `room.connect` overload, as there's no equivalent of JS `setRequest` thing

Sources/LiveKit/Auth/TokenSource.swift
pblazej: - yes, some form of check would be good here
- any alternative approaches in mind (for the protocol thing)? An `enum` with associated providers is one option (`.literal(Response)`, `.endpoint(...)`), but it does not compose so well, not sure if sacrificing that for the sake of `Literal` handling is a good tradeoff - `Literal` is more "I know what I'm doing" kind of choice here

Sources/LiveKit/Auth/TokenSource.swift
pblazej: It's precisely because of this `extension Token.Literal: TokenSource` so you do `room.connect(Token.Literal)` (discussion below)

Sources/LiveKit/Auth/TokenSource.swift
pblazej: @lukasIO I think it resonates with your suggestion to make it stateful (as in JS). In general, it's hard not to discard `_` any `request` arg for literal if the whole concept is based on request-response. Embedding some default validation logic like "this is the same room" may be also pretty surprising and does not compose well.

I think the stateful service think is slightly better:
```swift
protocol TokenSource: Sendable {
    mutating func set(request: Token.Request)
    func fetch() async throws -> Token.Response
}

extension TokenSource {
    func set(request: Token.Request) {} // here the ignorance happens
}

extension Token.Literal: TokenSource {
    public func fetch() async throws -> Token.Response {
        self
    }
}
```

pblazej: No need for `Options` alias as well @bcherry ⬆️ The only thing I don't like is the non-frp nature of it, so that caching becomes less elegant - that was my main motivation.

1egoman: @pblazej In the web implementation, I solved this by adding another abstraction layer called a "refreshable token source" - `Literal` then implemented the regular kind and `Custom` / `SandboxTokenServer` / etc implemented the refreshable kind. Maybe that's something that could be done in swift as well.

Here's some pseudocode:
```tsx
abstract class TokenSource {
  abstract generate(): Promise<TokenSourceResponse>;
}

abstract class RefreshableTokenSource extends TokenSource {
  setRequest(request: TokenSourceRequest) {
    this.request = request;
  }

  abstract fetch(request: TokenSourceRequest): Promise<TokenSourceResponse>;

  async generate() {
    return this.fetch(this.request);
  }
  refresh() { /* (calls generate behind the scenes) */ }
}

// Literal only is a TokenSource
class LiteralTokenSource extends TokenSource {
  constructor(request) {
    this.request = request;
  }
 async generate() { return this.request; }
}

// Others are a RefreshableTokenSource
class OtherTokenSource extends RefreshableTokenSource {
  async fetch(request) {
    // compute response somehow
    return response;
  }
}
```

pblazej: Yes, you need 2 abstractions for that.

pblazej:
Not sure about the final form (stateless/stateful), depends on JS impl as well 👍

Sources/LiveKit/Core/Room.swift
xianshijing-lk: Found this interesting thread, and would like to put in some thoughts here. 

- regarding token refreshment:
Without much context, could any one helps me understand:
* what is the typical lifetime for our token ? 

For general use cases, I think I agree with Ryan that we should fetch the token only it is expired, and no need to invalidate the token in disconnect, as that is helpful to speed up the reconnect process.
And this is a pretty standard behavior, for instance, firebase authentication is using such approaches, and developers generally find it.

- regarding connect(), init() APIs
Any particular reason that we allow users to pass options in both init() and connect() ? 
If we can make the credentials/configuration immutable and centralized, like accept options on Session init() API, things will be cleaner and less confusing to the users ? 
For instance : 
let session = RoomSession(
    credentialsProvider: myCredentialsProvider,
    connectOptions: defaultConnectOptions,
    roomOptions: defaultRoomOptions
)

try await session.connect()

try avait session.reconnect() // not really sure if it is needed, maybe connect() will be sufficient ? 


- API deprecation / migration
How do we handle API changes ? do we have any existing process or guidelines for doing so ? 

I am almost sure that I will miss some important context, please feel free point them out.

Sources/LiveKit/Auth/Sandbox.swift
```diff
@@ -0,0 +1,35 @@
...
+public struct Sandbox: TokenEndpoint {
```
xianshijing-lk: Nit, is this like mostly for prototyping ? 

I am wondering if we should name it as LiveKitSandbox ? so that it will be more explicit. And we can use similar approaches for other prototype APIs.

Sources/LiveKit/Auth/TokenSource.swift
```diff
@@ -0,0 +1,287 @@
...
+            self.participantMetadata = participantMetadata
+            self.participantAttributes = participantAttributes
+            self.roomConfiguration = roomConfiguration
+        }
+    }
+
+    /// Response containing the credentials needed to connect to a room.
+    public struct Response: Decodable, Sendable {
+        /// The WebSocket URL for the LiveKit server.
+        public let serverURL: URL
```
xianshijing-lk: nit, should it be serverUrl ? what is the preferred naming style for url ?

Sources/LiveKit/Auth/TokenSource.swift
```diff
@@ -0,0 +1,287 @@
...
+
+import Foundation
+
+#warning("Fix camel case after deploying backend")
+
+// MARK: - Token
+
+/// `Token` represent the credentials needed for connecting to a new Room.
+/// - SeeAlso: [LiveKit's Authentication Documentation](https://docs.livekit.io/home/get-started/authentication/) for more information.
+public enum Token {
```
xianshijing-lk: curiously, any reason why we prefer public enum Token namespace style ? 

why not go flat like AuthRequest, AuthResult / AuthResponse ?

Sources/LiveKit/Auth/TokenSource.swift
```diff
@@ -0,0 +1,287 @@
...
+
+    public typealias Options = Request
+    public typealias Literal = Response
+}
+
+// MARK: - Source
+
+/// Protocol for types that can provide connection credentials.
+/// Implement this protocol to create custom credential providers (e.g., fetching from your backend API).
+public protocol TokenSource: Sendable {
```
xianshijing-lk: In general, source is vague, I think what you meant here is more like a provider.

If it is OK, I will suggest renaming it to TokenProvider or CredentialProvider

xianshijing-lk: In case the naming overlaps too much, like how other platforms use "source", then please disregard it. It's not worth the trouble.

Sources/LiveKit/Auth/TokenSource.swift
xianshijing-lk: Very good technical discussion. I don't think it matters that much, as it is more internal to us. 

In case things are not settled, will a optional field help ? 

For instance:

public protocol TokenSource: Sendable {
    func fetch(request: Token.Request?) async throws -> Token.Response
}

extension TokenSource {
    public func fetch(request: Token.Request? = nil) async throws -> Token.Response {
        try await fetch(request: request)
    }
}

extension Token.Literal: TokenSource {
    // Here we make it slightly more clear that it is accepting nil in this literal token case.
    public func fetch(request: Token.Request? = nil) async throws -> Token.Response {
        self
    }
}

Sources/LiveKit/Auth/TokenSource.swift
```diff
@@ -0,0 +1,287 @@
...
+        }
+
+        return try JSONDecoder().decode(Token.Response.self, from: data)
+    }
+}
+
+// MARK: - Cache
+
+/// `CachingTokenSource` handles caching of credentials from any other `TokenSource` using configurable store.
+public actor CachingTokenSource: TokenSource, Loggable {
```
xianshijing-lk: This looks like a TokenManager to me :)

xianshijing-lk (commented):
some thoughts from a quick glance, pretty solid code in general.

Sources/LiveKit/Auth/TokenSource.swift
```diff
@@ -0,0 +1,269 @@
...
+    public func cachedToken() async -> Token.Response? {
+        await store.retrieve()?.1
+    }
+}
+
+// MARK: - Store
+
+/// Protocol for abstract store that can persist and retrieve a single cached credential pair.
+/// Implement this protocol to create custom store implementations e.g. for Keychain.
+public protocol TokenStore: Sendable {
```
xianshijing-lk: Are we expecting the advanced users to extend this TokenStore protocol ? 

Any reason that a persistent store might not meet all the needs ? 

It might just be me, I wonder whyTokenStore is tied to one credential, will it be a valid use case that a token needs to be refreshed since a session life is long ? in that case, we might want to update a tokenStore rather than re-creating a new one ? (it might just make things a bit less error prone for the long term)

Sources/LiveKit/Auth/TokenSource.swift
pblazej: similar to https://developer.apple.com/documentation/foundation/urlsession

https://www.swift.org/documentation/api-design-guidelines/#conventions

Sources/LiveKit/Auth/TokenSource.swift
pblazej: I think I gonna move this finally 👍

Sources/LiveKit/Auth/TokenSource.swift
pblazej: We avoided the name `...Provider` per discussion with @lukasIO to keep it clear from React-based associations (provider is a patter used for DI there).

Sources/LiveKit/Auth/TokenSource.swift
pblazej: We can try to resolve naming in JS first to avoid back-and-forth 🥲

Sources/LiveKit/Auth/TokenSource.swift
pblazej: cc @lukasIO @1egoman

pblazej:
@xianshijing-lk thanks for all the great feedback, there's a slightly longer discussion on the JS PR already, it's hard to keep'em in sync: https://github.com/livekit/client-sdk-js/pull/1645

Sources/LiveKit/Auth/TokenSource.swift
lukasIO: Yeah, totally agree that `Provider` makes a lot of sense conceptually. 
Both @1egoman and me felt like `Provider` is unfortunately so overloaded in React as a concept that it would likely cause confusion. In order to provide naming consistency across client implementations we tried to come up with a name that is conceptually close to `Provider`, hence `TokenSource` was born.

Sources/LiveKit/Auth/TokenSource.swift
lukasIO: on an ongoing session tokens are automatically refreshed by the server and sent down to provide support for reconnects.

Sources/LiveKit/Auth/TokenSource.swift
1egoman: @xianshijing-lk Sorry it's hard to keep up with all the comments all over the place - we ended up introducing two new concepts on the js side called "TokenSourceFixed" and "TokenSourceConfigurable" which effectively codify this optionality in a way where certain methods / interfaces can limit which type they want to accept. But generally yes, what you are proposing here is generally what we decided, just a slightly different way to accomplish it!

Sources/LiveKit/Auth/TokenSource.swift
xianshijing-lk: Got it, that makes sense.  I do see the problems like "TokenProvider". lets keep the "Source" terminology as is. 

Just some technical context from some quick chatgpt (which aways explains things much better than me ):
Historical use of “provider” in auth

```
Long before React popularized the <Provider> pattern, “provider” was already the standard word for identity providers (Google, GitHub, Facebook, etc.).

Developers familiar with OAuth or OpenID usually expect terms like “auth provider” or “identity provider.”
```

So things make more sense for names like AuthProvider, or LiveKitAuthProvider, but TokenProvider might be more confusing to JS.

Sources/LiveKit/Auth/Sandbox.swift
```diff
@@ -0,0 +1,35 @@
...
+    public let url = URL(string: "https://cloud-api.livekit.io/api/sandbox/connection-details")!
```
davidliu: V2 token server api is now available at `/api/v2/sandbox/connection-details`.

Sources/LiveKit/Auth/TokenSource.swift
```diff
@@ -0,0 +1,269 @@
...
+        /// - SeeAlso: [Room Configuration Documentation](https://docs.livekit.io/home/get-started/authentication/#room-configuration) for more info.
+        public let roomConfiguration: RoomConfiguration?
+
+        enum CodingKeys: String, CodingKey {
+            case roomName = "room_name"
+            case participantName = "participant_name"
+            case participantIdentity = "participant_identity"
+            case participantMetadata = "participant_metadata"
+            case participantAttributes = "participant_attributes"
+            case roomConfiguration = "room_configuration"
```
davidliu: I think this should be roomConfig/room_config:

https://github.com/livekit/cloud-api-server/blob/b9804fc398beec24a335351bbaa19e5629eda54c/pkg/service/sandbox.go#L97

pblazej: Good point, I still need to move to protobufs for serialization - that's fixed there 👍

pblazej: I think I'll finally go for some equivalent of your `ProtoConverterTest` (that does not exist for Swift yet) vs exposing protobufs.

This needs a separate ticket as the approach is unclear (reflection, etc.)

Sources/LiveKit/Token/TokenSource.swift
```diff
@@ -0,0 +1,118 @@
...
+    public let participantAttributes: [String: String]?
+    /// Advanced room configuration options for token generation.
+    ///
+    /// Use this for advanced features like:
+    /// - Dispatching agents to the room
+    /// - Setting room limits and constraints
+    /// - Configuring recording or streaming options
+    ///
+    /// - SeeAlso: [Room Configuration Documentation](https://docs.livekit.io/home/get-started/authentication/#room-configuration) for more info.
+    public let roomConfiguration: RoomConfiguration?
```
pblazej: JS diff: skips `agentName/agentMetadata` as it's still nested inside `RoomConfiguration.agents` and can be injected from the upcoming session...

Sources/LiveKit/Token/TokenSource.swift
```diff
@@ -0,0 +1,118 @@
...
+///
+/// This protocol allows dynamic configuration of connection parameters, making it suitable for
+/// production applications that need flexible authentication and room management.
+///
+/// Common implementations:
+/// - ``SandboxTokenSource``: For testing with LiveKit Cloud sandbox [token server](https://cloud.livekit.io/projects/p_/sandbox/templates/token-server)
+/// - ``EndpointTokenSource``: For custom backend endpoints using LiveKit's JSON format
+/// - ``CachingTokenSource``: For caching credentials (or use the `.cached()` extension)
+public protocol TokenSourceConfigurable: Sendable {
+    func fetch(_ options: TokenRequestOptions) async throws -> TokenSourceResponse
```
pblazej, line 41: `TokenRequestOptions?` is also _an option_ here as mentioned above, similarly to `... connectOptions: ConnectOptions? = nil, roomOptions: RoomOptions? = nil`

lukasIO: I think at least for the sandbox token server the options would be required?

pblazej: Technically, I don't think so (but it obfuscates the intention)

image

pblazej:
@hiroshihorie @davidliu as we won't integrate that into the `Room` (could be `internal` as well) let's just make sure it's acceptable for agents and move on.

I think Android and Swift are almost identical atm, really easy to compare and follow, left some minor comments on Android.

xianshijing-lk:
Is this ready to land ?  Please let me know if you need help on the review or ping other reviewers

Thanks

pblazej:
Added `TokenSourceRequest` as suggested in the Android PR, will merge soon

Issue #177, opened 2023-01-20
Better video / audio devices enumeration API


hiroshihorie:
Currently there is no simple way to enumerate and specify Camera device on macOS.

pblazej:
This is a pretty old one, but I'd still love to have uniform query/obervation here:

```swift
        AudioManager.shared.onDeviceUpdate = { [weak self] _ in
            Task { @MainActor in
                self?.audioDevices = AudioManager.shared.inputDevices
                self?.selectedAudioDeviceID = AudioManager.shared.defaultInputDevice.deviceId
            }
        }

        Task {
            do {
                canSwitchCamera = try await CameraCapturer.canSwitchPosition()
                videoDevices = try await CameraCapturer.captureDevices()
                selectedVideoDeviceID = videoDevices.first?.uniqueID
            } catch {
                errorHandler(error)
            }
        }
```

Avoiding the use of `AVCaptureDevice` would be also cool (use `MediaDevice` wrapper instead)

pblazej:
@xianshijing-lk I think we should pull this out of the backlog (even creating new wrappers) as it would definitely improve the ergonomics, flutter (better) example: https://github.com/livekit/client-sdk-flutter/blob/main/lib/src/hardware/hardware.dart

pblazej:
I'll try to maybe sketch some new API (wrapper)

xianshijing-lk:
Sounds a good feature to implement

pradeepjena:
https://api.bettervideo.io

PR #789, opened 2025-09-18
Session API

pblazej:
Adds 3 basic building blocks for simple(r) agent experiences:
- `Session` - connection, pre-connect, agent dispatch, agent filtering (e.g. by name), all agents, messages (broadcasted and aggregated for now)
- `Agent` - wrapper around `Participant`, knows its tracks and internal state
- `LocalMedia` - (unrelated) helper to deal with local tracks in SwiftUI

Example: https://github.com/livekit-examples/agent-starter-swift/pull/29

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -61,11 +61,11 @@ jobs:
           # https://github.com/actions/runner-images/blob/main/images/macos/macos-26-arm64-Readme.md
           - os: macos-26
             xcode: latest
-            platform: "iOS Simulator,name=iPhone 17 Pro,OS=26.0"
+            platform: "iOS Simulator,name=iPhone 17 Pro,OS=26.1"
             symbol-graph: true
           - os: macos-26
             xcode: latest
-            platform: "iOS Simulator,name=iPhone 17 Pro,OS=26.0"
+            platform: "iOS Simulator,name=iPhone 17 Pro,OS=26.1"
             extension-api-only: true
           - os: macos-26
             xcode: latest
@@ -84,10 +84,10 @@ jobs:
             platform: "macOS,variant=Mac Catalyst"
           - os: macos-26
             xcode: latest
-            platform: "visionOS Simulator,name=Apple Vision Pro,OS=26.0"
+            platform: "visionOS Simulator,name=Apple Vision Pro,OS=26.1"
           - os: macos-26
             xcode: latest
-            platform: "tvOS Simulator,name=Apple TV,OS=26.0"
+            platform: "tvOS Simulator,name=Apple TV,OS=26.1"
 
     runs-on: ${{ matrix.os }}
     timeout-minutes: 60


Sources/LiveKit/Agent/Chat/Receive/TranscriptionStreamReceiver.swift
pblazej, line 1: Finally add some unit tests to that.

Sources/LiveKit/Agent/Session.swift
```diff
@@ -0,0 +1,199 @@
...
+             .connecting where isListening,
+             .connected,
+             .reconnecting:
+            true
+        default:
+            false
+        }
+    }
+
+    @Published public private(set) var agents: [Participant.Identity: Agent] = [:]
```
pblazej: I think `.ID` is still a better key than name itself.

Sources/LiveKit/Agent/Conversation.swift
```diff
@@ -0,0 +1,199 @@
...
+    private let senders: [any MessageSender]
+    private let receivers: [any MessageReceiver]
+
+    // MARK: - Internal state
+
+    private var waitForAgentTask: Task<Void, Swift.Error>?
+
+    // MARK: - Init
+
+    public init(credentials: CredentialsProvider, room: Room = .init(), agentName: String? = nil, senders: [any MessageSender]? = nil, receivers: [any MessageReceiver]? = nil) {
```
pblazej: @1egoman @lukasIO I think that's the discussion about the logic:
- `agentName` should take part in the direct dispatch
  - we'll introduce plural case later while keeping an internal array? it creates some confusion why the conversation cannot happen "with multiple agents"
- wait for agents can only happen at the conversation level (as the `Agent` will be published when joining)
  - I believe we should check the names vs who actually joined
  - name shouldn't take part in the filtering? so that I'll keep an agent that I formally did not pass?

1egoman: > we'll introduce plural case later while keeping an internal array? it creates some confusion why the conversation cannot happen "with multiple agents"

This was generally what I had in mind yes, start with a singular `agentName`-type parameter, and then in the future add an internal array which can be backed by a new `agentNames` parameter (In swift I'd think it could probably be a new overload? On web, the way to accomplish the same thing would be that the parameter would now accept either a `string` or `Array<string>`, and renamed since parameter names aren't part of the external interface in js)

This link may be useful and represents the initial state of this on the web: https://github.com/livekit/components-js/pull/1207/files#diff-c2401cb9c778162d5def12d137a663f03477b02c804d9372846c318e903df77bR125-R146

> wait for agents can only happen at the conversation level (as the Agent will be published when joining)

On the web right now, what I'm doing is in `useConversation` I'm calling `useAgent` to get access to the current agent, and then on the agent I exposed a method called `waitForAvailable` which returns a promise which resolves once `agent.isAvailable` is true. `await agent.waitForAvailable()` is then being called in `conversation.start()`.

Associated code here: https://github.com/livekit/components-js/pull/1207/files#diff-c2401cb9c778162d5def12d137a663f03477b02c804d9372846c318e903df77bR306-R346

In the future, what I had been thinking (and another use-case for conversation) is it could maintain a registry of all currently connected agents. I started going down the road of building that in javascript _sort of_ with `useAgentTimeoutStore` [here](https://github.com/livekit/components-js/pull/1207/files#diff-2a1ce2d680a89c3b24ff1cbf55fdc0569314e8bc04192ef7bdef99462799ed45R109-R120) (internal hook / not exposed). Right now it is largely responsible for agent timeout management but I could see that growing, storing data across multiple agents, and moving underneath conversation in the future.

pblazej: @1egoman I think like we're still scratching the surface with "absence" 🥲 

My high-level approach (not implemented yet) would be:
- Conversation controls the (global) timeout (as it does the dispatch), agent does not know about its own timeout, etc. 🟢 
- Currently, passing one `agentName` is a little misleading for people having multiple agents anyway (without direct dispatch)...
- When `agentName` is passed:
  - we do wait for this particular agent to join
    - should this wait be awaitable or background
- When multiple `agentNames` are passed:
  - do we wait for all? 
- When no `agentName` is passed:
  - we do wait for any agent 🟠 
  - we do not wait at all?
- `Agent` is added to the registry when it joins the room (agent lifecycle == participant lifecycle), regardless of its (conversational)state 🟠 theoretically we can add them with `"listening" | "thinking" | "speaking";`, maybe including `"idle"` but it also limits the space of states (if we don't register them - with the guarantee that they won't disappear after turning `idle`) or is a little confusing (if we register them with another concept of "availability"). IMO, consumers should learn what `AgentState` means for their use cases, it's hard to tell what "available" means universally.
- I'm not sure if doing that `await agent.waitUntilAvailable(signal);` is a great idea - should it be awaitable if you still wanna present some placeholder UI while agent joins/dispatches; this `await` can be modeled by agent's optionality `Agent?`/its tracks anyway 🟠 
- How to model `iCanSpeak` state - which is crucial for the UI:
  - During pre-connect: `capturingAudio + .disconnected, .connecting, .reconnecting, .connected`
  - After pre-connect: probably `connectionState` (room) is not enough, but how to handle e.g. handoff - I'd rather interpret that as "I'm in the room and someone is listening" rather than "some agent X is listening" as there may be gaps, etc.

The key difference is probably in how we think about "presence" - shall we make some arbitrary decisions here or no?

Sources/LiveKit/Agent/Session.swift
bcherry: yes because name isn't unique anyways - you can have many agents with the same agent_name and unique participant identities

bcherry (commented):
this generally looks good - lmk when the API is considered final

Sources/LiveKit/Agent/Session.swift
```diff
@@ -0,0 +1,269 @@
...
+                            senders: [any MessageSender]? = nil,
+                            receivers: [any MessageReceiver]? = nil)
+    {
+        self.init(tokenSourceConfiguration: .configurable(tokenSource, tokenOptions),
+                  options: options,
+                  senders: senders,
+                  receivers: receivers)
+    }
+
+    public convenience init(agentName: String,
```
pblazej: I think this is the best tradeoff with options, here `agentName` takes priority and makes the intent clear (?)

pblazej:
## JS diff

🟢 - good enough
🟠 - needs discussion
🔴 - complete mess

### Session Lifecycle 🟢 
- **Swift `Session.start`**: Handles token fetch, optional pre-connect audio, and `Room.connect`, exposing readiness via `isReady`/`isListening` that mirror `Room` state.
- **JS `useSession.start`**: Coordinates `Room.connect` through hooks, waits for transport plus `useAgent.waitUntilAvailable`, and surfaces async helpers (`waitUntilConnected`, `waitUntilDisconnected`) instead of stored readiness.
- **Connection Tracking**: Swift mirrors `Room.connectionState`; JS recomputes derived booleans each render and broadcasts via an internal `EventEmitter`.
- **Room Ownership**: Swift `Session` owns its `Room`; JS can take a provided/context `Room` or lazily create one.

### Agent Representation 🟠 
- **Multiplicity**: Swift keeps a dictionary of `Agent` instances keyed by `Participant.Identity`; JS resolves a single `agentParticipant` plus optional worker. 🔴 
- **State Propagation**: Swift `Agent` subscribes to `Participant.changes` and publishes `AgentState`, audio, and avatar tracks via `@Published`; JS derives state by combining participant attributes, local microphone state, and timeout info to yield fields like `isAvailable` and `isBufferingSpeech`.
- **Access Pattern**: Swift exposes `agent(named:)`, subscript access, and environment wrappers (`LiveKitAgent`); JS returns a memoized object with helpers (`waitUntilCamera`, `waitUntilMicrophone`) backed by React context.

### Messaging Pipeline 🟢 
- **Swift Protocol Layer**: Defines `MessageSender`/`MessageReceiver` with `AsyncStream`, defaulting to `TextMessageSender` loopback and `TranscriptionStreamReceiver` aggregating into an `OrderedDictionary` of `ReceivedMessage`.
- **JS Hook Composition**: `useSessionMessages` merges `useChat` and `useTranscriptions`, sorts merged `ReceivedMessage` arrays, and emits via an `EventEmitter`.
- **Send Path**: Swift `Session.send` multiplexes through configured senders and returns a `SentMessage`; JS delegates to `useChat.send`, returning a `ReceivedChatMessage` without centralized session error handling.
- **History**: Swift exposes `getMessageHistory`/`restoreMessageHistory`; JS recomputes combined arrays each render and leaves persistence to consumers.

### Timeout & Errors 🟠 
- **Agent Arrival**: Swift schedules `waitForAgentTask` after connect to set `.agentNotConnected`; JS uses `useAgentTimeoutIdStore` to surface `failureReasons` when an agent never joins or initializes. 🔴 
- **Error Surface**: Swift centralizes errors via `Session.Error` (`.failedToConnect`, `.failedToSend`); JS distributes through rejected promises, `SessionEvent` emitters, and agent state fields.
- **Reset**: Swift offers `resetError()`; JS relies on hook reinitialization and timeout resets.

### Media & Local Resources 🟢 
- **Swift `LocalMedia`**: Dedicated `ObservableObject` managing device toggles, selection, and `LiveKitLocalMedia` wrapper tied to the session `Room`.
- **JS View**: `useSession` includes local media references (`cameraTrack`, `microphoneTrack`) via hooks like `useLocalParticipant`, avoiding a standalone media manager.
- **Pre-connect**: Swift optionally wraps connection with `room.withPreConnectAudio` to manage `isListening`; JS infers buffering from a local microphone publication via `isBufferingSpeech`.

### State Exposure & Reactivity 🟢 
- **Swift Model**: Uses `ObservableObject` and `@Published` properties consumed through SwiftUI wrappers (`LiveKitSession`, `LiveKitAgent`).
- **JS Integration**: Returns plain objects optimized for React renders, with internal `EventEmitter`s supporting async waits and updates.
- **Readiness Signals**: Swift exposes convenience flags (`isReady`, `hasAgents`); JS relies on waitable helpers (`waitUntilAvailable`) and derived flags (`isAvailable`, `failureReasons`).

pblazej:
I haven't included most of the comments - until the API is approved 🟢

Sources/LiveKit/Agent/Session.swift
1egoman: suggestion(non-blocking): The web agents sdk puts `tokenSource` first because it was brought up in our meeting with ben/dz that the abstraction could be used outside of agent contexts as well. I see you have some other constructors above but if this is supposed to be an agent use case specific constructor pattern maybe it's worth making it a static method like `Session.withAgent(agentName:agentMetadata:/* ... */)` or something like that?

Sources/LiveKit/Agent/Session.swift
```diff
@@ -0,0 +1,269 @@
...
+             .connecting where isListening,
+             .connected,
+             .reconnecting:
+            true
+        default:
+            false
+        }
+    }
+
+    @Published public private(set) var agents: [Participant.Identity: Agent] = [:]
```
1egoman: thought: You had mentioned in your comment [here](https://github.com/livekit/client-sdk-swift/pull/789#issuecomment-3410475532) that js doesn't store a list of agents, only one right now - that is correct. The way I had been thinking about it, it's unclear if the agents sdk could even really support multiple agents right now (ie, ben had said that agents are unable to hear other agents, etc), so I opted to not implement this.

That being said, the external interfaces the web sdk provides leaves the door open for this in the future (ie, right now `useSession(..., { agentName: 'foo' })` / `useAgent(session)`, and in the future something like `useSession(..., { agents: [{agentName: 'foo'}] })` / `useAgent(session, 'foo')`, with `useAgent(session)` becoming deprecated but returning the first agent that joined for backwards compatibility).

I think I slightly prefer what I did in the web sdk than what you did here but I am curious what you are thinking. I will say that I think whatever we decide, all SDKs should align on this so if we opt for fully supporting multiple agents I will make those changes on the web in a new pull request.

Sources/LiveKit/Agent/Session.swift
```diff
@@ -0,0 +1,269 @@
...
+        defer {
+            waitForAgentTask = Task { [weak self] in
+                try await Task.sleep(nanoseconds: UInt64(timeout * Double(NSEC_PER_SEC)))
+                try Task.checkCancellation()
+                guard let self else { return }
+                if connectionState == .connected, agents.isEmpty {
+                    self.error = .agentNotConnected
+                }
+            }
+        }
```
1egoman: thought: Just wanted to surface another thing you had mentioned in your comment [here](https://github.com/livekit/client-sdk-swift/pull/789#issuecomment-3410475532) - I think you missed that the web implementation not only stores a list of `failureReasons` but ALSO transitions to a new `failed` state. That state transition IMO is the more important behavior.

1egoman: issue: Reading through this, it looks like it doesn't handle the case of multiple agents joining properly because it is one global task, even though your `agents` data structure does handle multiple agents properly.

For example, I'm thinking of this scenario:
1. Agent `a` is dispatched
2. Wait 10 seconds
3. Agent `b` is dispatched
4. Wait 10 more seconds
5. At this point, both agents are still in a non timed out state even though `a` hasn't connected after 20 seconds (the default timeout).

I think in practice this case won't ever be an issue right now because the external `Session` interface won't let you do this, but IMO this bug shouldn't be left lurking when the plan is to allow behavior like this later, and I'd think it wouldn't be that hard to fix (start a timeout task for each agent independently).

Or alternatively, maybe it's worth just storing one agent internally for now like I mentioned [in another comment](https://github.com/livekit/client-sdk-swift/pull/789#discussion_r2435972016), in which case what you are doing here I think works fine.

1egoman: suggestion: I think I prefer something closer to what I did on the web slightly to what you did here - what do you think about this:
1. Add a `failure` case to `AgentState` [here](https://github.com/livekit/client-sdk-swift/blob/main/Sources/LiveKit/Types/Attributes/AttributeTypings.swift#L50-L56).
2. Apply this state to any agent that doesn't join after the specified timeout in this above logic, along with maybe some sort of more detailed error into on the agent (maybe continue the `self.error` pattern on `Agent` as well?)
3. I think it's fine to still store a value on session's `self.error`, as long as it's clear to implementers that the error means that "at least one agent didn't connect" and not "all agents didn't connect".

Sources/LiveKit/Agent/Session.swift
pblazej: ah yes, factory method is even better 👍

Sources/LiveKit/Agent/Session.swift
pblazej: My thinking was just this decision has to be made anyway, e.g. once you introduce this helper "hook" `@LiveKitAgent`, the default behavior must change when introducing multiple agents.

I'd prefer to align all SDKs now because deprecation may lead to awkward patterns on some platforms.

Also, returning to this code after a while - I started thinking about generalizing that even more as dz mentioned - to participants, as the main entry point to the (legacy) room. Wdyt?

Sources/LiveKit/Agent/Session.swift
pblazej: I agree, I subconsciously avoided adding more states 😄
- `Agent` in SwiftUI world must be optional anyway, as it's injected via the magical `Environment` object, and we cannot just crash when it's not there
- introducing the error state that you must handle (from the enum) just for the sake of the error scenario - maybe that's an overkill?
- IMO the core difference is I think more in terms of `Agent == Participant` - it's there where it's there, while you do more of an `AgentPromise` kind of thing that may resolve with an error

My biggest argument here is still introducing more states that we need to handle sort of "manually" (are not just observations of the engine) may lead to awkward behaviors (like with the preconnect API when you need to "superimpose" your local state over the connection state).

I'll try to revisit the JS code, maybe find some consensus.

cc @lukasIO

Sources/LiveKit/Agent/Session.swift
pblazej: 💌 9cf68ff6fdb3d697e675e94301a300593f7c0212

Sources/LiveKit/Agent/Session.swift
1egoman: At least right now on the web, the only maybe "awkward pattern" that would exist is that `useAgent(session, /* no agent name */)` would return the first agent which joined. I don't see that as all that awkward at least on the web, and was thinking that when eventually multiple agents gets introduced more officially down the line the old `useAgent(session, /* no agent name */)` signature could just be deprecated. But that's just for the web, other platforms may be more challenging.

This is to say - from an external interface perspective I don't think the decision here effects the web sdk that much. I'd like to get @lukasIO's perspective just to see if he has a different perspective, but it sounds like you're pushing for the web sdk internals to support multiple agents now, and I'm fine with doing that and keeping the same external web sdk interface for now.

> Also, returning to this code after a while - I started thinking about generalizing that even more as dz mentioned - to participants, as the main entry point to the (legacy) room. Wdyt?

I'm not sure I understand what you are suggesting - can you write up an example?

Sources/LiveKit/Agent/Session.swift
1egoman: > introducing the error state that you must handle (from the enum) just for the sake of the error scenario - maybe that's an overkill?

thought: My opinion - I see it as less overkill and more that without it, the agent `state` value isn't properly representing all the options, so the `Agent` ends up being a "leaky" state machine.

> IMO the core difference is I think more in terms of Agent == Participant - it's there where it's there, while you do more of an AgentPromise kind of thing that may resolve with an error

thought: Yea you are right, on the web returning a "placeholder" agent response is important for a js-related reason (preserving the ability to destructure the `useAgent` return). One other nice thing it unlocks is it means that if an end user queries an agent and gets back an empty value, they don't have to guess exactly what that "lack of value" means - the agent could be in the midst of dispatching, it could have not connected, etc. So with this approach you have to query somewhere else to get that state, and at least on the web, that state being "disconnected" from the agent object results in some ergonomic challenges (type narrowing via discriminated unions won't work properly).

> My biggest argument here is still introducing more states that we need to handle sort of "manually" (are not just observations of the engine) may lead to awkward behaviors (like with the preconnect API when you need to "superimpose" your local state over the connection state).

question: I'm not exactly sure I understand the preconnect API nuance you are describing, can you ellaborate further?

thought: Reading this though, you are making me think that maybe your concerns would be alleviated by splitting up the `AgentState` enum into two levels. Then you are never replacing a state value from `lk.agent.state` with an "internal" state value. So something like:
```typescript
type AgentState = 
| { state: "connecting" }
| {
  state: "connected",
  agent: {
    state: 'initializing' | 'idle' | 'listening' | 'thinking' | 'speaking',
    /* todo: add other agent related metadata in here */
  }
| {
  state: "failed",
  failureReasons: Array<string>, // Or maybe something else, just some way to capture the "why" behind the failure
}
```
(sorry for the typescript and not swift, I'm hoping that gets the point across!)

Sources/LiveKit/Agent/Session.swift
```diff
@@ -0,0 +1,269 @@
...
+                    messages.updateValue(message, forKey: message.id)
+                }
+            }
+        }
+    }
+
+    // MARK: - Agents
+
+    public func agent(named name: String) -> Agent? {
+        agents.values.first { $0.participant.attributes["lk.agent_name"] == name || $0.participant.identity?.stringValue == name }
```
lukasIO: thought: to me at least it's not entirely clear if checking against both identity and name will fall on our feet later on. I don't have a concrete scenario in mind, but it feels a bit unexpected to check implicitly against both.

Sources/LiveKit/Agent/Session.swift
lukasIO: I also believe that even once we introduce multiple agent use cases we don't necessarily have to deprecate the `no agent name passed` variant, as it (as of judging from current use cases) it will probably still be the main usage pattern for a vast majority of users. 

I think we should allow for usage patterns that don't require an agent name to be passed because it will be the main use case. WRT to my other comment about conflating name with identity and agentName, I'd personally be in favour for not exposing agent name at all right now (mirroring JS) and introducing it later as an optional parameter.

Sources/LiveKit/Agent/Session.swift
lukasIO: thought: generally I like Blaze's approach over an explicit failed state. I think we've discussed this in length on the JS PR though and Ryan went with the explicit state. There were pros and cons with either solution, so I don't want to reignite that discussion on the next PR now.

I think we also discussed different AgentState's at one point (conversational state vs connection state of an agent) but that also got dropped. 

From my perspective we consented to these decisions already, and now we should try to make the cross client experience as consistent as possible.

Sources/LiveKit/Agent/Session.swift
pblazej: > not exposing agent name at all right now

@lukasIO so keeping an internal storage for [agents] also does not make sense then?

> generalizing that even more to participants

@1egoman I'm just imagining wrapping `Room` (session "replaces" room) so you want to handle a general case of "participants", even for non-agent users. Then you cannot introduce serious asymmetry between them (agent vs participant).

Sources/LiveKit/Agent/Session.swift
pblazej: @lukasIO so let's adopt web's additional state?

As you see from above, my main point is if we need to introduce additional case, means that `AgentState` is "malformed" internally (does not represent assumptions).

Sources/LiveKit/Agent/Session.swift
pblazej: thx for pointing that out, I think it can be postponed in a single agent world without name param (?)

Sources/LiveKit/Agent/Session.swift
lukasIO: > so keeping an internal storage for [agents] also does not make sense then?

does it hurt? I mean once we add support for passing names it would be easier to do it if the internal map/storage exists already? Totally up to you though.

pblazej: single 🤖 ae38145c2b952cd0d98f45256f7f97b1f1555384

Sources/LiveKit/Agent/Session.swift
pblazej: removed in ae38145c2b952cd0d98f45256f7f97b1f1555384

Sources/LiveKit/Agent/Session.swift
pblazej: @lukasIO @1egoman I started refactoring in this direction, the part I find awkward to port is:
https://github.com/livekit/components-js/blob/f118da6e678c4a91be91c4dfc9b3b61eb7f64e2a/packages/react/src/hooks/useAgent.ts#L394-L425

The obstacles:
- it requires bi-directional `Room` observation
```js
    if (roomConnectionState !== ConnectionState.Disconnected) {
      state = 'connecting';
    }
```
- I'm unable to calculate preconnect state in the agent itself (as the local track is not associated with local participant at the time)
```js
    if (localMicTrack) {
      state = 'listening';
      bufferingSpeachLocally = true;
    }
```
- I could move the `failed` state to the agent, but:
  - to represent `.connecting` (and start the timeout clock) I need speculative `Agent` creation (not via simple unidirectional `Room` observation as it is right now)

Finally, the core difference is:
- in JS Agent is a "computed state"
- in Swift Agent is an object with its own lifecycle
  - why? I started with a similar approach, more like "value semantics", but having a class instead allows us to track local state (without the fear of mutating the copy), inject Agent separately, etc.

In other words, it would be easier now to move the whole state back to `Session` that would own `agentState` - making multiple agents more awkward.

OR

Publish separate Agent(s) where all the state is computed in the `Session` itself.

Hope it makes sense (?)

pblazej: So, coming back to square one:
```swift
waitForAgentTask = Task { [weak self] in
                try await Task.sleep(nanoseconds: UInt64(timeout * Double(NSEC_PER_SEC)))
                try Task.checkCancellation()
                guard let self else { return }
                if state == .connected, agents.isEmpty {
                    self.error = .agentNotConnected // I cannot transition the agent to failed state as agents.isEmpty by definition
                }
            }
```

The closest alternative would be to move `Session` to a `.failed` state, but that's really just another representation of `Error`, just more awkward to consume on the frontend.

lukasIO: > it requires bi-directional Room observation


Maybe I'm misunderstanding "bi-directional" here, but the session itself should not depend on the agent's state. 
A session's connection state (and connection) are separate from whether or not an agent is connected.

pblazej: @lukasIO I mean the other way around - agent depends on session (room indeed): https://github.com/livekit/components-js/blob/f118da6e678c4a91be91c4dfc9b3b61eb7f64e2a/packages/react/src/hooks/useAgent.ts#L402

in Swift these are separate entities as mentioned above, agent is not a computed property over session - so must update its local state, thus the word "bidirectional".

pblazej:
@1egoman @lukasIO in the spirit of "talk is cheap", here's an alternative design, with more "stupid" `Agent`, separate states, etc.

a93fcf0199d2b88e7b44c0d7f5629ce6ea14c171

The key part is:
```swift
public enum Agent {
    case disconnected
    case connecting
    case connected(AgentState, (any AudioTrack)?, (any VideoTrack)?)
    case failed(Error)
}
```
I think the benefits have been justified earlier, possible downsides:
- no local mutable state in the agent - that may be a limitation
  - why not? because it would require either the knowledge of `Room`/`Participant` as it used to be or foreign mutation (from the `Session`) which is equally awkward
- the necessity to use "prism" aka computed property to pull stuff from this shiny pure enum impl
- the state machine that mutates the agent "state" isn't fully safe
  - what to do with multiple agents then? e.g. create them all with corresponding names as `.connecting` then resolve?
- the observation in SwiftUI isn't that granular (more view updates than _really_ needed)

Sources/LiveKit/Agent/Agent.swift
```diff
@@ -0,0 +1,161 @@
...
+        case connecting
+        case connected(agentState: AgentState, audioTrack: (any AudioTrack)?, avatarVideoTrack: (any VideoTrack)?)
+        case failed(Error)
+    }
+
+    private var state: State = .disconnected
+
+    // MARK: - Transitions
+
+    mutating func disconnected() {
```
pblazej, line 60: @lukasIO @1egoman this is the crème de la crème

pblazej:
The state machines:

https://github.com/user-attachments/assets/6a7670ab-42ba-47a4-82ab-f18597a63d67

1egoman (approved):
Nice work on this @pblazej, looking forward to seeing this get merged!

pblazej:
Did one more fix - I noticed that there was a "gap" that we discussed in session vs agent preconnect states: a787606a2e68c926df855adc5e2b4d3fb01a3614

Now it's kinda more obvious with `case connecting(buffering: Bool)` (not exposed as the whole `State`)

Sources/LiveKit/Agent/Agent.swift
```diff
@@ -0,0 +1,151 @@
...
+                "Agent did not connect"
+            }
+        }
+    }
+
+    // MARK: - State
+
+    private enum State {
+        case disconnected
+        case connecting(buffering: Bool)
```
xianshijing-lk, line 51: [question](https://pullpo.io/cc?l=question) 
what this |buffering| mean ? or prebuffering is what it means ?

Sources/LiveKit/Agent/Agent.swift
```diff
@@ -0,0 +1,151 @@
...
+    mutating func failed(error: Error) {
+        log("Agent failed with error \(error) from \(state)")
+        // From any state
+        state = .failed(error: error)
+    }
+
+    mutating func connecting(buffering: Bool) {
+        log("Agent connecting from \(state)")
+        switch state {
+        case .disconnected, .connecting:
```
xianshijing-lk, line 75: [question](https://pullpo.io/cc?l=question) 
should you allow .connected here ? that might means reconnect ?

Sources/LiveKit/Agent/Agent.swift
```diff
@@ -0,0 +1,151 @@
...
+            state = .connecting(buffering: buffering)
+        default:
+            log("Invalid transition from \(state) to connecting", .warning)
+        }
+    }
+
+    mutating func connected(participant: Participant) {
+        log("Agent connected to \(participant) from \(state)")
+        switch state {
+        case .connecting, .connected:
```
xianshijing-lk, line 85: [nitpick](https://pullpo.io/cc?l=nitpick) 

should you handle the noOp operation ? like adding a 

private func assign(_ new: State) {
    guard new != state else { return }
    state = new
}

Then use 
assign(.connected(agentState: participant.agentState,
                               audioTrack: participant.agentAudioTrack,
                               avatarVideoTrack: participant.avatarVideoTrack))

Sources/LiveKit/Agent/Chat/Receive/TranscriptionDelegateReceiver.swift
```diff
@@ -0,0 +1,68 @@
...
+/// { id: "1", content: "Hello world!" }
+/// ```
+@available(*, deprecated, message: "Use TranscriptionStreamReceiver compatible with livekit-agents 1.0")
+actor TranscriptionDelegateReceiver: MessageReceiver, RoomDelegate {
+    private let room: Room
+    private var continuation: AsyncStream<ReceivedMessage>.Continuation?
+
+    init(room: Room) {
+        self.room = room
+        room.add(delegate: self)
```
xianshijing-lk, line 36: [question](https://pullpo.io/cc?l=question) 

should this TranscriptionDelegateReceiver hold the room strong reference ? or it should be weak ? 

Could you please confirm that there is no cycle reference if they hold each other strongly

Sources/LiveKit/Agent/Chat/Receive/TranscriptionDelegateReceiver.swift
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    deinit {
+        room.remove(delegate: self)
+    }
+
+    /// Creates a new message stream for the transcription delegate receiver.
+    func messages() -> AsyncStream<ReceivedMessage> {
+        let (stream, continuation) = AsyncStream.makeStream(of: ReceivedMessage.self)
+        self.continuation = continuation
```
xianshijing-lk, line 47: [question](https://pullpo.io/cc?l=question) 
Can this messages() will be called multiple times? 
If yes,  the current code stores a single continuation, calling messages() again overwrites it and could it leave the old stream hanging ? 
In that case, I am a bit worrying that it never finish the stream either, so consumers can hang ? 

And I wonder if we should have an explicit stop function like 
func stop() {
        room?.remove(delegate: self)
        room = nil
        continuation?.finish()
        continuation = nil
    }

Sources/LiveKit/Agent/Chat/Receive/TranscriptionStreamReceiver.swift
```diff
@@ -0,0 +1,174 @@
...
+    /// Creates a new message stream for the chat topic.
+    func messages() async throws -> AsyncStream<ReceivedMessage> {
+        let (stream, continuation) = AsyncStream.makeStream(of: ReceivedMessage.self)
+
+        let topic = topic
+
+        try await room.registerTextStreamHandler(for: topic) { [weak self] reader, participantIdentity in
+            for try await message in reader where !message.isEmpty {
+                guard let self else { return }
+                await continuation.yield(processIncoming(partialMessage: message, reader: reader, participantIdentity: participantIdentity))
```
xianshijing-lk, line 97: [nitpick](https://pullpo.io/cc?l=nitpick) 
continuation yield() is not async, right ? 
https://developer.apple.com/documentation/swift/asyncstream/continuation/yield(with:)

should you do 
let msg = await self.processIncoming(partialMessage: message, reader: reader, participantIdentity: participantIdentity)
continuation.yield(msg)

Sources/LiveKit/Agent/Session.swift
```diff
@@ -0,0 +1,320 @@
...
+    public func start() async {
+        guard connectionState == .disconnected else { return }
+
+        error = nil
+        waitForAgentTask?.cancel()
+
+        let timeout = options.agentConnectTimeout
+
+        defer {
+            waitForAgentTask = Task { [weak self] in
```
xianshijing-lk: should you move this block of code to the success path ? 

using defer here means that we will start the task when it fails to connect, is it intentional ?

Sources/LiveKit/Agent/Session.swift
```diff
@@ -0,0 +1,320 @@
...
+            true
+        default:
+            false
+        }
+    }
+
+    /// The ``Agent`` associated with this session.
+    @Published public private(set) var agent = Agent()
+
+    @Published private var messagesDict: OrderedDictionary<ReceivedMessage.ID, ReceivedMessage> = [:]
```
xianshijing-lk, line 82: [question](https://pullpo.io/cc?l=question) 

does this ReceiveMessage.ID guarantee to be unique with different participants ? or when connection reconnects ?

Sources/LiveKit/Agent/Session.swift
```diff
@@ -0,0 +1,320 @@
...
+            agent.connected(participant: firstAgent)
+        } else {
+            agent.connecting(buffering: options.preConnectAudio)
+        }
+    }
+
+    private func observe(receivers: [any MessageReceiver]) {
+        for receiver in receivers {
+            Task { [weak self] in
+                for await message in try await receiver.messages() {
```
xianshijing-lk: any chance that  messages() will throw ? 
If that is possible,  this loop just exits quietly. 

How about wrapping in do/catch and log so users know why a stream stopped.

Sources/LiveKit/Agent/Session.swift
```diff
@@ -0,0 +1,320 @@
...
+
+    deinit {
+        waitForAgentTask?.cancel()
+    }
+
+    private func observe(room: Room) {
+        Task { [weak self] in
+            for try await _ in room.changes {
+                guard let self else { return }
+                updateAgent(in: room)
```
xianshijing-lk, line 204: [question](https://pullpo.io/cc?l=question) 
curiously, when room gets changed, do we cancel the current tasks ? or do we need to worry about the cancellation for disconnect ?

Sources/LiveKit/SwiftUI/LocalMedia.swift
```diff
@@ -0,0 +1,219 @@
...
+    public convenience init(session: Session) {
+        self.init(room: session.room)
+    }
+
+    private func observe(_ localParticipant: LocalParticipant) {
+        Task { [weak self] in
+            for try await _ in localParticipant.changes {
+                guard let self else { return }
+
+                microphoneTrack = localParticipant.firstAudioTrack
```
xianshijing-lk, line 105: This localMedia class is @MainActor,
I think you should run code on the main thread: 
await MainActor.run {
      self.microphoneTrack = localParticipant.firstAudioTrack
      self.cameraTrack = localParticipant.firstCameraVideoTrack
      self.screenShareTrack = localParticipant.firstScreenShareVideoTrack
      self.isMicrophoneEnabled = localParticipant.isMicrophoneEnabled()
      self.isCameraEnabled = localParticipant.isCameraEnabled()
      self.isScreenShareEnabled = localParticipant.isScreenShareEnabled()
    }

Sources/LiveKit/SwiftUI/LocalMedia.swift
```diff
@@ -0,0 +1,219 @@
...
+        }
+
+        AudioManager.shared.onDeviceUpdate = { _ in
+            Task { @MainActor [weak self] in
+                self?.audioDevices = AudioManager.shared.inputDevices
+                self?.selectedAudioDeviceID = AudioManager.shared.defaultInputDevice.deviceId
+            }
+        }
+
+        Task {
```
xianshijing-lk, line 129: as these code touches the self properties, use @MainActor [weak self] in

Sources/LiveKit/SwiftUI/LocalMedia.swift
```diff
@@ -0,0 +1,219 @@
...
+    }
+
+    /// Selects a video capture device.
+    /// - Parameter videoDevice: The ``AVCaptureDevice`` to select.
+    public func select(videoDevice: AVCaptureDevice) async {
+        selectedVideoDeviceID = videoDevice.uniqueID
+
+        guard let cameraCapturer = getCameraCapturer() else { return }
+        let captureOptions = CameraCaptureOptions(device: videoDevice)
+        _ = try? await cameraCapturer.set(options: captureOptions)
```
xianshijing-lk: if this cameraCaptureer.set fails, should we still set the selectedVideoDeviceID ? 

How about 
guard let capturer = getCameraCapturer() else { return }
do {
    try await capturer.set(options: .init(device: videoDevice))
    await MainActor.run {
      self.selectedVideoDeviceID = videoDevice.uniqueID
    }
  } catch {
    self.error = .mediaDevice(error)
  }

xianshijing-lk:
looks good. I have some questions / comments, please address them.

Sources/LiveKit/Agent/Agent.swift
pblazej: It's equivalent to JS `isBufferingSpeech` so pre-connect buffer, it's not exposed anywhere, but we can rename it

Sources/LiveKit/Agent/Agent.swift
pblazej: there's no `.reconnecting` state in the agent itself, this overall `Agent.State` is a little artificial as it's derived partially from the room, as mentioned above: https://github.com/livekit/client-sdk-swift/pull/789#discussion_r2448067240

Sources/LiveKit/Agent/Agent.swift
pblazej: From the perspective of UI framework, it does not really matter (for perf) https://medium.com/airbnb-engineering/understanding-and-improving-swiftui-performance-37b77ac61896 as we use the non-equatable (default) comparison for the `State` struct.

Sources/LiveKit/Agent/Chat/Receive/TranscriptionDelegateReceiver.swift
pblazej: I think `Room` does not keep a strong ref to the delegates https://github.com/livekit/client-sdk-swift/blob/585036b367ed23281222082bb696b2d06ce235a7/Sources/LiveKit/Support/MulticastDelegate.swift#L50

Sources/LiveKit/Agent/Chat/Receive/TranscriptionDelegateReceiver.swift
pblazej: This is a general problem with `AsyncStream<>` being exposed here (without storing anything internally). It's not intended to be used by multiple consumers at all (known issue). There's also no equivalent [AnyAsyncSequence](https://forums.swift.org/t/anyasyncsequence/50828) (as `AnyPublisher`).

>  it never finish the stream either,

It can be cancelled from the outside like this:

```swift
let locations = AsyncLocationStream()

let task = Task {
    for await location in locations.stream {
        print(location)
    }
}

task.cancel()
```

Sources/LiveKit/Agent/Chat/Receive/TranscriptionStreamReceiver.swift
pblazej: It's about the method, really: `func processIncoming(...)` is isolated to this `actor`.

Sources/LiveKit/Agent/Session.swift
pblazej: I've been thinking about that, this path essentially guards us: `if isConnected, !agent.isConnected` but I'm fine with moving it as well 385d2bb28400c8de599bfbd541eedaa876d471e1

Sources/LiveKit/Agent/Session.swift
pblazej: Currently, it's just client-side `SentMessage(id: UUID().uuidString ...)`. I decided not to explicitly encode `UUID` vs `String` as there's no real contract with the backend yet.

Sources/LiveKit/Agent/Session.swift
pblazej: 68464148d181c16665858cb7fcc182e583a56913 yeah I think it's doable

Sources/LiveKit/Agent/Session.swift
pblazej: the room is immutable `public let room: Room`, I don't see a reason for cancelling on disconnect while we wanna observe connection state as well: `connectionState = room.connectionState`

Sources/LiveKit/SwiftUI/LocalMedia.swift
pblazej: We're compiling in Swift 6 mode with strict concurrency, so it would not compile if wrong:

image

Non-detached `Task` inherits its actor context implicitly.

What would not compile is `Task.detached {}`

Sources/LiveKit/SwiftUI/LocalMedia.swift
pblazej: as above https://github.com/livekit/client-sdk-swift/pull/789#discussion_r2459670794

Sources/LiveKit/SwiftUI/LocalMedia.swift
pblazej: Yep 99e74b8a900677dc9e718f7a54c8b3684c937ef6

pblazej: BTW there's an issue to improve this API overall https://github.com/livekit/client-sdk-swift/issues/177

pblazej:
@xianshijing-lk thank you for looking into that!

I think I added missing error handlers.

The only thing that is a tradeoff and cannot be easily fixed is the `AsyncStream` in the public API, I think the limitations of that (single consumer) are already known to the community.

I did not want to revert it to `Combine` or callbacks, to keep the cancellation semantics and everything related to `AsyncSequence`.

Sources/LiveKit/Agent/Chat/Receive/TranscriptionDelegateReceiver.swift
pblazej: Re: stream idempotence, I think we've got 2 choices:
- leave is as is - it won't register another consumer for this topic, just `throw`ing `StreamError.handlerAlreadyRegistered`
- unregister before registering, so the previous stream will stop working - I think that's the worst one

Sources/LiveKit/Agent/Agent.swift
xianshijing-lk: either works for me, prebuffering seems more clear to me, but up to you.

PR #800, opened 2025-10-02
CI Test Plan

pblazej:


pblazej:
@hiroshihorie you can disable the device tests (ideally move them into a new plan called `Device` or sth) in spare time 🚀 

If you encounter sth flaky, maybe it's worth to create a 3rd suite called `Flaky` for investigation.

Let's make it 🟢 then try to parallelize.

pblazej:
Here's the status quo:

Screenshot 2025-10-03 at 10 25 24 AM

My plan is to move them to a separate target/filter out.

Unfortunately, test plans vs xcodebuild vs spm is a pain, no silver bullet here.
Moving to a separate test target is also an option (`LiveKitTestsDevice` etc.) but we need to extract test utils first.

pblazej:
@hiroshihorie what I ended up doing:
- extracted common test utils into `Tests/LiveKitTestSupport` target
- split tests into 3 targets (`Core`/`Audio` (things that makes more sense on device)/`ObjC`)
- removed the default test plan - for SPM it does more harm than good
- enabled ALL `Core` + `ObjC` tests on CI
  - retry 2x
  - timeout total 10m
  - parallel (about 1.5x of speed up)

Let's see how flaky it is 🤞 Marking that as ready for review.

BTW I haven't touched the `LKTestHost` at all, if you've got any ideas for simplifying on-device testing, just let me know or push it in a separate PR.

pblazej:
Added tsan/asan variants to the matrix

PR #804, opened 2025-10-03
Fix serial runner cancellation

hiroshihorie:
Problem

The testSerialRunnerCancel() test was failing with the counter ending up negative (-1, -8, etc.) instead of the expected 0, indicating a race condition.

Root Cause

The condition !Task.isCancelled was causing cancelled tasks to skip waiting for the previous task to complete. This broke the serial execution guarantee, allowing multiple blocks to run concurrently and causing unsynchronized access to the shared counter.

Fix

Removed the !Task.isCancelled condition to ensure all tasks wait for the previous task, even when cancelled. This maintains the serial ordering guarantee that the actor is designed to provide.

### DIFF
--- a/Sources/LiveKit/Support/SerialRunnerActor.swift
+++ b/Sources/LiveKit/Support/SerialRunnerActor.swift
@@ -21,8 +21,8 @@ actor SerialRunnerActor<Value: Sendable> {
 
     func run(block: @Sendable @escaping () async throws -> Value) async throws -> Value {
         let task = Task { [previousTask] in
-            // Wait for the previous task to complete, but cancel it if needed
-            if let previousTask, !Task.isCancelled {
+            // Always wait for the previous task to maintain serial ordering
+            if let previousTask {
                 // If previous task is still running, wait for it
                 _ = try? await previousTask.value
             }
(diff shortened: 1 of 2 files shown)


pblazej:
Nice, will take a deeper look early next week 💯

Sources/LiveKit/Support/SerialRunnerActor.swift
```diff
@@ -21,8 +21,8 @@ actor SerialRunnerActor<Value: Sendable> {
 
     func run(block: @Sendable @escaping () async throws -> Value) async throws -> Value {
         let task = Task { [previousTask] in
-            // Wait for the previous task to complete, but cancel it if needed
-            if let previousTask, !Task.isCancelled {
+            // Always wait for the previous task to maintain serial ordering
+            if let previousTask {
```
pblazej, line 25: I'm just wondering what would happen if we move this wait before `Task` (essentially "blocking" `run`)?

pblazej: NVM, it just breaks the assumptions made in the 3rd test case 👍

pblazej (approved):
I rebased it onto `main` as my test branch needs more work 👍

PR #807, opened 2025-10-08
Fixes audio renderer adapter

hiroshihorie:
#784 Broke `LocalAudioTrack.add(audioRenderer: AudioRenderer)` so this PR fixes it...

### DIFF
--- a/Sources/LiveKit/Audio/Manager/AudioManager.swift
+++ b/Sources/LiveKit/Audio/Manager/AudioManager.swift
@@ -117,9 +117,17 @@ public class AudioManager: Loggable {
 
     // MARK: - AudioProcessingModule
 
-    private lazy var capturePostProcessingDelegateAdapter = AudioCustomProcessingDelegateAdapter(label: "capturePost")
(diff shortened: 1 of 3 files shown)


pblazej:
Trying to understand the remaining failures (iOS device) here:
Screenshot 2025-10-09 at 8 51 09 AM

`testStartingTwice()` is redundant, as the API just changed (to `stop` first).

Can we do sth about the others?

hiroshihorie:
Yea most of them need to run with LKTestHost  + local LK server running. Without the test host these audio test are blocked at mic perms I think 🤔 Also tests can't run concurrently because its sharing the ADM singleton.

pblazej:
> Yea most of them need to run with LKTestHost + local LK server running. Without the test host these audio test are blocked at mic perms I think 🤔 Also tests can't run concurrently because its sharing the ADM singleton.

Yes, I'm referring to LKTestHost here. The new test plans are not concurrent, just random. You can try running them on device as well.

pblazej:
Let's unblock `main` and sync on the rest next week.

PR #808, opened 2025-10-08
Stop audio publishing when audio session activation fails

hiroshihorie:
More audio safety checks

### DIFF
--- a/Sources/LiveKit/Audio/AudioSessionEngineObserver.swift
+++ b/Sources/LiveKit/Audio/AudioSessionEngineObserver.swift
@@ -64,28 +64,8 @@ public class AudioSessionEngineObserver: AudioEngineObserver, Loggable, @uncheck
         set { _state.mutate { $0.next = newValue } }
     }
 
-    public init() {
-        _state.onDidMutate = { new_, old_ in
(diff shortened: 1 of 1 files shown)


Sources/LiveKit/Audio/AudioSessionEngineObserver.swift
```diff
@@ -81,26 +81,36 @@ public class AudioSessionEngineObserver: AudioEngineObserver, Loggable, @uncheck
         set { _state.mutate { $0.next = newValue } }
     }
 
-    public init() {
-        _state.onDidMutate = { new_, old_ in
```
pblazej, line 85: Hm, probably a small regression: this change won't work at runtime `isSpeakerOutputPreferred`, as well as `isAutomaticConfigurationEnabled`

pblazej: Reproducible with example app ⏫

pblazej: Also, not sure if `func copy` is the right thing as it may override it atomically (?)

hiroshihorie: Oops yes this isn't great.

PR #809, opened 2025-10-09
Improve logging

pblazej:
Turns out that the configuration closure:
- must be run _before_ any logs (expected) - this is kinda fixable
- may run once per process, so will break for anyone using the same framework... and we cannot do any default setup

If we need the flexibility anyway, maybe it's better to drop the `swift-logging` right now...

My assupmtions:
- we use pure OSLog by default (info or warning)
- we can easily enable rtc internal logs via the same channel
- we can easily provide our custom implementation

### DIFF
--- a/Sources/LiveKit/Support/Logger.swift
+++ b/Sources/LiveKit/Support/Logger.swift
@@ -75,7 +75,6 @@ open class OSLogger: Logger, @unchecked Sendable {
     private let queue = DispatchQueue(label: "io.livekit.oslogger", qos: .utility)
     private var logs: [String: OSLog] = [:]
 
-    private lazy var rtcLog = OSLog(subsystem: Self.subsystem, category: "WebRTC")
     private lazy var rtcLogger = LKRTCCallbackLogger()
 
     private let minLevel: LogLevel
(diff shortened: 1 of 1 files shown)


pblazej:
@hiroshihorie @bcherry I decided to rewrite it from scratch (without waiting for v3), as the previous setup was kinda problematic in multiple ways.

Let me know what you think, especially is there anything "harder to understand than before" for the consumer.

The output looks nice(r) already, can be filtered more easily, etc.:

Screenshot 2025-10-13 at 10 31 25 AM

hiroshihorie:
@pblazej Nice, do we have a example app branch for this ? If not I'll make one.

pblazej:
@hiroshihorie I think these are the only changes you need to make - feel free to test it in spare time 👍 

(ofc you can even omit this change)

image

Sources/LiveKit/Support/Logger.swift
```diff
@@ -0,0 +1,217 @@
...
+
+    private let queue = DispatchQueue(label: "io.livekit.oslogger", qos: .utility)
+    private var logs: [String: OSLog] = [:]
+
+    private lazy var rtcLog = OSLog(subsystem: Self.subsystem, category: "WebRTC")
+    private lazy var rtcLogger = LKRTCCallbackLogger()
+
+    private let minLevel: LogLevel
+
+    public init(minLevel: LogLevel = .info, rtc: Bool = false) {
```
pblazej, line 82: The default level is ofc opinionated here, we can go for `.warning` as well (no consensus on other platforms)... We use these logs sparingly, so for normal connection you just get a couple of messages.

PR #813, opened 2025-10-13
Bump up RPC timeout to 7 seconds, 15s as total response time

xianshijing-lk:
See issue https://linear.app/livekit/issue/CLT-2132/increase-rpc-max-rt-time-to-7s

This is the swift PR for https://github.com/livekit/rust-sdks/pull/729


hiroshihorie (approved):
✅ LGTM

PR #818, opened 2025-10-14
Add lastSpokeAt property to Swift SDK

brainwith:
This PR adds the lastSpokeAt property to the Swift SDK to align with Android and JS implementations.
It represents the last time a participant spoke, allowing consistent behavior across platforms.


brainwith:
> Can we use `Date?` similarly to `joinedAt` I think this is consistent in other SDKs (`Date` in JS, `Long?` in Kotlin, `DateTime` in Flutter)

That’s a great suggestion — I can update it to use Date. The original use of Long was mainly for convenience when sorting on the caller side.

PR #820, opened 2025-10-15
fix: guard unsupported os for extension builds

0xymoro:
Fix for:

Build issue when trying to include the LiveKit CocoaPods package inside the broadcast extension (followed this guide for cocoapods https://github.com/livekit/client-sdk-swift/blob/main/Docs/cocoapods.md), and getting the following error: ".../Pods/LiveKitClient/Sources/LiveKit/Track/Capturers/CameraCapturer.swift:97:32 'LKRTCCameraVideoCapturer' is unavailable in application extensions for iOS: Camera not available in app extensions." Looks like it's caused by the camera capture code being compiled into the extension target, which isn't allowed by iOS, even though only the LKSampleHandler is needed.

### DIFF
--- a/Sources/LiveKit/Track/Capturers/CameraCapturer.swift
+++ b/Sources/LiveKit/Track/Capturers/CameraCapturer.swift
@@ -23,6 +23,8 @@ import ReplayKit
 
 internal import LiveKitWebRTC
 
+@available(iOSApplicationExtension, unavailable, message: "Camera capture is not supported in app extensions.")
+@available(tvOSApplicationExtension, unavailable, message: "Camera capture is not supported in app extensions.")
 public class CameraCapturer: VideoCapturer, @unchecked Sendable {
     /// Current device used for capturing
     @objc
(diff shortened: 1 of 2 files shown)


pblazej:
@0xymoro thank you for the fix, can you try testing this existing PR instead? https://github.com/livekit/client-sdk-swift/pull/811

0xymoro:
Actually believe we need more than this to fix for extension - we're doing this on our side and building but running into more bugs. will figure out

0xymoro:
We fixed our issue by forking, adding a subspec, and some tweaks. We saw 2.8.1 broke so we did these changes on top of 2.7.0 and now broadcast with subspec works properly.

0xymoro:
Actually running into lots of trouble w app validation and framework/embeddings. Waiting on official fix for now while we tinker more, we can build it alright but just can't deploy to app store/testflight.

pblazej:
@0xymoro could you be a little more specific

> into lots of trouble w app validation

are you referring to this issue? cc @hiroshihorie

0xymoro:
Subspec did not work but entirely diff spec did work. This is hacky and we had to rewrite a good bunch of things other than the simple fix so closing this for now

hiroshihorie:
Hi @0xymoro , #811 seems to fix compile issues but what other issues have you ran into ? Would be great to know more details.

PR #821, opened 2025-10-15
clamped the effectiveTimeout and fixed the comment

xianshijing-lk:
So the problem is that the performRpc is a public function, and we never sanity check if users provide values larger than the connection_timeout, which might cause unexpected behavior.

This PR fixes the comment and also clamp the connection response timeout if needed.

See similar PR on JS SDK: https://github.com/livekit/client-sdk-js/pull/1694

### DIFF
--- a/.changes/clamp-response-timeout
+++ b/.changes/clamp-response-timeout
@@ -1,0 +1,1 @@
+patch type="fixed" "clamp response timeout in performRpc to minimal 1s"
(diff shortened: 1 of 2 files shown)


pblazej:
You can add a `.changes/` file if it's worth including in the CHANGELOG.

xianshijing-lk:
I actually have no idea why many tests start to fail with this PR, while the previous PR that bumped up the timeout from 10s to 15s passed the CI.

PR #833, opened 2025-11-05
Camera format selection logic

pblazej:
Trying to fix the issue with non-multi cam formats creating big and unnecessary jumps (e.g. from 640x480 to 1280x1280).

Checks if there's an actual multicam session running (not only the capability).


pblazej:
The "Manhattan" 🏙️ part is not crucial for this PR, but it went through successful customer testing, I think Android impl makes a lot of sense @davidliu @hiroshihorie

PR #845, opened 2025-11-18
Reconnect errors

### LINKED ISSUE #844: reconnect error
The default fast reconnection count is configured to 10 times. Why does it start from the beginning again after the 6th reconnection, starting from the 1st reconnection, and there is a high probability that it will fail to reconnect?

pblazej:
Resolves #844

- Handles `offerId` (symptoms)
- Fixes double offer on ICE restart (root cause)
- Fixes deadlock, probably introduced by strict ordering in #804 (the fix is still valid though)
- Handles `leaveAction` param vs legacy `canReconnect` (requires protocol v16 introduced elsewhere)


Sources/LiveKit/Core/Transport.swift
```diff
@@ -157,12 +172,14 @@ actor Transport: NSObject, Loggable {
...
         func _negotiateSequence() async throws {
+            _latestOfferId += 1
             let offer = try await createOffer(for: constraints)
             try await set(localDescription: offer)
-            try await _onOffer(offer)
+            try await _onOffer(offer, _latestOfferId)
         }
 
         if signalingState == .haveLocalOffer, iceRestart, let sd = remoteDescription {
+            _reNegotiate = false // Clear flag to prevent double offer
```
pblazej, line 182: JS around https://github.com/livekit/client-sdk-js/blob/65e0bfe38ba594bbfe73de60ce72dbc7b96be3a2/src/room/PCTransport.ts#L271

```ts
if (this._pc && this._pc.signalingState === 'have-local-offer') {
    const currentSD = this._pc.remoteDescription;
    if (options?.iceRestart && currentSD) {
        // 1. Rollback: Sets remote description
        await this._pc.setRemoteDescription(currentSD);
        // 2. DOES NOT set this.renegotiate = true
        // 3. Falls through to create offer (line 291)
    } else {
        // 1. Defer: Sets renegotiate = true
        this.renegotiate = true;
        // 2. Returns immediately (skips offer creation)
        return;
    }
}
```

vs Swift

```swift
if signalingState == .haveLocalOffer {
    if !(iceRestart && remoteDescription != nil) {
        // 1. Defer: Sets _reNegotiate = true
        _reNegotiate = true
        // 2. Returns immediately
        return
    }
    
    // Else: ICE Restart path falls through...
}

// ... (offer ID increment) ...

if signalingState == .haveLocalOffer, iceRestart, let sd = remoteDescription {
    // 1. Force clear _reNegotiate (Matches JS not setting it)
    _reNegotiate = false  
    // 2. Rollback: Sets remote description
    try await set(remoteDescription: sd)
    // 3. Creates new offer immediately
    return try await _negotiateSequence()
}
```

Sources/LiveKit/Core/Room+SignalClientDelegate.swift
```diff
@@ -32,10 +32,12 @@ extension Room: SignalClientDelegate {
            // engine is currently connected state
            case .connected = _state.connectionState
         {
-            do {
-                try await startReconnect(reason: .websocket)
-            } catch {
-                log("Failed calling startReconnect, error: \(error)", .error)
+            Task {
```
pblazej, line 37:
```
SerialRunnerActor (inside SignalClient._delegate)
│
├─> [Task 1] didUpdateConnectionState
│   └─> await startReconnect()  ← blocking the actor
│       └─> waiting for offer...
│
└─> [Task 2] didReceiveOffer
    └─> Can't enter because actor is busy with Task 1
```

pblazej:
@hiroshihorie this ain't trivial, I'd appreciate if you give it a spin in ✈️ mode

Sources/LiveKit/Core/Room+SignalClientDelegate.swift
```diff
@@ -32,23 +32,29 @@ extension Room: SignalClientDelegate {
...
     func signalClient(_: SignalClient, didReceiveLeave canReconnect: Bool, reason: Livekit_DisconnectReason) async {
         log("canReconnect: \(canReconnect), reason: \(reason)")
 
+        let error = LiveKitError.from(reason: reason)
+
         if canReconnect {
             // force .full for next reconnect
             _state.mutate { $0.nextReconnectMode = .full }
+            // Abort current connection attempt
+            await signalClient.cleanUp(withError: error)
```
pblazej, line 58: This is more or less equivalent to JS:

```ts
        case LeaveRequest_Action.RECONNECT:
          this.fullReconnectOnNext = true;
          // reconnect immediately instead of waiting for next attempt
          this.handleDisconnect(leaveReconnect);
```

lukasIO: **suggestion (if-minor):** maybe this is a good moment to adopt the `.action` of the leave request as well here? (backwards compatible of course in case it's unset/`0`). 

We don't actually need a full reconnect on every leave request. But if it makes more sense as a follow up that also sounds good.

pblazej: Good point, moved to the action param here 👍

Re: backwards compatibility, I discarded the legacy `canReconnect` param, which mimics JS and no-op if unknown.

PR #847, opened 2025-11-20
Text stream corruption

pblazej:
Using a new `Task` at actor boundary wasn't perfect (not only perf-wise), may lead to phenomena like:
```
Expected: "¡Hola, soy Robin! ¿Eres amigo(a) de Amy?"
Received: "¡Hola, soy Robin!E ¿res amigo(a) de Amy?"
```

Switched to a single stream instead, ensuring FIFO.

### DIFF
--- a/.changes/text-stream-corruption
+++ b/.changes/text-stream-corruption
@@ -1,0 +1,1 @@
+patch type="fixed" "Processing data packets out-of-order leading to stream corruption"


pblazej:
I'm not adding any unit tests here as I was unable to force this particular problem (via concurrent calls).

hiroshihorie (approved):

PR #848, opened 2025-11-20
Do not leak Sendable extensions to consumers

dfed:
LiveKit currently marks multiple types that are not thread-safe as being `Sendable`. These extensions mean that consumer's applications also treat these types as `Sendable`, which can lead to thread-safety bugs.

This PR removes the Sendable extensions without adding new warnings. This is accomplished by utilizing internal wrapper types that are marked as `@unchecked Sendable`.


Sources/LiveKit/Extensions/Sendable.swift
```diff
@@ -48,14 +48,6 @@ extension LKRTCCallbackLogger: @unchecked Swift.Sendable {}
 
 // MARK: Collections
 
-extension NSHashTable: @unchecked Swift.Sendable {} // cannot specify Obj-C generics
```
dfed, line 51: This extension was not used.

Sources/LiveKit/Support/MapTable.swift
```diff
@@ -0,0 +1,54 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import Foundation
+
+/// A thin unchecked sendable wrapper around NSMapTable.
+final class MapTable<KeyType, ObjectType>: @unchecked Sendable where KeyType: AnyObject, ObjectType: AnyObject {
```
dfed, line 20: I did not implement every method on `NSMapTable` here. Just ones that were used.

Sources/LiveKit/Extensions/Sendable.swift
```diff
@@ -58,4 +56,3 @@ extension Dictionary: Swift.Sendable where Key: Sendable, Value: Sendable {}
 
 extension AVCaptureDevice: @unchecked Swift.Sendable {}
 extension AVCaptureDevice.Format: @unchecked Swift.Sendable {}
-extension CVPixelBuffer: @unchecked Swift.Sendable {}
```
dfed, line 61: I worked around this one locally – no need for a global change

Sources/LiveKit/Extensions/Sendable.swift
```diff
@@ -58,4 +56,3 @@ extension Dictionary: Swift.Sendable where Key: Sendable, Value: Sendable {}
 
 extension AVCaptureDevice: @unchecked Swift.Sendable {}
 extension AVCaptureDevice.Format: @unchecked Swift.Sendable {}
```
dfed, line 58: These two are effectively sendable per documentation, so I left them

.changes/sendable-extension
```diff
@@ -0,0 +1,2 @@
+patch type="changed" "Removed '@unchecked Sendable' extensions on common types"
+
```
dfed:
```suggestion
```

Sources/LiveKit/Support/MapTable.swift
pblazej:
```suggestion
final class MapTable<KeyType, ObjectType>: @unchecked Sendable where KeyType: AnyObject & Sendable, ObjectType: AnyObject & Sendable {
```

then you'll need sth like:
```swift
final class VideoRendererAdapter: NSObject, @unchecked Sendable, LKRTCVideoRenderer
```

(another thin wrapper)

wdyt?

Sources/LiveKit/Extensions/Sendable.swift
pblazej: Yup looks like it was replaced with `NSMapTable` below

pblazej:
Thank you for noticing and removing that! Especially, `CVPixelBuffer` thing was not worth it. Left a minor comment.

Sources/LiveKit/Support/MapTable.swift
dfed: I mean, [`NSMapTable`'s functions are **not** thread safe](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/Multithreading/ThreadSafetySummary/ThreadSafetySummary.html), so I'd rather not add more layers of pretend safety. You're already modifying this object from within the lock, which is what makes this code effectively `Sendable`.

I'd be down to rename this type to `UnsafeMapTable` to make it more clear to in-library consumers that this type is not actually `Sendable` and should be treated with caution? This is an `internal` type at least, so it can be iterated on post-merge without affecting consumers of the library.

pblazej: Good point, the whole thing is _unsafe_ anyway.

Sources/LiveKit/Support/MapTable.swift
pblazej, line 1: let's move it to `Support` maybe

PR #857, opened 2025-12-02
Protocol v16, handle Room moved

pblazej:


Sources/LiveKit/Core/Room+SignalClientDelegate.swift
```diff
@@ -135,6 +135,46 @@ extension Room: SignalClientDelegate {
         }
     }
 
+    func signalClient(_: SignalClient, didReceiveRoomMoved response: Livekit_RoomMovedResponse) async {
+        log("didReceiveRoomMoved to room: \(response.hasRoom ? response.room.name : "unknown")")
+
```
hiroshihorie, line 142: We might need to update the room's token to `Livekit_RoomMovedResponse.token`.

Sources/LiveKit/Core/Room+SignalClientDelegate.swift
```diff
@@ -135,6 +135,46 @@ extension Room: SignalClientDelegate {
...
+        // Re-add participants
+        var participantsToAdd: [Livekit_ParticipantInfo] = []
+        if response.hasParticipant {
+            participantsToAdd.append(response.participant)
+        }
+        participantsToAdd.append(contentsOf: response.otherParticipants)
+
+        for info in participantsToAdd {
+            _state.mutate { $0.updateRemoteParticipant(info: info, room: self) }
+        }
```
hiroshihorie: It’s not well documented, but response.participant appears to be a LocalParticipant. We should probably call localParticipant.updateInfo instead. (I’m not sure if any of the info actually changes, though.)

Sources/LiveKit/Core/Room+SignalClientDelegate.swift
```diff
@@ -135,6 +135,46 @@ extension Room: SignalClientDelegate {
...
+        log("didReceiveRoomMoved to room: \(response.hasRoom ? response.room.name : "unknown")")
+
+        // Update room info if available
+        if response.hasRoom {
+            _state.mutate {
+                $0.metadata = response.room.metadata
+                $0.isRecording = response.room.activeRecording
+                $0.numParticipants = Int(response.room.numParticipants)
+                $0.numPublishers = Int(response.room.numPublishers)
+            }
```
hiroshihorie: Other fields like sid / name / maxParticipants / creationTime etc. could be missing ?

hiroshihorie (changes requested):
I think JS SDK emits both RemoteParticipant disconnected / connected events (by using the normal participant disconnect flow etc), maybe we should do that also ?

pblazej:
@hiroshihorie I don't think that's a hi-priority task tbh, any reliable ways to test it?

hiroshihorie:
@pblazej Yes, probably not super high priority. I was hoping Livekit_SimulateScenario as a case but it looks like it doesn't.

pblazej:
Let's try to prioritize https://github.com/livekit/client-sdk-swift/pull/845 (more severe), then fix this one.

pblazej:
@lukasIO lo-priority question: how was it tested on JS (e2e)?

hiroshihorie:
We should bump protocol ver to v16
We can test it with https://docs.livekit.io/intro/basics/rooms-participants-tracks/participants/#moveparticipant

Sources/LiveKit/Core/Room+SignalClientDelegate.swift
```diff
@@ -135,6 +135,84 @@ extension Room: SignalClientDelegate {
...
+            delegates.notify(label: { "room.didMoveToRoomNamed \(response.room.name)" }) {
+                $0.room?(self, didMoveToRoomNamed: response.room.name)
+            }
+            localParticipant.delegates.notify(label: { "participant.didMoveToRoomNamed \(response.room.name)" }) {
+                $0.participant?(self.localParticipant, didMoveToRoomNamed: response.room.name)
+            }
+        }
+
+        // Update local participant
+        if response.hasParticipant {
```
pblazej, line 150: Not sure if we wanna validate this to avoid partial states?

Sources/LiveKit/Participant/LocalParticipant.swift
```diff
@@ -205,6 +205,29 @@ public class LocalParticipant: Participant, @unchecked Sendable {
         sender._set(subscribedQualities: qualities)
     }
 
+    override func set(info: Livekit_ParticipantInfo, connectionState: ConnectionState) {
```
pblazej, line 210: This is more of an inconsistency with JS.

Sources/LiveKit/Types/ProtocolVersion.swift
```diff
@@ -20,9 +20,20 @@ import Foundation
...
-    case v10 = 10 /// Sync stream id
-    case v11 = 11 /// Supports ``ConnectionQuality/lost``
-    case v12 = 12 /// Faster room join (delayed ``Room/sid``)
+    /// Sync stream id
+    case v10 = 10
+    /// Supports ``ConnectionQuality/lost``
+    case v11 = 11
+    /// Faster room join (delayed ``Room/sid``)
+    case v12 = 12
+    /// Regions in leave request, `canReconnect` obsoleted by `action`
```
pblazej, line 29: This would be a nice prerequisite to the changes actually introduced in #845

pblazej:
@hiroshihorie I think the basic logic + protocol is there, working with the CLI:

https://github.com/user-attachments/assets/1080475a-c628-46d0-bef2-aa717fee5182

Sources/LiveKit/Core/Room+SignalClientDelegate.swift
```diff
@@ -135,6 +137,26 @@ extension Room: SignalClientDelegate {
...
+
+        if response.hasRoom {
+            _notifyRoomMoved(name: response.room.name)
+        }
+
+        if response.hasParticipant {
+            localParticipant.set(info: response.participant, connectionState: _state.connectionState)
+        }
+
+        _republishLocalTracks()
```
hiroshihorie, line 154: @pblazej Can you check if this is required ? 🙏 We might not need to re-publish ?

pblazej: yes it is unfortunately, you can check by commenting that out

PR #861, opened 2025-12-05
Fix race condition during sync

### LINKED ISSUE #859: Reconnection complete before remote track set causing an incorrect state sync
**Describe the bug**
At the start of the call, the user's network changed (in our scenario, the user left the office and entered an elevator), switching from Wi-Fi to cellular, a brief network outage occurred.

- After the network was restored, a quick reconnection occurred, the WebSocket connection was completed, marked as `{{1}}` in log.

- A response (ParticipantUpdate) was received from the server, marked as `{{2}}` in log.

- The client then began constructing a state synchronization request (SyncState), but unfortunately, the **actual remote track was not yet ready**, marked as `{{3}}` in log.

  This caused an error in UpdateSubscription within SyncState. The incorrect UpdateSubscription is as follows:

  ```json
  { "track_sids": ["TR_AMTuXiKoMxwLbW"], "subscribe": false }
  ```

- PeerConnection onTrack event fired, but it was too late, server is unsubscribing (LiveKit Server rtc/subscriptionmanager.go:181), then the user could no longer hear the sound, marked as `{{4}}` in log.

**SDK Version**
2.10.0

**iOS/macOS Version**
26.1

**Logs**
Please refer to the attachment and pay attention to the marked locations.

[debug.log](https://github.com/user-attachments/files/23954334/debug.log)

pblazej:
Resolves #859 

Swift SDK uses `isSubscribed` which checks if a track is actually attached, but during reconnection, the `PeerConnection` hasn't fired the `onTrack` event yet, so the track is `nil`, making `isSubscribed = false`.

Using JS convention `isDesired` here.

### DIFF
--- a/Sources/LiveKit/Track/Capturers/CameraCapturer.swift
+++ b/Sources/LiveKit/Track/Capturers/CameraCapturer.swift
@@ -50,9 +50,9 @@ public class CameraCapturer: VideoCapturer, @unchecked Sendable {
     public var isMultitaskingAccessSupported: Bool {
         #if (os(iOS) || os(tvOS)) && !targetEnvironment(macCatalyst)
         if #available(iOS 16, *, tvOS 17, *) {
-            self.capturer.captureSession.beginConfiguration()
-            defer { self.capturer.captureSession.commitConfiguration() }
-            return self.capturer.captureSession.isMultitaskingCameraAccessSupported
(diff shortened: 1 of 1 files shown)


pblazej:
Upgrading swiftformat
  0.57.2 -> 0.58.7

PR #862, opened 2025-12-05
Fix default degradation preference for non-simulcast

pblazej:
Looks like the cropping issues/changing resolution are also related to the default degradation preference that's not applied outside simulcast (not sure why?)

JS:

```ts
export function getDefaultDegradationPreference(track: LocalVideoTrack): RTCDegradationPreference {
  // a few of reasons we have different default paths:
  // 1. without this, Chrome seems to aggressively resize the SVC video stating `quality-limitation: bandwidth` even when BW isn't an issue
  // 2. since we are overriding contentHint to motion (to workaround L1T3 publishing), it overrides the default degradationPreference to `balanced`
  if (
    track.source === Track.Source.ScreenShare ||
    (track.constraints.height && unwrapConstraint(track.constraints.height) >= 1080)
  ) {
    return 'maintain-resolution';
  } else {
    return 'balanced';
  }
}
``

### DIFF
--- a/.changes/degradation-default
+++ b/.changes/degradation-default
@@ -1,0 +1,1 @@
+patch type="fixed" "Default degradation preference for non-simulcast tracks"


pblazej:
User report:

[user_camera_019aea3e-dacb-7b6d-b7ad-8cdce298f4f3.webm](https://github.com/user-attachments/assets/b1731059-761f-467d-a4c6-dd01a559944c)

PR #865, opened 2025-12-09
FrameProcessor protocol

pblazej:
Python reference https://github.com/livekit/python-sdks/pull/533

This is a compromise between existing Audio/Video interfaces, mostly to get uniform "add" behavior from the Room/Session.

The ideal solution in pure Swift would be kinda:

```swift
public protocol FrameProcessor: Sendable {
    associatedtype Input
    associatedtype Output

    func process(frame: Input) -> Output
}
```

This encapsulates the idea of Audio/Video-agnostic interface.

However, the entry points in audio stack must remain ObjC compatible anyway (no generics/associated types), and you cannot get "one protocol conforming to the other" for free (without wrapping/adapters).

### Alternative solution

We can just remove explicit `FrameProcessor` protocol and move `RoomDelegate` conformance to both Audio and Video processor protocols or typealiases. Everything is optional there anyway, so this is a non-braking change.

Then we're left with these 2 "dangling" methods that don't have any explicit representation anywhere, but also we cannot call them directly from the `RoomDelegate`, so the benefits are negligible:

```python
    def _update_stream_info(
        self,
        *,
        room_name: str,
        participant_identity: str,
        publication_sid: str,
    ): ...

    def _update_credentials(self, *, token: str, url: str): ...
```


Sources/LiveKit/Agent/SessionOptions.swift
pblazej, line 1: The minimal implementation is:

```swift
    private static let noiseFilter = AICAudioFilter()
    private let session = Session(
        tokenSource: ...,
        options: SessionOptions(audioProcessor: Self.noiseFilter)
    )
```

Sources/LiveKit/Protocols/AudioCustomProcessingDelegate.swift
```diff
@@ -21,9 +21,14 @@ internal import LiveKitWebRTC
 
 public let kLiveKitKrispAudioProcessorName = "livekit_krisp_noise_cancellation"
 
+public typealias AudioFrameProcessor = AudioCustomProcessingDelegate & FrameProcessor
+
 /// Used to modify audio buffers before they are sent to the network or played to the user
 @objc
 public protocol AudioCustomProcessingDelegate: Sendable {
+    @objc
+    var isEnabled: Bool { get set }
```
pblazej, line 30: This IS a breaking change for Krisp, we can go for `@objc optional` but then this requirement does not make much sense to me.

Sources/LiveKit/Protocols/FrameProcessor.swift
```diff
@@ -0,0 +1,32 @@
...
+    @objc optional
+    func updateStreamInfo(roomName: String, participantIdentity: String, publicationSid: String)
+
+    /// Credentials information.
+    @objc optional
+    func updateCredentials(token: String, url: String)
+
+    /// Called when the processor is no longer needed.
+    @objc optional
+    func close()
```
pblazej: That's a little blind python copy, not sure how can we really utilize that outside `deinit` (destruction).

Sources/LiveKit/Protocols/FrameProcessor.swift
```diff
@@ -0,0 +1,32 @@
...
+
+@objc
+public protocol FrameProcessor: RoomDelegate, Sendable {
+    /// Contextual information about the track.
+    @objc optional
+    func updateStreamInfo(roomName: String, participantIdentity: String, publicationSid: String)
+
+    /// Credentials information.
+    @objc optional
+    func updateCredentials(token: String, url: String)
```
pblazej: We cannot magically call it, so this is basically just a hint that "you need to get the credentials from the `RoomDelegate` somehow".

Sources/LiveKit/Protocols/AudioCustomProcessingDelegate.swift
lukasIO: the alternative would be to not define it here and keep `isEnabled` on the processor implementations, right? 

Maybe that's preferable if the alternative is a breaking change

pblazej: You can also use `@objc optional` annotation, but then you get `Bool?` you need to default to `?? true` for legacy things like Krisp.

Sources/LiveKit/Protocols/FrameProcessor.swift
lukasIO: currently the python code will allow to `close` the aic processor and implicitly recreate it on a next `.process` call. 
The reason why it works this way right now is that on the agents side we need the ability to switch tracks for a processor.

Sources/LiveKit/Protocols/AudioCustomProcessingDelegate.swift
pblazej: The motivation here is obviously to avoid casting in the audio/video thread.

Sources/LiveKit/Protocols/FrameProcessor.swift
lukasIO: should we remove it then?

pblazej: See Alternative solution above 🙏

Sources/LiveKit/Protocols/AudioCustomProcessingDelegate.swift
lukasIO: but implementations could still implement an isEnabled check in the process function and just return early?

Sources/LiveKit/Protocols/FrameProcessor.swift
pblazej: 76c1ebec41a435e40895145862b8af4fa5524d36

Sources/LiveKit/Protocols/AudioCustomProcessingDelegate.swift
pblazej: or should WE do it internally (not passing frames) (see below)

pblazej:
Alternative solution 76c1ebec41a435e40895145862b8af4fa5524d36

pblazej:
Closing this for now due to pausing new filter work ⏯️ Let's revisit if needed.

Issue #828, opened 2025-10-27
When the room is released, there is a high probability of causing a crash.


std-s:
**Describe the bug**
Cause: My app is running version 2.4.0. Testing on iOS 15 revealed that the recipient occasionally couldn't hear audio after making a voice call, but could hear audio during a video call. Investigation revealed that after connecting to a room, the microphone failed to be published (`Failed to publish microphone, error: Error Domain=io.livekit.swift-sdk Code=100 "Cancelled" UserInfo={NSLocalizedDescription=Cancelled}`).

Try this: Updating the LiveKit SDK to version 2.9.0. Occasionally, the call can be heard, but the app crashes.

**SDK Version**
2.9.0

**iOS/macOS Version**
15.8.3

**Xcode Version**
16.1

**Steps to Reproduce**
 When the room is released, there is a high probability of causing a crash.  

**Screenshots**

<img width="367" height="188" alt="Image" src="https://github.com/user-attachments/assets/379cd7b1-1b91-4c60-8c75-e01b50c648b5" />

hiroshihorie:
Hi, just want to collect more info.
Can you share more stack trace / logs when this happens ?
If you Publish audio & release the Room it often causes a crash ?

pblazej:
@std-s can you share the full stack trace? Is it reproducible only with 15.x?

pblazej:
@std-s I pushed a hypothetical fix, would be great to have it reproduced ⤴️

PR #870, opened 2025-12-10
Patch audio processing adapter

hiroshihorie:
Adjusted the audio custom processing adapter to stop reading the WebRTC weak delegate slot (which was calling `objc_loadWeakRetained` during cleanup) and instead track connection state internally before setting/clearing the delegate.

Probably reading `RTC.audioProcessingModule.capturePostProcessingDelegate` (weak) during deinit was not a good idea.

Probably related to #828


pblazej (approved):
lg 🔬

PR #874, opened 2025-12-15
Fix Task cancellation leaks

### LINKED ISSUE #866: Memory Leak in LiveKitWebRTC when streaming via BufferCapturer
Hi @pblazej and @hiroshihorie ,
this is a follow-up to my previous report:  
[Previous issue: Memory Leaking - High memory after Room.disconnect #827](https://github.com/livekit/client-sdk-swift/issues/827)

I can now provide a minimal, runnable LiveKit + ARKit reproduction project that isolates the leak.  
This project contains only the essential components from the LiveKit example plus a minimal custom AR → BufferCapturer bridge.  
This issue focuses **exclusively on internal leaks**.

---

## Summary

The minimal project supports two Room lifecycle modes:

- **Global shared Room instance** (matching the original LiveKit Example)  
  → Reproduces **one leak(LiveKitWebRTC)**

- **Per-connection mode** (fresh `Room()` for each join)  
  → After a leave + rejoin, Instruments reports **more than 24 leaks**

In the per-connection mode, the previous `Room` instance becomes unreachable from the app side (replaced by a new placeholder `Room`), but internal LiveKit objects remain alive and are reported as leaks.

---

## Minimal Reproduction Project

@pblazej and @hiroshihorie, I invited you to the private GitHub project.

The project is a very small SwiftUI app based on these components from the official LiveKit examples:

- `RoomSwitchView`  
- `RoomContextView`  
- `ParticipantView`  
- `ConnectView`  
- `LKButton`, `LKTextField`  
- `SecureStore`  
- `Participant+Helper`  
- `ExampleRoomMessage`  
- `ConnectionHistory`  
- `Bundle` helper  

Added on top:

- A minimal ARKit → BufferCapturer bridge (`ARVideoCapturer`) publishing AR frames as a `LocalVideoTrack`.

---

## How to Run the Minimal Setup

1. **Clone** the provided sample project.  
2. **Configure your LiveKit server URL and token** in `RoomContext` (file: `RoomContext.swift`):

   ```swift
   url = "wss://<your-livekit-host>.livekit.cloud"
   token = "<your-jwt-token>"
   ```
3. **Open** `LiveKit_ARKit_MinimalSetupApp.swift`.  
4. **Choose one of the two lifecycle configurations**:

---

### A. Global shared instance (original example behavior)

```swift
RoomContextView(lifecycleMode: .globalSharedInstance)
```

→ Produces one reproducible leak after leaving the room.

---

### B. New Room per join (`perConnectionNewInstance`)

```swift
RoomContextView(lifecycleMode: .perConnectionNewInstance)
```

→ Produces more than 24 leaks after rejoining a session.

In this mode:

- `connect()` creates a fresh `Room()`.  
- After `disconnect()`, the previous Room is replaced by a **new placeholder Room**  
  (alternatively it could be set to `nil`, but that's not required for the repro).

5. **Run** the application
6. **Join** a session
7. Activate the camera (top right corner)
8. **Leave** the session
9. Repeat joining a session

However, the **previous** Room instance is no longer referenced anywhere in the app.  
Despite that, Instruments shows multiple LiveKit internal structures as leaked  
(DataChannel / Async / Delegate / StateSync objects, etc.).

pblazej:
Resolves #866

Fixes remaining concurrency antipatterns (mostly implicit `self` inside a long-running `Task`) so that `deinit` (and cancel etc.) is never called.

Introduces `AnyTaskCancellable` wrapper for more Combine-like semantics (RAII).

⚠️ This is a global change affecting core functionality e.g. WebSocket, DataChannel, etc.

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -19,6 +19,6 @@ type_name:
 custom_rules:
   no_manual_task_management:
     name: "No Manual Task Management"
-    regex: "(let|var)\\s+\\w+\\s*:\\s*Task<Void,\\s*Never>"
-    message: "Prefer AnyTaskCancellable over manually managing Task<Void, Never>. This allows for safer cleanup, similar to Combine's AnyCancellable."
+    regex: "(let|var)\\s+\\w+\\s*:\\s*Task<Void,"
+    message: "Prefer AnyTaskCancellable over manually managing Task<Void, ...>. This allows for safer cleanup, similar to Combine's AnyCancellable."
     severity: warning
--- a/Sources/LiveKit/Track/Capturers/ARCameraCapturer.swift
+++ b/Sources/LiveKit/Track/Capturers/ARCameraCapturer.swift
@@ -21,7 +21,7 @@ import Foundation
 internal import LiveKitWebRTC
 
 @available(visionOS 2.0, *)
-public class ARCameraCapturer: VideoCapturer {
+public class ARCameraCapturer: VideoCapturer, @unchecked Sendable {
     private let capturer = RTC.createVideoCapturer()
     private let arKitSession = ARKitSession()
     private let cameraFrameProvider = CameraFrameProvider()
@@ -29,8 +29,7 @@ public class ARCameraCapturer: VideoCapturer {
     /// The ``ARCaptureOptions`` used for this capturer.
     public let options: ARCameraCaptureOptions
 
-    // swiftlint:disable:next no_manual_task_management
-    private var captureTask: Task<Void, Never>?
+    private var captureTask: AnyTaskCancellable?
 
     init(delegate: LKRTCVideoCapturerDelegate, options: ARCameraCaptureOptions) {
         self.options = options
@@ -59,7 +58,7 @@ public class ARCameraCapturer: VideoCapturer {
                     capture(pixelBuffer: sample.pixelBuffer, capturer: capturer, options: options)
                 }
             }
-        }
+        }.cancellable()
 
         return true
     }
@@ -84,7 +83,6 @@ public class ARCameraCapturer: VideoCapturer {
         guard didStop else { return false }
 
         arKitSession.stop()
-        captureTask?.cancel()
         captureTask = nil
 
         return true
--- a/Sources/LiveKit/Track/Capturers/MacOSScreenCapturer.swift
+++ b/Sources/LiveKit/Track/Capturers/MacOSScreenCapturer.swift
@@ -42,7 +42,7 @@ public class MacOSScreenCapturer: VideoCapturer, @unchecked Sendable {
         var scStream: SCStream?
         // Cached frame for resending to maintain minimum of 1 fps
         var lastFrame: LKRTCVideoFrame?
-        var resendTimer: Task<Void, Error>?
+        var resendTimer: AnyTaskCancellable?
     }
 
     private var _screenCapturerState = StateSync(State())
@@ -127,7 +127,6 @@ public class MacOSScreenCapturer: VideoCapturer, @unchecked Sendable {
 
         // Stop resending paused frames
         _screenCapturerState.mutate {
-            $0.resendTimer?.cancel()
             $0.resendTimer = nil
         }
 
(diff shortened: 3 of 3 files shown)


Sources/LiveKit/Core/SignalClient.swift
```diff
@@ -207,14 +202,16 @@ actor SignalClient: Loggable {
     }
 
     func cleanUp(withError disconnectError: Error? = nil) async {
+        if disconnectError is CancellationError { return }
```
pblazej: This is a little ugly and started to fail as the upstream task got (properly?) cancelled, similarly to `Room`:

```swift
if !Task.isCancelled {
  // Finally disconnect if all attempts fail
  await cleanUp(withError: error)
}
```

I don't have any immediate idea here though.

Sources/LiveKit/Support/Async/AsyncSequence+Subscribe.swift
```diff
@@ -0,0 +1,193 @@
...
+    ///
+    /// - Parameters:
+    ///   - observer: The observer object (captured weakly).
+    ///   - priority: The priority of the task.
+    ///   - state: The initial mutable state.
+    ///   - onElement: Called for each element.
+    ///   - onFailure: Called when the sequence terminates with an error. Cancellation errors are ignored.
+    /// - Returns: The task cancellable.
+    func subscribe<O: AnyObject & Sendable, State: Sendable>(
+        _ observer: O,
```
pblazej, line 33: Maybe I'll add some more descriptive label like `subscribe(weakly: self)` etc.

Sources/LiveKit/Core/SignalClient.swift
hiroshihorie: How about something like this instead https://github.com/livekit/client-sdk-swift/commit/60536ba4c8ad5c2ffc345ab54034bca3fceda12e

pblazej: Yeah, it boils down to our previous discussion about `type == .cancelled` vs `CancellationError`, I think there should be one single way to propagate that.

At this point I'm fine with both versions, neither is perfect, I'd rather invest into porting that to Rust + proper state machine.

In the future we need a stronger abstraction, something like:

```swift
Task {
   // do something
} cleanup: {
  // defer
}
```

pblazej: Or more event-driven approach...

Tests/LiveKitCoreTests/Room/RoomTests.swift
```diff
@@ -47,8 +47,34 @@ class RoomTests: LKTestCase, @unchecked Sendable {
...
             let socket = await room.signalClient._state.socket
-            self.noLeaks(of: socket!)
+            try self.noLeaks(of: XCTUnwrap(socket))
+
+            let (publisher, subscriber) = room._state.read { ($0.publisher, $0.subscriber) }
+            if let publisher { self.noLeaks(of: publisher) }
+            if let subscriber { self.noLeaks(of: subscriber) }
+
+            self.noLeaks(of: room.publisherDataChannel)
+            self.noLeaks(of: room.subscriberDataChannel)
```
pblazej, line 59: @hiroshihorie I compared with main again and confirmed that the main leak comes from DC (its main event loop).

Sources/LiveKit/Core/SignalClient.swift
pblazej: Tested that and it won't work in practice:

```
/Users/blaze/livekit/client-sdk-swift/Tests/LiveKitCoreTests/DataChannel/RealiableDataChannelTests.swift:67 testReliableRetry(): Asynchronous wait failed: Exceeded timeout of 10 seconds, with unfulfilled expectations: "Data received".
```

pblazej:
Linking this one for reference: https://medium.com/the-swift-cooperative/using-async-for-await-youre-probably-doing-it-wrong-88b66fbb0e84

pblazej:
FYI, I added a brand-new SwiftLint rule to prevent bare `Task<>`

PR #876, opened 2025-12-18
Fix memory leaks in AppStateListener and Room

viachaslauinnowise:
Fix: memory leaks in AppStateListener observers and Room cleanup

AppStateListener:
- Store NotificationCenter observer tokens for proper cleanup
- Add deinit to remove all observers
- Use [weak self] in all observer closures

Room:
- Remove from AppStateListener.delegates in deinit (safety net)
- Explicitly remove in cleanUp() (primary path)
- Prevent memory leaks when Room is destroyed without cleanUp()

### DIFF
--- a/Sources/LiveKit/Core/Room.swift
+++ b/Sources/LiveKit/Core/Room.swift
@@ -454,6 +454,18 @@ public class Room: NSObject, @unchecked Sendable, ObservableObject, Loggable {
             $0.reconnectTask = nil
         }
     }
+
+    deinit {
+        // Ensure Room is removed from AppStateListener to prevent memory leaks
+        // This is a safety net in case cleanUp() wasn't called explicitly
+        // Note: Task may not complete before deinit finishes, but delegates use weak references
+        // so Room will be removed from memory eventually even if this doesn't execute
+        // The primary cleanup should happen in cleanUp() method
+        let room = self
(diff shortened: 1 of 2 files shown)


Sources/LiveKit/Core/Room.swift
```diff
@@ -500,6 +512,11 @@ extension Room {
                 disconnectError: LiveKitError.from(error: disconnectError)
             )
         }
+
+        // Stop listening to app state changes to avoid retaining Room longer than needed
+        await MainActor.run {
+            AppStateListener.shared.delegates.remove(delegate: self)
```
pblazej, line 518: Can you explain why this is needed if `AppStateListener` uses:

```swift
private struct State {
  let delegates = NSHashTable<AnyObject>.weakObjects()
}
```

Sources/LiveKit/Support/AppStateListener.swift
```diff
@@ -40,50 +40,75 @@ class AppStateListener: Loggable {
...
+                guard let self else { return }
+                handler()
+            }
+        }
+
+        _observerTokens.append((center, token))
+    }
+
+    deinit {
+        for (center, token) in _observerTokens {
```
pblazej, line 64: This is a singleton, so where the removal will occur?

PR #880, opened 2025-12-23
docs: refresh audio guide and README

hiroshihorie:
Add more audio related docs


pblazej (approved):

pblazej:
We can think of following nitpicks re: docs
- breaking into (expandable) sections as the README gets too long/flat
- (re)moving some legacy UIKit docs etc. that are not super important
- adding graphical overview of the audio pipeline 📈

PR #882, opened 2025-12-30
Deprecate cocoapods

pblazej:
Probably it makes sense for native, not sure about flutter and rn.

https://blog.cocoapods.org/CocoaPods-Specs-Repo/


.github/workflows/cocoapods-push.yaml
pblazej, line 1: This is failing trying to push to `main`.

hiroshihorie (approved):

PR #883, opened 2025-12-31
CI: Fix publish docs workflow

pblazej:
I think getting something from `main` may interfere with other runs (won't wait), switched to explicit commit.

Let's see how it goes during the next release.


pblazej:
https://github.com/livekit/client-sdk-swift/actions/runs/20615303561/job/59206797284

hiroshihorie (approved):

PR #884, opened 2025-12-31
SwiftLint

pblazej:
Bypasses "expected" cyclomatic complexity etc. checks, worth improving while refactoring ofc.


.swiftlint.yml
```diff
@@ -0,0 +1,17 @@
+excluded:
+  - Sources/LiveKit/Protos
+
+disabled_rules:
+  - todo
+  - nesting
+  - identifier_name
```
pblazej, line 7: This is mostly for `_privateFields`, not sure if that's worth fixing.

`Found 231 violations, 231 serious in 321 files.`

pblazej: Maybe in a separate PR

PR #905, opened 2026-02-04
Workaround for Catalyst crash during teardown

### LINKED ISSUE #864: [Mac Catalyst] EXC_BAD_ACCESS crash in audio device cleanup during room.disconnect()
Body:

  ## Description

  When calling `room.disconnect()` on Mac Catalyst, the app crashes with `EXC_BAD_ACCESS` in the WebRTC audio device
  module. This does not occur on iOS - only Mac Catalyst.

  ## Environment

  - **SDK Version:** 2.10.1
  - **Platform:** Mac Catalyst (iOS app running on Mac)
  - **macOS:** Sequoia 15.x
  - **Xcode:** 16.x

  ## Crash Details

  LiveKitWebRTC`___lldb_unnamed_symbol14016:
  -> 0x104242384 <+0>: str xzr, [x0, #0x40]
     0x104242388 <+4>: ret

  worker_thread: EXC_BAD_ACCESS (code=1, address=0x40)

  ## Console Output Before Crash

  AudioHardware-mac-imp.cpp:773 AudioObjectGetPropertyData: no object with given ID 113
  AVAEInternal.h:83 required condition is false: [AVAudioEngine.mm:1269:CheckCanPerformIO: (canPerformIO)]
  AudioEngineDevice setVoiceProcessingEnabled error: (null)

  ## Reproduction Steps

  1. Run iOS app on Mac via Catalyst
  2. Connect to a LiveKit room with audio
  3. Start a voice session (microphone enabled)
  4. Call `room.disconnect()` or end the session
  5. App crashes immediately

  ## Minimal Code

  ```swift
  // Simple disconnect - crashes on Mac Catalyst
  await room.disconnect()

  What I've Tried

  - Adding delays before/after disconnect (100ms to 1000ms)
  - Disabling microphone before disconnect
  - Clearing room references before disconnect
  - Using setRecordingAlwaysPreparedMode(true)
  - SDK versions 2.8.1 and 2.10.1

  None of these workarounds prevent the crash.

  Analysis

  The crash appears to be in the WebRTC audio device module trying to access an aggregate audio device (ID 113) that
  has already been destroyed. The worker_thread is still running when the audio device cleanup occurs, causing a
  dangling pointer access at offset 0x40.

  Expected Behavior

  room.disconnect() should cleanly disconnect from the room on Mac Catalyst without crashing.

  Workaround

  Currently none - the only option is to disable Mac Catalyst support.

  ---

pblazej:
Resolves #864

### DIFF
--- a/Sources/LiveKit/Audio/Manager/AudioManager.swift
+++ b/Sources/LiveKit/Audio/Manager/AudioManager.swift
@@ -413,6 +413,16 @@ public class AudioManager: Loggable {
         _state = StateSync(State(engineObservers: engineObservers))
         _admDelegateAdapter.audioManager = self
         RTC.audioDeviceModule.observer = _admDelegateAdapter
+
+        #if targetEnvironment(macCatalyst)
+        // Disable Apple's Voice-Processing I/O on Catalyst to avoid aggregate device
+        // race condition crashes during teardown. WebRTC's software processing is used instead.
+        // See https://github.com/livekit/client-sdk-swift/issues/864
+        let result = RTC.audioDeviceModule.setVoiceProcessingEnabled(false)
+        if result != 0 {
+            log("Failed to disable voice processing on Catalyst: \(result)", .warning)
+        }
+        #endif
     }
 }
 
--- a/Sources/LiveKit/Types/Options/AudioCaptureOptions.swift
+++ b/Sources/LiveKit/Types/Options/AudioCaptureOptions.swift
@@ -26,6 +26,13 @@ public final class AudioCaptureOptions: NSObject, CaptureOptions, Sendable {
     public static let defaultEchoCancellation = true
     public static let defaultAutoGainControl = true
     public static let defaultNoiseSuppression = true
+    #elseif targetEnvironment(macCatalyst)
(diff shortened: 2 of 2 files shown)


pblazej:
@hiroshihorie Based on the warnings, I think the root cause is a timing issue in audio callbacks during teardown that works "slightly differently" on Catalyst (maybe Apple's internal bug).

However, to fix that we'd need some synchronization layer in the audio path, which seems risky (deadlock etc) - I'd rather disable VP on (legacy) Catalyst platform - as a better tradeoff.

hiroshihorie:
@pblazej Thanks for investigating. Yes, it is likely one of the many weird audio issues of Audio Foundation / Apple.

For now, we can suggest:

AudioManager.shared.setVoiceProcessingEnabled(false)

for Mac Catalyst users who are facing this issue.

I am not sure we should disable it entirely for all Mac Catalyst users, since I do not think many users run into this issue. For example, we cannot reproduce it.

I'd like to see if changing the timing of stopping the AVAudioEngine (in WebRTC ADM) helps/reduces aggregated device related logs, which could be related to the crash.

pblazej:
> timing of stopping the AVAudioEngine

you mean hard-coding some delay directly?

PR #911, opened 2026-02-09
WebRTC m144

pblazej:
Notable metnions:
- visionOS 26 (Package v6.2)
- `WebRTC-IceHandshakeDtls/Enabled/` enabled by **default**

### DIFF
--- a/.changes/vision-26
+++ b/.changes/vision-26
@@ -1,0 +1,1 @@
+minor type="changed" "Changed minimal supported visionOS version to 26 due to WebRTC dependencies"
--- a/.changes/webrtc-m144
+++ b/.changes/webrtc-m144
@@ -1,0 +1,1 @@
+minor type="changed" "Update WebRTC to m144"


Sources/LiveKit/Types/DegradationPreference.swift
```diff
@@ -20,21 +20,25 @@ internal import LiveKitWebRTC
 public enum DegradationPreference: Int, Sendable {
     /// The SDK will decide which preference is suitable or will use WebRTC's default implementation.
     case auto
+    @available(*, deprecated, renamed: "maintainFramerateAndResolution")
```
pblazej, line 23: Removed from the header

Package@swift-6.2.swift
pblazej, line 1: This is to support visionOS v26, I hope we'll drop Swift 5.9 soon https://developer.apple.com/news/upcoming-requirements/?id=02032026a

Sources/LiveKit/Core/RTC.swift
```diff
@@ -96,6 +96,9 @@ actor RTC {
 
         Room.log("Initializing PeerConnectionFactory...")
 
+        // Enable DTLS during ICE handshake for faster connection setup.
+        LKRTCPeerConnectionFactory.configureFieldTrials("WebRTC-IceHandshakeDtls/Enabled/")
```
pblazej, line 100: It's a bit awkward (static), mostly due to multiple `[init]` paths in PC, but I think we can live with that.

PR #919, opened 2026-02-17
Single Peer Connection

pblazej:
- Type-safe transport management via `TransportMode` (mirrors JS naming) 🎉 
  - This is not purely additive, so requires some regression testing 💣 
- Benchmarks will come in a separate PR (once we have the methodology)
- There's no separate test for this, just switching the **default** behavior
  - This is debatable, but the question is what to duplicate here... (which cases)
  - I think JS does the exact [opposite](https://github.com/livekit/client-sdk-js/blob/ea4230ea8f20fc538d21e523fc69b59196ccb258/src/api/SignalClient.test.ts#L88-L94)
  - Rust creates separate (basic) [cases](https://github.com/livekit/rust-sdks/blob/main/livekit/tests/peer_connection_signaling_test.rs)

### DIFF
--- a/Sources/LiveKit/Core/TransportMode.swift
+++ b/Sources/LiveKit/Core/TransportMode.swift
@@ -16,7 +16,7 @@
 
 internal import LiveKitWebRTC
 
-enum TransportMode: Equatable, Sendable {
+enum TransportMode: Equatable {
     /// Single peer connection: publisher handles both publishing and receiving.
     case publisherOnly(publisher: Transport)
     /// Dual peer connection with subscriber as primary (default).


Sources/LiveKit/Core/Room+Engine.swift
```diff
@@ -472,13 +484,21 @@ extension Room {
...
-        guard let subscriber = _state.subscriber else {
-            log("Subscriber is nil", .error)
-            return
-        }
+        let previousAnswer: LKRTCSessionDescription?
+        let previousOffer: LKRTCSessionDescription?
 
-        let previousAnswer = await subscriber.localDescription
-        let previousOffer = await subscriber.remoteDescription
+        if _state.isSinglePeerConnection {
```
pblazej: Refactor into single "transport" state

Sources/LiveKit/Core/SignalClient.swift
```diff
@@ -115,19 +115,28 @@ actor SignalClient: Loggable {
...
         if let reconnectMode {
             log("[Connect] mode: \(String(describing: reconnectMode))")
         }
 
-        let url = try Utils.buildUrl(url,
-                                     connectOptions: connectOptions,
-                                     reconnectMode: reconnectMode,
-                                     participantSid: participantSid,
-                                     adaptiveStream: adaptiveStream)
+        let url: URL = if singlePeerConnection {
```
lukasIO, line 132: missing a bit of context here around the reconnection logic, but I don't see any part in this diff that would allow for auto fallback to `buildUrl` in case the connection with join request fails?

pblazej: yup, it should cover "unsupported server" cases (fallback) automatically, right?

lukasIO: yeah

pblazej: this is closer to the intention 2b98cf8860ff869f1870b140446cd318eaac7ef7

Sources/LiveKit/Core/Room+Engine.swift
```diff
@@ -147,23 +145,19 @@ extension Room {
...
-            let isSubscriberPrimary = joinResponse.subscriberPrimary
-            log("subscriberPrimary: \(joinResponse.subscriberPrimary)")
-
-            let subscriber = try Transport(config: rtcConfiguration,
-                                           target: .subscriber,
-                                           primary: isSubscriberPrimary,
-                                           delegate: self)
+            let isSinglePC = singlePeerConnection
+            let isSubscriberPrimary = isSinglePC ? false : joinResponse.subscriberPrimary
+            log("subscriberPrimary: \(isSubscriberPrimary), singlePeerConnection: \(isSinglePC)")
```
xianshijing-lk, line 155: is the log intentional ?

xianshijing-lk:
can you port these single / dual PC rust tests over to swift ? 
https://github.com/livekit/rust-sdks/blob/c0b0868cef64a5d705d7f91def1d99f715730aba/livekit/tests/peer_connection_signaling_test.rs

They should be able to catch most of the issues I ran into when working on the single peer connection issues on Rust

pblazej:
> rust tests

Done, there's some duplication with existing tests, plus I left `true` as the default pc mode for other tests (no hard feelings here).

The reason I did not add them is most of these scenarios are covered *somewhere* e.g. in data channel tests (reconnect, etc.).

pblazej:
Getting some timeouts in tests, looks like general problem with the runner.

xianshijing-lk:
Curiously, do you need the same DSP munging in Swift SDK like in Rust SDK ? 
```
Munge SDP to change a=inactive to a=recvonly for RTP media m-lines in single PC mode.
This is needed because WebRTC can generate inactive direction even when transceivers
were configured as recvonly.
```

pblazej:
> DSP munging

Yes, seems like it's necessary for the same reasons ⬇️

pblazej:
There's some flakiness in these stress tests - not sure if caused by invalid server state or something on the client yet.

Tests/LiveKitTestSupport/Room.swift
pblazej, line 1: I made some *global* changes here to reduce test flakiness, key findings:
- `publishManyTracks` can poison other tests as the server is unable to listen for socket connections
- it does not look like a bug in the server code (listener is a separate goroutine), but rather a restricted VM environment (3 cores here)
- it works reasonably well for fewer tracks
- we can consider moving to `-xlarge` runners for other "stress" tests, however it's paid

pblazej:
@lukasIO @xianshijing-lk do you see any gaps in the logic? You've got the full context.

@hiroshihorie is the swift side acceptable?

Sources/LiveKit/Core/Room+Engine.swift
```diff
@@ -147,23 +145,20 @@ extension Room {
...
 
+            // Publisher always created; is primary in single PC mode
             let publisher = try Transport(config: rtcConfiguration,
                                           target: .publisher,
-                                          primary: !isSubscriberPrimary,
+                                          primary: isSinglePC || !isSubscriberPrimary,
```
xianshijing-lk, line 160: curiously, what does primary mean here ? why primary is true when isSinglePC is true ?

Sources/LiveKit/Support/Utils.swift
```diff
@@ -210,6 +202,103 @@ class Utils: Loggable {
...
+
+        guard var builder = URLComponents(url: url, resolvingAgainstBaseURL: false) else {
+            throw LiveKitError(.failedToParseUrl)
+        }
+
+        let wsScheme = url.isSecure ? "wss" : "ws"
+
+        var pathSegments = url.pathComponents
+        pathSegments.removeAll(where: { $0.isEmpty || $0 == "/" })
+        if !url.hasDirectoryPath, !pathSegments.isEmpty,
```
xianshijing-lk: nit, wonder if it is easier to just do 
if let last = pathSegments.last, ["rtc", "validate"].contains(last)

Sources/LiveKit/Core/Room+Engine.swift
pblazej: it's .debug and it existed before: `log("subscriberPrimary: \(joinResponse.subscriberPrimary)")`

xianshijing-lk:
> @lukasIO @xianshijing-lk do you see any gaps in the logic? You've got the full context.
> 
> @hiroshihorie is the swift side acceptable?

@pblazej , can you check if things are working with auto_subscribe is set to false ? 
We just found out some bugs there in Rust.

Sources/LiveKit/Core/Room+Engine.swift
pblazej: it maps the inverse relationship between the `TransportMode` and `Transport`, mostly in `Room+TransportDelegate` (transport wouldn't know if it's "the only one" in single PC mode):

```swift
    func transport(_ transport: Transport, didUpdateState pcState: LKRTCPeerConnectionState) {
        log("target: \(transport.target), connectionState: \(pcState.description)")

        // primary connected
        if transport.isPrimary {
            if pcState.isConnected {
                primaryTransportConnectedCompleter.resume(returning: ())
```

Sources/LiveKit/Support/Utils.swift
pblazej: done cecef05027b4c1797866b42c30b2d6cb541c37fb

pblazej:
> We just found out some bugs there in Rust.

@xianshijing-lk I think there's some regression here vs main: after unsubscribing a video track I'm left with 8x8 "empty" video - is it confirmed to be on the SFU side?

pblazej:
The above issue with (un)subscribe seems to be tangentially related to transceiver management again, as we were listening to a wrong signal: b64d2961b1dc8e22f4e84256295f5b9c629706bd

Now it mimics Rust https://github.com/livekit/rust-sdks/blob/56669aecf4bbd1875889b7af4612cf37e86ed364/livekit/src/room/publication/remote.rs#L251

There's no way to mimic the above `TODO`, as there's no such event in Swift.

cc @xianshijing-lk anything else reproducible in Rust?

pblazej:
Merging as there are no more _known_ issues + the upcoming release must be re-tested for regressions anyway.

PR #920, opened 2026-02-17
v2.12.1

pblazej:


pblazej:
As mentioned elsewhere, protocol will be updated in the next PR.

PR #922, opened 2026-02-19
Migrate tests to Swift Testing

pblazej:
- Better support for concurrency
- Explicit serialization, tags, etc.
- Parametrization


pblazej:
I'll probably pause ⏸️ it for now, as we need to drop Xcode 15 first, which should happen later this year (https://developer.apple.com/news/upcoming-requirements/?id=02032026a)

pblazej:
Looks like there's some _usual flakiness_ for sims/catalyst - will monitor that.

hiroshihorie (approved):

pblazej:
I made a couple of tweaks to make it manageable on CI, I don't see any systematic/reproducible errors anymore, hard to isolate that from runner performance these days...

To be monitored 🔍

PR #925, opened 2026-02-25
Benchmarks

pblazej:


Benchmarks/LiveKitBenchmark/BenchmarkSupport/BenchmarkConfig.swift
```diff
@@ -0,0 +1,60 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import Foundation
+
+/// Configuration for benchmark runs, read from environment variables.
+struct BenchmarkConfig {
```
xianshijing-lk, line 20: do we have similar code for other use cases ? like integration tests ?

xianshijing-lk (approved):
lgtm

Benchmarks/LiveKitBenchmark/BenchmarkSupport/BenchmarkConfig.swift
pblazej: nope, this is a completely separate framework (with separate infra) just including the sdk (`../`)

pblazej:
Also @xianshijing-lk can you take a look at https://github.com/livekit/client-sdk-benchmarks/pull/1 first (as the source of truth)?

PR #926, opened 2026-02-27
ObjC tests

pblazej:
Adds *actual* integration tests for ObjC bindings to avoid confusion, especially with auto-generation of `async` life forms.

### DIFF
--- a/.changes/objc-tests
+++ b/.changes/objc-tests
@@ -1,0 +1,1 @@
+patch type="added" "Missing data stream and RPC ObjC exposure with ObjC tests"


Sources/LiveKit/DataStream/Outgoing/ByteStreamWriter.swift
```diff
@@ -70,23 +70,3 @@ extension ByteStreamWriter {
 
     private static let fileReadChunkSize = 4096
 }
-
-// MARK: - Objective-C compatibility
-
-public extension ByteStreamWriter {
```
pblazej, line 76: These were `unavailable` from objc anyway... so no breaking change. Relying on auto-generation instead (where possible).

Issue #914, opened 2026-02-12
The stream of the shared screen frequently lags on the receiver's end, especially when the sender rotates the device to landscape orientation.


indenpendman:
iOS all version 
livekitclient  2.7.x -  2.11.x

  func connect(token: String, enableCamera:Bool, enableMic: Bool) {
        // 连接配置
        var connectOptions: ConnectOptions
        switch XMAppEnvironmentManager.shared.currentType {
        case .test, .test2, .dev:
            connectOptions = ConnectOptions(
                autoSubscribe: true,                     // 自动订阅其他参与者的音视频
                reconnectAttempts: 2,                   // 重连尝试次数（默认10次）
                reconnectAttemptDelay: 1,              // 初始重连延迟（秒）
                reconnectMaxDelay: 10,                   // 最大重连延迟（秒）
                socketConnectTimeoutInterval: 10,        // WebSocket连接超时（秒）
                primaryTransportConnectTimeout: 30,       // 主传输连接超时（秒）
                publisherTransportConnectTimeout: 5,      // 发布传输连接超时（秒）
                iceTransportPolicy: .all
            )
        default:
            connectOptions = ConnectOptions(
                autoSubscribe: true,                     // 自动订阅其他参与者的音视频
                reconnectAttempts: 2,                   // 重连尝试次数（默认10次）
                reconnectAttemptDelay: 1,              // 初始重连延迟（秒）
                reconnectMaxDelay: 10,                   // 最大重连延迟（秒）
                socketConnectTimeoutInterval: 10,        // WebSocket连接超时（秒）
                primaryTransportConnectTimeout: 30,       // 主传输连接超时（秒）
                publisherTransportConnectTimeout: 5,      // 发布传输连接超时（秒）
                iceTransportPolicy: .relay
            )
        }

        // 房间配置
        let roomOptions = RoomOptions(
            // 摄像头采集配置
            defaultCameraCaptureOptions: CameraCaptureOptions(
                position: .front,
                dimensions: .h720_169,
                fps: 24
            ),
            // 共享屏幕配置
            defaultScreenShareCaptureOptions: ScreenShareCaptureOptions(
                dimensions: .h720_169,
                fps: 30,
                appAudio: true
            ),
            // 音频采集配置，算法复杂，消耗CPU
            defaultAudioCaptureOptions: AudioCaptureOptions(
                // 回声消除
                echoCancellation: true,
                // 自动增益控制
                autoGainControl: false,
                // 降噪
[...]

PR #928, opened 2026-03-03
Fix broadcast crash in PiP

### LINKED ISSUE #927: Crash in BroadcastImageCodec.jpegDecode when app is backgrounded with PiP screenshare active (iOS) [EXC_BAD_ACCESS (KERN_INVALID_ADDRESS)]
Our iOS app crashes inside the new broadcast IPC pipeline while a LiveKit screen share is being displayed in Picture‑in‑Picture and the main app is in the background.

From the Xcode debugger, the crash happens on a background thread in `BroadcastImageCodec.jpegDecode(_:) ` Please see attached screenshot from Xcode showing the crash and the highlighted code in BroadcastImageCodec.swift.

This occurs while receiving screen‑share video from a remote participant; the user has put the app into the background and is watching the screen share via iOS PiP.

It looks like a decoder crash while converting the incoming JPEG frames to a CVImageBuffer in BroadcastImageCodec.swift.

Steps to reproduce:
- Join a LiveKit room from iOS using the Swift client SDK.
- Have another participant start screen share.
- On the iOS device, enter Picture‑in‑Picture so the screen share continues in a PiP window.
- Background the app (e.g. press Home / swipe up to go to SpringBoard, or switch to another app) while PiP is still showing the remote screen share.
- Keep the call running for a while.
- Result: after some time (often while still in PiP with the app backgrounded), the app crashes with a stack trace showing BroadcastImageCodec.jpegDecode_L_3.

Expected behavior
- The app should not crash while decoding broadcast screen‑share frames.
- Screen share should continue rendering in PiP with the app in the background, or if backgrounding is no longer supported, it should fail gracefully without a decoder crash.

Additional context
- At the time of the crash, the main app is backgrounded and only the PiP window is visible.
- Regular video (camera) and screen share work fine while the app is in the foreground; we have only seen this crash with: App in background & PiP active when Remote screen share ongoing
- I did not find any existing open issue mentioning `BroadcastImageCodec / jpegDecode `crashes. The closest related work I saw was PR #565 “Revamp broadcast IPC”, which introduced `BroadcastImageCodec`.

LiveKit Swift SDK version: 137.7151.10
iOS version: 18.5 
Device: TODO: iPhone 13 Pro Max
Xcode version: 16.2 (16C5032a)

Image

Image

pblazej:
Resolves #927

More (performance) changes may come in #914 
Bg blur uses gpu by default

### DIFF
--- a/Sources/LiveKit/Extensions/CoreImage.swift
+++ b/Sources/LiveKit/Extensions/CoreImage.swift
@@ -16,7 +16,7 @@
 
 import CoreImage
 
-public extension CIContext {
+extension CIContext {
     /// CPU-based context safe for use when the app may be backgrounded (e.g. during PiP).
     static func softwareRenderer() -> CIContext {
         CIContext(options: [.useSoftwareRenderer: true])
--- a/Sources/LiveKit/VideoProcessors/BackgroundBlurVideoProcessor.swift
+++ b/Sources/LiveKit/VideoProcessors/BackgroundBlurVideoProcessor.swift
@@ -71,12 +71,15 @@ public final class BackgroundBlurVideoProcessor: NSObject, @unchecked Sendable, 
 
     // MARK: Init
 
+    private let useMetal: Bool
+
     /// Initialize the background blur video processor.
     /// - Parameters:
     ///   - highQuality: If true, use more detailed segmentation, but at the cost of performance.
-    ///   - gpu: If true, use GPU rendering for better performance. Set to false when the app may be backgrounded (e.g. during PiP).
-    public init(highQuality: Bool = true, gpu: Bool = true) {
-        ciContext = gpu ? .metal() : .softwareRenderer()
+    ///   - useMetal: If true, use Metal rendering for better performance. Set to false when the app may be backgrounded (e.g. during PiP).
+    public init(highQuality: Bool = true, useMetal: Bool = true) {
+        self.useMetal = useMetal
+        ciContext = useMetal ? .metal() : .softwareRenderer()
         super.init()
         segmentationRequest.qualityLevel = highQuality ? .balanced : .fast
     }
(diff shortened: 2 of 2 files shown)


pblazej:
I'll rebase the other PR onto that before the release as it contains the drop frame calc.

Sources/LiveKit/VideoProcessors/BackgroundBlurVideoProcessor.swift
hiroshihorie: nit: Is this path still invoked for gpu(metal) = false ?

Sources/LiveKit/VideoProcessors/BackgroundBlurVideoProcessor.swift
```diff
@@ -72,8 +72,12 @@ public final class BackgroundBlurVideoProcessor: NSObject, @unchecked Sendable,
     // MARK: Init
 
     /// Initialize the background blur video processor.
-    /// - Parameter highQuality: If true, use more detailed segmentation, but at the cost of performance.
-    public init(highQuality: Bool = true) {
+    /// - Parameters:
+    ///   - highQuality: If true, use more detailed segmentation, but at the cost of performance.
+    ///   - gpu: If true, use GPU rendering for better performance. Set to false when the app may be backgrounded (e.g. during PiP).
+    public init(highQuality: Bool = true, gpu: Bool = true) {
```
hiroshihorie: nit: Maybe `useMetal` better ?

Sources/LiveKit/Extensions/CoreImage.swift
hiroshihorie: This could be internal / private ?

hiroshihorie (approved):
✅ LGTM just some nits 🙂

Sources/LiveKit/VideoProcessors/BackgroundBlurVideoProcessor.swift
pblazej: Metal attrs should not be harmful, but may introduce copies, so removed them.

PR #936, opened 2026-03-06
Fix preffered buffer size

pblazej:
Resolves #934

PR #938, opened 2026-03-09
Set preferred audio session sample rate and IO buffer duration

hiroshihorie:
Set preferred sample rate (48kHz) and IO buffer duration (20ms) to match WebRTC defaults


pblazej:
Started a similar thing here but doesn't seem to solve the problem https://github.com/livekit/client-sdk-swift/pull/936

hiroshihorie:
Closing this for now in favor of https://github.com/livekit/client-sdk-swift/pull/936

PR #943, opened 2026-03-10
Sound API 2

hiroshihorie:
Conveniently prepare and play short audio clips locally and/or for remote participants.

```swift
// Prepare
let sound = try await SoundPlayer.shared.prepare(
    fileURL: fileURL,
    named: "sample01" // optional
)

// Playback
try await sound.play(
    options: SoundPlaybackOptions(
        mode: .concurrent,
        loop: false,
        destination: .localAndRemote
    )
)

// Stop / release
await sound.stop()
await sound.release()
```

Remote playback is best-effort and usually requires the microphone/WebRTC input path to be published.

Example: https://github.com/livekit-examples/swift-example/pull/89

### DIFF
--- a/Sources/LiveKit/Audio/AudioSessionEngineObserver.swift
+++ b/Sources/LiveKit/Audio/AudioSessionEngineObserver.swift
@@ -116,8 +116,18 @@ public class AudioSessionEngineObserver: AudioEngineObserver, Loggable, @uncheck
     /// Use this to keep the audio session active from external components
     /// (e.g., ``SoundPlayer``) that need playout or recording independently
     /// of the WebRTC engine lifecycle.
-    public func set(requirement: SessionRequirement, for id: UUID) {
-        _state.mutate { $0.sessionRequirements[id] = requirement }
+    ///
+    /// - Throws: ``LiveKitError`` if the audio session fails to configure or activate.
+    public func set(requirement: SessionRequirement, for id: UUID) throws {
+        try _state.mutate {
+            let oldState = $0
+            $0.sessionRequirements[id] = requirement
+            let result = configureIfNeeded(oldState: oldState, newState: $0)
+            if result != 0 {
+                $0 = oldState
+                throw LiveKitError(.audioSession, message: "Failed to configure audio session")
+            }
+        }
     }
 
     // MARK: - Audio Session Configuration
--- a/Sources/LiveKit/Audio/AudioSessionEngineObserver.swift
+++ b/Sources/LiveKit/Audio/AudioSessionEngineObserver.swift
@@ -210,16 +210,11 @@ public class AudioSessionEngineObserver: AudioEngineObserver, Loggable, @uncheck
 
     public func engineWillEnable(_ engine: AVAudioEngine, isPlayoutEnabled: Bool, isRecordingEnabled: Bool) -> Int {
         let requirement = SessionRequirement(isPlayoutEnabled: isPlayoutEnabled, isRecordingEnabled: isRecordingEnabled)
-        let result: Int = _state.mutate {
-            let oldState = $0
-            $0.sessionRequirements[sessionRequirementId] = requirement
-            let result = configureIfNeeded(oldState: oldState, newState: $0)
-            if result != 0 {
-                $0 = oldState
-            }
-            return result
+        do {
+            try set(requirement: requirement, for: sessionRequirementId)
+        } catch {
+            return kAudioEngineErrorFailedToConfigureAudioSession
         }
-        guard result == 0 else { return result }
         return _state.next?.engineWillEnable(engine, isPlayoutEnabled: isPlayoutEnabled, isRecordingEnabled: isRecordingEnabled) ?? 0
     }
 
@@ -227,16 +222,11 @@ public class AudioSessionEngineObserver: AudioEngineObserver, Loggable, @uncheck
         let nextResult = _state.next?.engineDidDisable(engine, isPlayoutEnabled: isPlayoutEnabled, isRecordingEnabled: isRecordingEnabled) ?? 0
 
         let requirement = SessionRequirement(isPlayoutEnabled: isPlayoutEnabled, isRecordingEnabled: isRecordingEnabled)
-        let result: Int = _state.mutate {
-            let oldState = $0
-            $0.sessionRequirements[sessionRequirementId] = requirement
-            let result = configureIfNeeded(oldState: oldState, newState: $0)
-            if result != 0 {
-                $0 = oldState
-            }
-            return result
+        do {
+            try set(requirement: requirement, for: sessionRequirementId)
+        } catch {
+            return kAudioEngineErrorFailedToConfigureAudioSession
         }
-        guard result == 0 else { return result }
         return nextResult
     }
 }
--- a/Sources/LiveKit/Audio/AudioSessionEngineObserver.swift
+++ b/Sources/LiveKit/Audio/AudioSessionEngineObserver.swift
@@ -87,7 +87,11 @@ public class AudioSessionEngineObserver: AudioEngineObserver, Loggable, @uncheck
         _state.onDidMutate = { [weak self] new, old in
             guard let self,
                   new.isSpeakerOutputPreferred != old.isSpeakerOutputPreferred else { return }
-            _ = configureIfNeeded(oldState: old, newState: new)
+            do {
+                try configureIfNeeded(oldState: old, newState: new)
+            } catch {
+                log("Failed to configure audio session after speaker preference change: \(error)", .error)
+            }
         }
     }
 
@@ -128,8 +132,9 @@ public class AudioSessionEngineObserver: AudioEngineObserver, Loggable, @uncheck
             let oldState = $0
             block(&$0.sessionRequirements)
             guard $0.sessionRequirements != oldState.sessionRequirements else { return }
-            let result = configureIfNeeded(oldState: oldState, newState: $0)
-            if result != 0 {
+            do {
+                try configureIfNeeded(oldState: oldState, newState: $0)
+            } catch {
                 $0 = oldState
                 throw LiveKitError(.audioSession, message: "Failed to configure audio session")
             }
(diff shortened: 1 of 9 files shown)


Sources/LiveKit/Audio/SoundPlayer.swift
```diff
@@ -0,0 +1,308 @@
...
+        return converter.convert(from: buffer)
+    }
+
+    // MARK: - Public API
+
+    /// Decodes and caches audio for the given identifier.
+    ///
+    /// Preparing a sound also acquires a playback session requirement and starts the
+    /// local engine early to reduce first-play latency.
+    public func prepare(url: URL, withId id: String) throws {
```
pblazej: I think we should be explicit about not supporting remote URLs (which is a good decision) e.g. `fileURL` + validate scheme (throw).

Also, main thread behavior: have you considered turning `SoundPlayer` into (global) actor instead?

Nit: it can return some `SoundHandle` (discardable) passed into `play` as an alternative to `id` (maybe an overkill ☠️)

Sources/LiveKit/Audio/SoundPlayer.swift
```diff
@@ -0,0 +1,308 @@
...
+    /// Stops all playing or queued sounds for the specified id without releasing prepared audio buffers.
+    public func stop(id: String, destination: PlaybackOptions.Destination = .localAndRemote) {
+        _state.mutate { state in
+            state.sounds[id]?.stop(destination: destination)
+        }
+    }
+
+    /// Plays a prepared sound.
+    ///
+    /// Remote playback is best-effort and is skipped if remote routing is unavailable.
```
pblazej: There's no way to propagate the error atm, right?

Sources/LiveKit/Audio/SoundPlayer.swift
```diff
@@ -0,0 +1,308 @@
...
+    ///
+    /// - Throws: ``LiveKitError`` if the sound is not prepared or local playback setup fails.
+    public func play(id: String, options: PlaybackOptions = PlaybackOptions()) throws {
```
pblazej: Nit: maybe we can introduce `playOnce(fileURL: URL, options: ...)` for the simplest cases (without separate `prepare`)

Sources/LiveKit/Audio/SoundPlayer.swift
```diff
@@ -0,0 +1,308 @@
...
+        self.mode = mode
+        self.loop = loop
+        self.destination = destination
+    }
+}
+
+public class SoundPlayer: Loggable, @unchecked Sendable {
+    // MARK: - Public
+
+    public static let shared = SoundPlayer()
```
pblazej, line 37: Have you considered an explicit lifecycle here (e.g. coupled with `AudioManager`, `Room` etc.)

Sources/LiveKit/Audio/AudioSessionEngineObserver.swift
```diff
@@ -64,18 +64,40 @@ public class AudioSessionEngineObserver: AudioEngineObserver, Loggable, @uncheck
         set { _state.mutate { $0.isSpeakerOutputPreferred = newValue } }
     }
 
+    /// Represents an audio session requirement from a specific component.
+    ///
+    /// Multiple components (e.g., WebRTC engine, SoundPlayer) can independently
+    /// register their requirements. The audio session stays active as long as
+    /// any component requires playout or recording.
+    public struct SessionRequirement: Sendable, Equatable {
```
pblazej: This idea is great and can save us a lot of headaches in other areas.

Nit: I'd convert it to `OptionSet` similarly to this:

```swift
  public struct SessionRequirement: OptionSet, Sendable {
      public let rawValue: Int
      public init(rawValue: Int) { self.rawValue = rawValue }

      public static let playout   = SessionRequirement(rawValue: 1 << 0)
      public static let recording = SessionRequirement(rawValue: 1 << 1)
  }
```

Sources/LiveKit/Audio/AudioSessionEngineObserver.swift
```diff
@@ -89,6 +111,47 @@ public class AudioSessionEngineObserver: AudioEngineObserver, Loggable, @uncheck
...
+    /// Register or update an audio session requirement for the given identifier.
+    ///
+    /// Use this to keep the audio session active from external components
+    /// (e.g., ``SoundPlayer``) that need playout or recording independently
+    /// of the WebRTC engine lifecycle.
+    ///
+    /// Setting ``SessionRequirement/none`` removes the requirement.
+    ///
+    /// - Throws: ``LiveKitError`` if the audio session fails to configure or activate.
+    public func set(requirement: SessionRequirement, for id: UUID) throws {
```
pblazej: Nit: is it the best choice to provide `id`s from the outside vs return a handle (to unregister myself)?

Sources/LiveKit/Audio/SoundPlayer.swift
```diff
@@ -0,0 +1,308 @@
...
+
+    // MARK: - Private
+
+    private let engine = AVAudioEngine()
```
pblazej: Is alternative design (e.g. Twilio's single `VoiceProcessingIO` AU) possible atm? I think we don't care about having two engines vs latency etc.?

Sources/LiveKit/Audio/PlayerNodePool.swift
```diff
@@ -0,0 +1,176 @@
...
+    private struct NodeItem {
+        let node: AVAudioPlayerNode
+        var isInUse: Bool = false
+        var generation: UInt64 = 0
+    }
+
+    private let audioCallbackQueue = DispatchQueue(label: "audio.playerNodePool.queue")
+    private let _state: StateSync<[NodeItem]>
+
+    init(poolSize: Int = 10) {
```
pblazej, line 54: Probably should be lazy.

Sources/LiveKit/Audio/SoundPlayer.swift
pblazej: The same applies to one node per sound design vs "mix before".

Sources/LiveKit/Audio/PlayerNodePool.swift
```diff
@@ -0,0 +1,176 @@
...
+            guard items[index].generation == generation else { return nil }
+            items[index].isInUse = false
+            return items[index].node
+        }
+        queueNodeStop(node)
+    }
+
+    private func queueNodeStop(_ node: AVAudioPlayerNode?) {
+        guard let node else { return }
+        audioCallbackQueue.async(qos: .default, flags: .enforceQoS) {
```
pblazej: This is to prevent prioritizing `stop()` if hi-priority thread is waiting?

Sources/LiveKit/Audio/SoundPlayer.swift
```diff
@@ -0,0 +1,308 @@
...
+    }
+
+    private func stopEngine(state: inout State) {
+        state.playerNodeFormat = nil
+        playerNodePool.stop()
+        engine.stop()
+    }
+
+    /// Converts a buffer to the target format. Returns the buffer as-is if formats already match.
+    private func convertBuffer(_ buffer: AVAudioPCMBuffer, to targetFormat: AVAudioFormat) throws -> AVAudioPCMBuffer {
```
pblazej: Maybe the converted one should be cached?

pblazej (commented):
Great job ⏯️ 🎉 

Added some general discussion/questions to the _mix_.

I'd also add some documentation focusing on:
- supported formats
- prepare+play example
- pool exhaustion etc.

Are we going to (slowly) deprecate this?

```swift
// A helper class to play-out audio which conforms to the AudioRenderer protocol.
public class AudioPlayerRenderer
```

Sources/LiveKit/Audio/SoundPlayer.swift
hiroshihorie: Yes, good point and  I wanted to go single engine design at first (https://github.com/webrtc-sdk/webrtc/pull/218) but i didn't want to introduce extra complexity yet, I can migrate to that in later version without breaking the API.

Sources/LiveKit/Audio/Manager/AudioManager.swift
```diff
@@ -22,6 +22,63 @@ import Combine
...
+    private let releaseImpl: @Sendable () throws -> Void
+
+    init(releaseImpl: @escaping @Sendable () throws -> Void) {
+        self.releaseImpl = releaseImpl
+    }
+
+    /// Releases the associated audio session requirement.
+    ///
+    /// Releasing the same handle multiple times is a no-op.
+    public func release() throws {
```
pblazej, line 86: This probably should be called on `deinit`.

Sources/LiveKit/Audio/SoundPlayer.swift
pblazej: Nit: if this is a singleton, can leverage https://www.avanderlee.com/concurrency/global-actor/

pblazej (approved):
I think this is more than enough for V1, good point about gradual evolution.

Trivia I'd address before merging:
- rename `url` to `fileURL` still (just self-documenting)
- auto-release session on handle's `deinit`
- adding `globalActor`

Follow-ups:
- `playOnce` addition (no hard feelings)?
- lazy pool
- caching converted audio

For me as a consumer the biggest confusion comes from the fact that `.remote` cannot be played without mic track - which is not clear just reading the APIs. I get the best-effort tradeoff, maybe we can escalate the log to `.warning` at least / or rename the enum more aggressively:

```swift
    public enum Destination: Sendable {
        case local
        case remoteIfAvailable // ?
        case localAndRemoteIfAvailable
```

that could be an `OptionSet` as well e.g.

```swift
  public struct Destination: OptionSet, Sendable {
      public static let local  = Destination(rawValue: 1 << 0)
      public static let remoteIfAvailable = Destination(rawValue: 1 << 1)
  }
```

you pass like this:

```swift
  options: .init(destination: [.local, .remoteIfAvailable])
  options: .init(destination: .local)
```

hiroshihorie:
Thanks @pblazej , will work on a second pass of polishing this up.

hiroshihorie:
playOnce API is good to have but with current implementation, it will configure the audio session like setActive(true) -> play out -> setActive(false) for each call and isn't great user experience.
I think we can add it after adding some kind of debounce logic for setActive(false). Thats a good one for v2 🙂

PR #947, opened 2026-03-12
docs: add missing doc comments and fix copy-paste error in Room.swift

Sleywill:
## Summary

Two types of fix in `Room.swift`:

### 1. Copy-paste error (bug)

`serverNodeId` had the same doc comment as `serverRegion`:

```swift
// Before (wrong — copy-pasted from serverRegion)
/// Region code the client is currently connected to.
public var serverNodeId: String? { _state.serverInfo?.nodeID.nilIfEmpty }

// After (correct)
/// The ID of the server node the client is currently connected to.
public var serverNodeId: String? { _state.serverInfo?.nodeID.nilIfEmpty }
```

### 2. Missing doc comments

Added `///` documentation to 10 public properties that had none:

`name`, `serverVersion`, `remoteParticipants`, `activeSpeakers`, `creationTime`, `maxParticipants`, `participantCount`, `publishersCount`, `token`, `disconnectError`

No behaviour changes.

### DIFF
--- a/Sources/LiveKit/Core/Room.swift
+++ b/Sources/LiveKit/Core/Room.swift
@@ -44,32 +44,41 @@ public class Room: NSObject, @unchecked Sendable, ObservableObject, Loggable {
         try await _sidCompleter.wait()
     }
 
+    /// The name of the Room, as assigned on the LiveKit server.
     public var name: String? { _state.name }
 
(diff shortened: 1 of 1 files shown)


pblazej:
@Sleywill could you sign the cla?

PR #954, opened 2026-03-24
Workaround for video view orientation freeze

pblazej:
Resolves #949 

This is just a workaround - there are some possible improvements on WebRTC side (avoid unbounded waits etc.), but we're unable to isolate the root cause for now.

IMO that's a good tradeoff for older devices (approaching end of support).

### DIFF
--- a/Sources/LiveKit/Views/VideoView.swift
+++ b/Sources/LiveKit/Views/VideoView.swift
@@ -775,7 +775,7 @@ extension VideoView {
 
     /// `true` when no Metal device exists or the GPU is Apple family ≤ 6
     /// (pre-A14 / M1), where the WebRTC Metal path is unstable under rotation.
-    static let prefersSampleBufferRenderer: Bool = {
+    static let shouldAvoidMetalRenderer: Bool = {
         guard let device = MTLCreateSystemDefaultDevice() else { return true }
         return !device.supportsFamily(.apple7)
     }()
@@ -784,7 +784,7 @@ extension VideoView {
         #if os(iOS) || os(macOS)
         let resolvedRenderMode: VideoView.RenderMode = {
             if case .auto = renderMode {
-                return prefersSampleBufferRenderer ? .sampleBuffer : .metal
+                return shouldAvoidMetalRenderer ? .sampleBuffer : .metal
             }
             return renderMode
         }()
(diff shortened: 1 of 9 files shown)


pblazej:
No progress on that as it seems to depend on texture orientation (on selected devices); no way to reproduce yet.

rykkers:
> No progress on that as it seems to depend on texture orientation (on selected devices); no way to reproduce yet.

I can send a device to you that experiences the issue every time if that helps?

pblazej:
Thanks @rykkers for joining the discussion, what kind of device it is (older iPhone)?

I believe this is due to subtle timing difference on older gpus or different metal behavior (or both), unfortunately there's no obvious place to put `sleep` in.

If you wanna dig deeper and contribute, you are also welcome to pull: https://github.com/webrtc-sdk/webrtc

EDIT: I tried downgrading to 12, but still 📹 does not freeze...

rykkers:
> Thanks @rykkers for joining the discussion, what kind of device it is (older iPhone)?
> 
> I believe this is due to subtle timing difference on older gpus or different metal behavior (or both), unfortunately there's no obvious place to put `sleep` in.
> 
> If you wanna dig deeper and contribute, you are also welcome to pull: https://github.com/webrtc-sdk/webrtc
> 
> EDIT: I tried downgrading to 12, but still 📹 does not freeze...

It's an iPad 9th Gen running 26.2

Trust me, you don't want me contributing code. I'm a web-dev that's just dabbling in iOS as a workaround to issues in Webkit!

hiroshihorie (approved):

Sources/LiveKit/Views/VideoView.swift
```diff
@@ -773,9 +773,23 @@ extension VideoView {
         #endif
     }
 
+    /// `true` when no Metal device exists or the GPU is Apple family ≤ 6
+    /// (pre-A14 / M1), where the WebRTC Metal path is unstable under rotation.
+    static let prefersSampleBufferRenderer: Bool = {
```
xianshijing-lk: nit, maybe it is more clear to name it as 
shouldAvoidMetalRenderer

xianshijing-lk (approved):
lgtm with one nit. I will ok regardless if you change the name or not.

PR #963, opened 2026-04-13
Fix snapshot builds

pblazej:
Haven't found a better workaround yet - it's more about testing newer Swift than SPM itself.

### DIFF
--- a/Sources/LiveKit/Agent/Session.swift
+++ b/Sources/LiveKit/Agent/Session.swift
@@ -119,9 +119,10 @@ open class Session: ObservableObject {
                  senders: [any MessageSender]?,
                  receivers: [any MessageReceiver]?)
     {
+        let room = options.room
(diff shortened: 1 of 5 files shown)


Sources/LiveKit/Agent/Session.swift
```diff
@@ -119,9 +119,10 @@ open class Session: ObservableObject {
                  senders: [any MessageSender]?,
                  receivers: [any MessageReceiver]?)
     {
+        let room = options.room
```
pblazej, line 122: This is required due to hardened `self` access in `MainActor` isolated init.

pblazej:
Build: https://github.com/livekit/client-sdk-swift/actions/runs/24336559363/job/71054753156

Warnings in separate PRs.

hiroshihorie (approved):

PR #964, opened 2026-04-13
Skip audio session config in manual rendering mode

hiroshihorie:
I think we had this before but it got lost somewhere 🤔


pblazej (approved):
Probably removed in #793

PR #968, opened 2026-04-15
Add E2EE configuration to Session API

pblazej:
Mirrors the recent JS and Flutter improvements that let users enable end-to-end encryption directly through `Session` initialization, instead of building a `Room` with `RoomOptions` + `EncryptionOptions` manually.

- `SessionOptions(encryption: .sharedKey("..."))` convenience init builds the underlying Room with the encryption options baked in.
- `EncryptionOptions.sharedKey(_:encryptionType:)` static factory for the common shared-passphrase case.
- `Session.setEncryptionEnabled(_:)` and `Room.setE2EEEnabled(_:)` for runtime toggling, matching the JS/Flutter API surface.
- Mutual exclusion of `room:` and `encryption:` is enforced at compile time via init overloads (no overload accepts both), rather than the runtime validation Flutter uses.
- Fully non-breaking: the existing `var room: Room` field and `init(room:)` are unchanged.

### DIFF
--- a/.changes/session-e2ee
+++ b/.changes/session-e2ee
@@ -1,0 +1,1 @@
+minor type="added" "Configure end-to-end encryption directly through Session initialization"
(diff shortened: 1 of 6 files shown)


xianshijing-lk (approved):
lgtm

PR #973, opened 2026-04-21
Build xcframework

pblazej:
This is quite tricky/manual process, as tools like https://github.com/segment-integrations/swift-create-xcframework won't work with newer SPM (that does not allow direct `xcodeproj` generation, required for archiving) 💀 

Install https://github.com/mxcl/swift-sh to handle dependencies.

The ultimate test:
- run `./scripts/xcframework.swift --local` 
- you should be able to replace the package with binaries for every platform

image

### Repo

This will land in https://github.com/livekit/client-sdk-swift-xcframework (with LICENSE, maybe copied README).

### Safety

This is mostly additive, except `LK_XCFRAMEWORK` for internal imports.


scripts/Package.swift.stencil
```diff
@@ -0,0 +1,44 @@
...
+        .macCatalyst(.v14),
+        .tvOS(.v17),
+        .visionOS(.v26),
+    ],
+    products: [
+        .library(name: "LiveKit", targets: ["LiveKitTargets"]),
+    ],
+    targets: [
+        .target(
+            name: "LiveKitTargets",
```
pblazej: That's another variant of "stub" target - just to include some source files.

Sources/LKObjCHelpers/module.modulemap
pblazej, line 1: Required of ObjC files are in the same target, should be ignored in plain SPM builds.

hiroshihorie (approved):
Just some things I want to check:
* Do we need to bump to .iOS(.v14) etc ?
* If platform archive fails does the script still continue to publish ?


Other than that LGTM

pblazej:
> Just some things I want to check:
> 
>     * Do we need to bump to .iOS(.v14) etc ?
> 
>     * If platform archive fails does the script still continue to publish ?
> 
> 
> Other than that LGTM

- `"IPHONEOS_DEPLOYMENT_TARGET": "14.0",` it was due to catalyst ☠️ conflict, I don't think there's a way to differentiate that via pure build settings, but I may be wrong - will double-check that
- this will be manual anyway, publishing xcframework will be fixed during uniffi deployment process soon, so we may re-apply that to webrtc + this xcframework; not sure if that answers your concern

pblazej:
@hiroshihorie the trick from uniffi seems to work here 09184da0fd3f1cdc4f086b97bde0ecc632f3df92

This project is only used for binary framework generation, so I guess it would work as-is, but now it's consistent - thanks for pointing it out 🎉

pblazej:
I also removed the stub target as it works on device (tested on fresh project):

Screenshot 2026-05-07 at 11 24 19 AM 5

PR #975, opened 2026-04-24
Data tracks

pblazej:
Adds [data tracks](https://github.com/livekit/rust-sdks/pull/1034) to the Swift SDK, implemented on top of the Rust data-track engine via UniFFI ([livekit-uniffi-xcframework 0.1.8](https://github.com/livekit/livekit-uniffi-xcframework/releases/tag/0.1.8)).

## API

```swift
// Publish + push
let track = try await room.localParticipant.publishDataTrack(name: "telemetry")
try track.tryPush(frame: DataTrackFrame(payload: data))          // non-blocking
try await track.send(contentsOf: frames)                          // AsyncSequence, drop-on-full by default

// Optionally declare what the frames are, and make the schema resolvable
let schema = DataTrackSchemaId(name: "telemetry.v1", encoding: .protobuf)
try await room.localParticipant.defineSchema(schema, definition: protoSource)
let typed = try await room.localParticipant.publishDataTrack(
    name: "typed",
    options: DataTrackPublishOptions(schema: schema, frameEncoding: .protobuf)
)

// Subscribe + receive
let stream = try await remoteTrack.subscribe(bufferSize: 16)
for await frame in stream { ... }                                 // DataTrackStream is an AsyncSequence

// A subscriber can resolve the schema its publisher declared
if let declared = remoteTrack.info.schema {
    let definition = try await room.localParticipant.getSchema(declared, publishedBy: remoteTrack.publisherIdentity)
}
```

Remote publish/unpublish surface on `RoomDelegate` and `ParticipantDelegate`; subscribed tracks are on `RemoteParticipant.dataTracks`, keyed by track name. Everything is Objective-C compatible (`DataTrackObjCTests` covers the surface).

## Design decisions

**Rust owns the protocol, Swift owns the session.** The UniFFI managers (`LocalDataTrackManager`/`RemoteDataTrackManager`) implement the publish/subscribe state machine, packetization, and E2EE. The SDK contributes what only it has: the signal connection, the WebRTC data channels, participant identity, and delegate fan-out. Concretely, a session-scoped `DataTracks` coordinator (one per `Room`, `Room+DataTrack.swift`) feeds SFU signal messages into the managers, forwards their outbound requests through `SignalClient` (serialized, so publish/unpublish ordering is preserved), and pumps packets between the managers and the DTP data channel.

**Thin public wrappers over the bindings.** The generated bindings are `internal import`ed and every public type (`LocalDataTrack`, `RemoteDataTrack`, `DataTrackFrame`, `DataTrackStream`, `DataTrackInfo`, schema/encoding types, error enums) is a small SDK-owned wrapper (~900 lines total). Tradeoff considered: exposing the generated types directly would save the layer, but pins the public API to regenerated code we don't control, and UniFFI emits Swift-only structs — `NSObject` wrappers are what makes the API reach Objective-C. The wrappers also carry the SDK idioms (lowerCamel error enums, `Track.Sid`-style identifiers, delegate naming, value semantics).

**Publication lifetime follows the handle (RAII).** The returned `LocalDataTrack` *is* the publication: releasing the last reference unpublishes it, as does calling `unpublish()`. This matches rust-sdks, where the handle drops the publication. Consequently there are no local publish/unpublish delegate events — Rust dispatches room events only for *remote* data tracks, and the handle (`isPublished`, `waitForUnpublish()`) is the local observer. `withDataTrack(name:body:)` scopes a publication to a block for the common case.

**Retain-cycle handling at the FFI boundary.** UniFFI callback interfaces hold their Swift delegate strongly (no weak references over FFI). A dedicated `ManagerDelegate` object references the `Room` and coordinator weakly, making it the designated weak link so `manager → delegate → room → manager` cycles can't leak (`Room+DataTrack.swift` documents the shape).

**Reconnect semantics.** Quick reconnect preserves publications via `SyncState.publishDataTracks` (the managers replay their publish responses); full reconnect republishes and re-asserts subscriptions. The `DataTracks` subsystem survives reconnects (channels are swapped in) and is torn down only on real disconnect, which also unpublishes remote tracks and notifies delegates. The publisher-channel readiness gate is re-armed — not failed — on teardown, so a `publishDataTrack` issued mid-reconnect waits for the new channel instead of racing a dead transport. A local full reconnect recreates participants but *not* their data tracks: those are detached silently and re-attached, so no spurious unpublish is reported.

When a *remote publisher* full-reconnects, the Rust manager treats the republication as a SID reassignment: the subscriber's existing `RemoteDataTrack` survives with its SID rewritten in place and active subscriptions transparently re-requested — no unpublish/republish events fire. This is why `RemoteParticipant.dataTracks` is keyed by **name**, not SID (Rust documents SIDs as unstable across reconnects; client-sdk-js keys by name for the same reason).

**Backpressure.** `tryPush` is non-blocking and throws `queueFull`, handing the rejected frame back so it can be retried (parity with Rust's `PushFrameError::into_frame`). `send(contentsOf:)` defaults to dropping frames when the queue is full (DTP is lossy by design — unbounded buffering would trade a dropped frame for unbounded latency). On the channel, `DataTrackFrameSender` meters packets out on buffered-amount events against an 8 KiB low-water mark, keeping send latency bounded while letting a frame of any size stream out: at most one frame waits while another drains, a newer frame evicts the waiting one (drop-oldest), and frames are handled whole so a partial frame is never left on the wire. This mirrors `DataChannelSender` in rust-sdks; client-sdk-js instead blocks the producer, which isn't available here (the producer is a fire-and-forget FFI callback with no backpressure channel). `DataTrackFrameSenderTests` pins these semantics, including the cross-SDK divergence. Subscribe-side buffer is caller-configurable (`subscribe(bufferSize:)`).

**E2EE.** When the room has E2EE configured, the same key provider drives data-track frame encryption via UniFFI `EncryptionProvider`/`DecryptionProvider` shims; `DataTrackInfo.usesE2ee` reflects it. Unlike data-channel payloads (a per-message property), data-track encryption is a track-level protocol property subscribers key their decryption on, so it can't be consulted per frame: the publishing manager is built on the *first publish* and captures the toggle then, which is late enough for a `setE2EEEnabled(true)` issued after connecting to apply. Toggling between publishes does not, and `Room.setE2EEEnabled` documents that. The cryptor resolves `Room.e2eeManager` on each call rather than capturing it at connect, so a manager assigned late is picked up for both encryption and decryption, and a missing one throws instead of silently sending plaintext.

## Protocol update

The `protocol` submodule moves v1.45.8 → v1.50.4 (regenerated with `make proto`). Two reasons:

- `DataTrackInfo` gained `frame_encoding` and `schema` after v1.45.8. Because `handleParticipantUpdate` re-serializes the parsed `ParticipantInfo` to hand the managers raw bytes, the older protocol silently stripped both fields before Rust saw them — declared metadata never reached subscribers. (The SFU was echoing it correctly; this was ours.)
- It brings `DataBlob`/`StoreDataBlobRequest`/`GetDataBlobRequest`, which `defineSchema`/`getSchema` are built on.

The general hazard remains and is worth a follow-up: any `ParticipantInfo` field newer than the pinned protocol is lost on the way to the managers. Threading the received wire bytes instead of re-serializing removes both the loss and a redundant encode.

## Size impact

App Size job measured **16.28 MB** for the SDK's uncompressed `.app` delta before the protocol update; the budget moves 16 → 16.5 MB to leave room for it. That figure includes the nanopb protocol migration from `main`, which cut ~1.8 MB; the data-track share is the 0.0.6 → 0.1.8 UniFFI bump plus this integration (~+0.4 MB: compiled bindings at the dead-strip floor plus the Rust dylib growth).

Isolating data tracks behind a Swift package trait was evaluated and rejected for now: the core SDK already depends on LiveKitUniFFI (tokens, log forwarding), both UniFFI components live in one dylib, and the bindings are one module — a trait could only gate the wrapper layer that dead-strip already removes. Revisit if the Rust side ships a feature-split artifact.

## Known upstream issue

`LocalDataTrack.tryPush` cannot surface the FFI's own error: `PushFrameErrorReason` is declared in the `livekit_datatrack` UniFFI component while `try_push` lives in `livekit_uniffi`, and the cross-component error lift fails, throwing `UniffiInternalError.bufferOverflow` instead. Left unhandled this leaked an FFI-internal type through the public API and broke `send(contentsOf:)`'s drop-on-full policy, which inspects the reason. The wrapper recovers the applicable reason from `isPublished`; the typed catch remains first and takes over once the bindings are fixed. Worth filing against livekit-uniffi.

## Not in this PR

- `tryPush` hands the rejected frame back, but the FFI still drops Rust's own `PushFrameError` payload.
- Remote pipeline options (`max_partial_frames`) aren't exported by UniFFI yet.
- Follow-ups filed from review: sharing the buffering state machine with `DataChannelPair`, and threading wire bytes through the signal path instead of re-serializing parsed join/participant updates.

## Testing

Two unit suites run without a server: `DataTrackFrameSenderTests` covers the outbound drain, and `DataTrackSendTests` covers `send(contentsOf:)`'s queue-full policy against a sink that rejects on demand — deliberately not by saturating a live pipeline, which made the outcome depend on how fast the SFU drained.

End-to-end tests against `livekit-server --dev` across six suites: publish/subscribe/roundtrip (small + multi-packet frames), publish options and schema/encoding metadata round-tripping through the SFU, schema definition storage and resolution, error cases (duplicate name, unauthorized, disconnected, push after unpublish, saturated queue), E2EE on, off, and toggled after connect, delegate events, unpublish paths, RAII handle release, `withDataTrack` scoping including cancellation, subscribe buffer clamping and drop-oldest eviction, concurrent multi-track push and 64-track publish, plus lifecycle: pre-join publications, publish during a full reconnect, publisher full-reconnect SID reassignment, quick-reconnect sync state, remote tracks surviving the local client's full reconnect, and room moves. `DataTrackObjCTests` covers the Objective-C surface including NSError bridging and value semantics.

`DataTrackStream` only ends when its track is unpublished, so every read goes through bounded `next(within:)`/`collect(_:)` helpers — a lost frame on an unreliable channel should fail a test, not hang the job. The repeated two-room publish/subscribe preamble lives in a `withPublishedDataTrack` fixture; tests that control *when* the publish happens keep driving `withRooms` directly, since that timing is what they exercise.

Schema definitions are stored as participant data blobs, which the server keeps behind `enable_participant_data_blob` — enabled for CI, and documented in `AGENTS.md` for local runs.


Sources/LiveKit/Core/Room+DataTrack.swift
```diff
@@ -0,0 +1,230 @@
...
+/// Owns the local/remote data track managers, the data channels, and the manager-delegate shim,
+/// and routes Room/participant calls to the right manager. The ``Room`` holds a single reference,
+/// keeping the subsystem off the Room's surface. Session-scoped: created at connect and kept across
+/// reconnects (its channels are swapped, but the managers persist so publications can be
+/// republished), released only on a real disconnect.
+///
+/// `@unchecked Sendable`: the only mutable state is `_publisherChannel` (StateSync-guarded); the
+/// managers and delegate are immutable after init. Not an actor — the UniFFI delegate callbacks
+/// are synchronous and can't `await`.
+final class DataTracks: NSObject, @unchecked Sendable {
```
pblazej, line 33: @1egoman an attempt to encapsulate the whole rust "subsystem".

Sources/LiveKit/Participant/LocalParticipant+DataTrack.swift
```diff
@@ -0,0 +1,74 @@
...
+        room.delegates.notify(label: { "room.didPublishDataTrack" }) {
+            $0.room?(room, participant: self, didPublishDataTrack: track)
+        }
+
+        // Retain the publication until it's unpublished, so callers aren't forced to hold the
+        // returned track to keep it published (the underlying handle unpublishes on drop). Matches
+        // JS, where the manager owns the publication. The wait ends on an explicit/SFU unpublish or
+        // teardown, and does NOT fire during a reconnect's republish — so session-scoped republish
+        // still works, and the unpublish delegate fires once when the publication really ends.
+        Task { [weak self, weak room] in
```
pblazej: This is where RAII semantics gets controversial, the big picture: 3e3043ec

pblazej:
@1egoman @ladvoc that may not be production ready yet (failing tests here and there etc.), so just opening the discussion 🎉 

Some interesting review vectors:
- the amount of wrapping required for `@objc`/`NSObject`
- the raii thing above - is it _really_ the right choice for swift?

Sources/LiveKit/Core/Room+DataTrack.swift
```diff
@@ -0,0 +1,230 @@
...
+
+    var publisherChannel: LKRTCDataChannel? { _publisherChannel.copy() }
+
+    init(room: Room) {
+        let managerDelegate = ManagerDelegate(room: room)
+        // Provide the cryptor only when E2EE is configured — its presence is what marks tracks as
+        // encrypted (usesE2ee), so it must be nil otherwise.
+        let cryptor: DataTrackCryptor? = room.e2eeManager.map(DataTrackCryptor.init)
+        local = LocalDataTrackManager(delegate: managerDelegate, encryptionProvider: cryptor)
+        remote = RemoteDataTrackManager(delegate: managerDelegate, decryptionProvider: cryptor)
```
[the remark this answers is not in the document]
pblazej: Boils down to missing DI structure on Room, will work on that 👍

Sources/LiveKit/Core/Room+DataTrack.swift
```diff
@@ -0,0 +1,327 @@
...
+            guard let channel = coordinator?.publisherChannel else { return }
+            let buffers = packets.map { RTC.createDataBuffer(data: $0) }
+            DispatchQueue.liveKitWebRTC.async { [weak room] in
+                // ponytail: drop the whole frame when the channel is congested. The DTP channel is
+                // unreliable, so dropping beats unbounded buffering; switch to drop-oldest if the
+                // newest-frame loss becomes a problem.
+                guard channel.bufferedAmount < Self.maxBufferedAmount else {
+                    room?.log("Data track channel congested (\(channel.bufferedAmount) bytes buffered), dropping frame", .warning)
+                    return
+                }
```
1egoman: nitpick: this isn't actually dropping a frame necessarily, it's dropping a series of packets. This right now I think is likely to be the packets that make up a frame but that's not guaranteed, so maybe update this error message?

Sources/LiveKit/Core/Room+DataTrack.swift
```diff
@@ -0,0 +1,327 @@
...
+        func onTrackPublished(track: LiveKitUniFFI.RemoteDataTrack) {
+            coordinator?.remoteTrackPublished(RemoteDataTrack(track))
+        }
+
+        func onTrackUnpublished(sid: LiveKitUniFFI.DataTrackSid) {
+            coordinator?.remoteTrackUnpublished(sid: DataTrack.Sid(from: sid))
+        }
+
+        // Bound the publisher data track channel buffer; parity with the lossy data channel threshold.
+        private static let maxBufferedAmount: UInt64 = 2 * 1024 * 1024
```
1egoman: Important caveat worth considering: If I'm understanding this properly, this is effectively right now the absolute max size of a data track frame. On the web this is an unbounded buffer right now because we still haven't made an official decision on the maximum data track frame size yet.

This also relates to some of the issues which we ran into on the web with integrating data tracks - on the web right now, data tracks occupy this weird place in between "reliable" and "lossy" where it's a lossy data channel but one where packets are buffered like the reliable data channel. We've since cleaned it up a lot from when I originally surfaced this a few months back, you might find this PR an interesting read: https://github.com/livekit/client-sdk-js/pull/2014. The high water mark / low water mark approach for buffer sending I think is also something that should make its way into other sdks too eventually.

Sources/LiveKit/DataTrack/LocalDataTrack.swift
```diff
@@ -0,0 +1,92 @@
...
+            } catch let error as DataTrackPushFrameError {
+                guard case .queueFull = error else { throw error }
+                switch onQueueFull {
+                case .throw: throw error
+                case .drop: continue
+                }
+            }
+        }
+    }
+}
```
1egoman, line 129: Ah this is pretty clever, so you give the user access to control whether non queueable messages drop or throw.

Two thoughts:
1. Is this always something you would want to expose? Or is this transitional until we come to a decision on the max data track frame size / the performance bounds which data tracks are expected to support?
2. Should the default be throw or drop? I would argue throw maybe makes more sense, or at a minimum, drop should log when it drops so a user isn't surprised when things break unexpectedly if they send a large data track frame.

cc @lukasIO for when you are back, this might be worth doing in web as well 🤔

Sources/LiveKit/DataTrack/RemoteDataTrack.swift
```diff
@@ -0,0 +1,55 @@
...
+    @objc public func subscribe(bufferSize: UInt32 = 16) async throws -> DataTrackStream {
+        do {
+            let options = LiveKitUniFFI.DataTrackSubscribeOptions(bufferSize: bufferSize)
+            let stream = try await track.subscribeWithOptions(options: options)
+            return DataTrackStream(stream)
+        } catch let error as LiveKitUniFFI.DataTrackSubscribeError {
+            throw DataTrackSubscribeError(error)
+        }
+    }
+}
```
1egoman, line 75: thought: Should this be async? What we did in the web (which worked quite well) was to immediately return the `DataTrackStream` equivalent object from `subscribe()` and lazily initialize the underlying sfu subscription on the first stream iteration `.next()` type call. The benefit to doing it this way is it centralized all error handling in one path (both stream read errors and stream initialization errors). Another side benefit was it abstracted away the actual "subscribe" call so it could be run initially or later after a full reconnect all transparently to a user.

Sources/LiveKit/Participant/LocalParticipant+DataTrack.swift
```diff
@@ -0,0 +1,74 @@
...
+    /// - Returns: The value returned by `body`.
+    func withDataTrack<T>(name: String, body: (LocalDataTrack) async throws -> T) async throws -> T {
+        let track = try await publishDataTrack(name: name)
+        return try await withTaskCancellationHandler {
+            defer { track.unpublish() }
+            return try await body(track)
+        } onCancel: {
+            track.unpublish()
+        }
+    }
```
1egoman, line 103: This is something else I think could be good to have on the web 🤔

Sources/LiveKit/Participant/LocalParticipant+DataTrack.swift
1egoman: I am not sure I fully understand the reason this has to be done - I would assume that a user would need to keep a `LocalDataTrack` reference alive so they can call `tryPush` / etc on it? So if it drops, then unpublishing should always be the right thing to do?

I think I might need to see a small example to fully wrap my head around it, if that isn't too hard to put together.

1egoman (approved):
Generally makes sense to me from a high level / as a non swift expert, I think it would be good to get a swift expert to go through this as well

Sources/LiveKit/DataTrack/LocalDataTrack.swift
pblazej: My initial idea was more of a "pipe" between (buffered) `AsyncSequence` and data track, the backpressure (default) is a follow-up, no real opinion/example here.

Sources/LiveKit/DataTrack/RemoteDataTrack.swift
pblazej: So now (without reading your code), these errors would propagate just from the stream? `try await track.subscribeWithOptions`?

Sources/LiveKit/Participant/LocalParticipant+DataTrack.swift
pblazej: Yes if you "produce"/yield the frames inside. Take these `async` helpers with a grain of salt tho as they've got no equivalent for other tracks.

Sources/LiveKit/Participant/LocalParticipant+DataTrack.swift
pblazej: Yes that's just a design decision, we can just revert it basically and it will work as you described.

Sources/LiveKit/Core/Room+DataTrack.swift
pblazej: Yup

pblazej:
@1egoman @ladvoc a variant of buffering, similar to our existing `DataChannel` infrastructure 91e9203ebae53d606c98bbc4c283416a51b78508

Semantic similarities/diffs between JS/Rust indicated in tests.

I also see some potential extraction to `FlowControlledDataChannel` 👍 

BTW this logic could be moved to Rust as well 🙃

Sources/LiveKit/DataTrack/RemoteDataTrack.swift
1egoman: Yes, they would propagate from the stream. For example, [here's](https://docs.livekit.io/transport/data/data-tracks/#reading-frames) how it works on the web currently (unrolling the `for await` to make it a little easier to comprehend):

```ts
const stream = track.subscribe();

while (true) {
  const { value, done } = await stream.next();
  if (done) {
    break;
  }
  console.log('Received frame:', value.payload);
}
```

The important thing is that this doesn't do `await track.subscribe()`, the only `await` happens at the `stream.next()` level.

1egoman: Also, to be clear - not trying to argue for / against doing this in swift, only that it turned out to work quite well in js and consistency might be nice.

Sources/LiveKit/Participant/LocalParticipant+DataTrack.swift
1egoman: Other datapoint I think is relevant: in the rust-sdk's `livekit` crate exposed interface, calling `drop(localDataTrack)` will unpublish. This was done intentionally and in early prototypes there wasn't even an explicit `.unpublish()` method on local data tracks in rust. So there's precedent for this behavior I think and not trying to keep local data tracks alive / circumventing RAII type semantics.

Sources/LiveKit/Participant/LocalParticipant+DataTrack.swift
```diff
@@ -0,0 +1,74 @@
...
+            $0.room?(room, participant: self, didPublishDataTrack: track)
+        }
+
+        // Retain the publication until it's unpublished, so callers aren't forced to hold the
+        // returned track to keep it published (the underlying handle unpublishes on drop). Matches
+        // JS, where the manager owns the publication. The wait ends on an explicit/SFU unpublish or
+        // teardown, and does NOT fire during a reconnect's republish — so session-scoped republish
+        // still works, and the unpublish delegate fires once when the publication really ends.
+        Task { [weak self, weak room] in
+            await track.waitForUnpublish()
```
ladvoc: note: This will also resolve when the track has been unpublished due to the manager having been shutdown. If the manager is shutdown before the room has been destroyed, this will resolve, match some for `self` and `room`, and fire the delegate events.

Sources/LiveKit/DataTrack/RemoteDataTrack.swift
pblazej: No hard feelings here, it's just not very natural to make subscribe non-throwing and move it to the stream @lukasIO any bias here?

Sources/LiveKit/Participant/LocalParticipant+DataTrack.swift
pblazej: I think it's way cleaner to just revert this ugly task-leak and keep the raii/retain approach (without edge cases).

Sources/LiveKit/DataTrack/RemoteDataTrack.swift
ladvoc: I agree, I think this question is somewhat based on language conventions, and as a Swift dev it would feel weird to me if the error was deferred. Same with Rust which has the subscribe method return a result.

pblazej: Yeah I'm more on the Rust side here.

ladvoc:
> LocalDataTrack.tryPush cannot surface the FFI's own error

Let's discuss this more, would like to better understand what is going on.

Sources/LiveKit/Core/Room+DataTrack.swift
```diff
@@ -0,0 +1,399 @@
...
+        // building the manager here would fix the encryption mode before the first publish.
+        if let local = _local.copy() {
+            try? local.handleSfuRequestResponse(res: data)
+            try? local.handleSfuPublishResponse(res: data)
+        }
+        try? remote.handleSubscriberHandles(res: data)
+    }
+
+    func handleParticipantUpdate(_ participants: [Livekit_ParticipantInfo], localIdentity: String) {
+        // The FFI takes raw bytes, so rebuild the response from the parsed participants.
```
ladvoc: question: When I designed the bindings for this, I assumed there would be a way to get the raw, encoded message from the signal client and pipe to here. Worth considering an alternative design to avoid re-serializing? This is not a particularly high frequency code path though, so maybe not?

Sources/LiveKit/DataTrack/DataTrackError.swift
```diff
@@ -0,0 +1,138 @@
...
+    }
+}
+
+/// The reason a frame could not be pushed via ``LocalDataTrack/tryPush(frame:)``.
+public enum DataTrackPushFrameError: Error, Sendable {
+    /// The track has been unpublished, by either the local participant or the SFU.
+    case trackUnpublished(String)
+    /// The send queue is full; the frame was not enqueued. It is handed back so the caller can
+    /// retry or re-queue it rather than losing it.
+    case queueFull(String, frame: DataTrackFrame)
```
ladvoc, line 64: question: In the Rust version, since `try_push` consumes the frame, we can return ownership of the same struct (same underlying memory) in the error variant without making a copy or reference counting. Does doing this in Swift have the same benefit?

Sources/LiveKit/DataTrack/DataTrackFrame.swift
```diff
@@ -0,0 +1,96 @@
...
+import Foundation
+
+internal import LiveKitUniFFI
+
+/// A single unit of application data sent or received over a data track.
+public final class DataTrackFrame: NSObject, Sendable {
+    /// The application payload carried by this frame.
+    @objc public let payload: Data
+
+    /// Optional sender-provided timestamp, in milliseconds since the Unix epoch.
```
ladvoc: suggestion: Technically, this does not have to be milliseconds since the Unix epoch. The helpers (i.e., `durationSinceTimestamp` and `now`) are good to have for when it is, however, an application, for example, may use this to attach a timestamp provided by a sensor that uses a different clock that both sides of the application (publisher and subscriber) agree on.

Sources/LiveKit/DataTrack/LocalDataTrack.swift
```diff
@@ -0,0 +1,124 @@
...
+        }
+    }
+
+    /// Unpublishes the track. Subsequent ``tryPush(frame:)`` calls throw.
+    @objc public func unpublish() {
+        track.unpublish()
+    }
+
+    /// Waits until the track is unpublished, by either the local participant or the SFU.
+    @objc public func waitForUnpublish() async {
```
ladvoc, line 72: suggestion: Might be worth adding these points from the equivalent Rust doc comment:
> Use this to trigger follow-up work once the track is no longer published. 
If the track is already unpublished, this method returns immediately.

A useful prompt to Claude I've found for other FFI work is to have it look at the Rust implementation and make sure doc comments for all public APIs align (where applicable).

Sources/LiveKit/DataTrack/LocalDataTrack.swift
```diff
@@ -0,0 +1,124 @@
...
+        await track.waitForUnpublish()
+    }
+}
+
+// MARK: - Sending an AsyncSequence
+
+/// The slice of a publication the sequence send drives — a seam so the queue-full policy is
+/// unit-testable, since saturating a live pipeline to observe it is inherently timing-dependent.
+protocol DataTrackFrameSink: AnyObject, Sendable {
```
ladvoc, line 81: question: Would this only ever be used in a unit testing context?

Sources/LiveKit/DataTrack/RemoteDataTrack.swift
```diff
@@ -0,0 +1,72 @@
...
+        super.init()
+    }
+
+    /// Whether the track is currently published by the remote participant.
+    @objc public var isPublished: Bool { track.isPublished() }
+
+    /// Metadata for this track.
+    @objc public var info: DataTrackInfo { DataTrackInfo(track.info()) }
+
+    /// Waits until the track is unpublished, by either the publisher or the SFU.
```
ladvoc, line 49: suggestion: Same suggestion as for the same method on `LocalDataTrack`.

Sources/LiveKit/Extensions/TimeInterval.swift
```diff
@@ -36,6 +36,8 @@ public extension TimeInterval {
     static let defaultPublisherDataChannelOpen: Self = 15
     static let resolveSid: Self = 7 + 5 // Join response + 5
     static let defaultPublish: Self = 10
+    // Matches client-sdk-rust (`DATA_BLOB_REQUEST_TIMEOUT`).
+    static let defaultDataBlobRequest: Self = 5
```
ladvoc, line 40: note: I would like to discuss with the team more generally how we can standardize behavior for signaling messages with request/response semantics. Let's leave it consistent with Rust for now.

Sources/LiveKit/Participant/LocalParticipant+DataTrack.swift
ladvoc: Here's my understanding, @pblazej correct me if I'm wrong. In GC languages, you don't want to rely on destructors for RAII since there is a non-deterministic timing gap between when the object is no longer referenced and when its destructor is run. Swift uses ARC (automatic reference counting) rather than GC, which means objects are destroyed as soon as the reference count reaches zero. The difference between this and Rust is there is not a 1:1 correlation with scope, so it couldn't be used for something like a mutex guard or accessing raw memory. For the latter, Switch uses this closure pattern to ensure you can only access a pointer within the scope of a closure. However, for the data track use case, do we really need publication to be linked exactly to scope? Maybe relying on `deinit` is ok in this case?

ladvoc (approved):
LGTM, amazing work @pblazej! Left a few comments/suggestions, but nothing blocking.

Sources/LiveKit/DataTrack/DataTrackFrame.swift
pblazej: Done in `a5e8265f` — the docs now describe `userTimestamp` as opaque and carried verbatim, with the Unix-milliseconds reading confined to `now(payload:)` and `durationSinceTimestamp`. The Objective-C names had the assumption baked in too, so `userTimestampMs` → `userTimestamp` and `initWithPayload:userTimestampMs:` → `initWithPayload:userTimestamp:`.

Sources/LiveKit/DataTrack/LocalDataTrack.swift
pblazej: Done in `a5e8265f`. Also took the wider suggestion and did a pass against the Rust docs, which turned up two more gaps: `tryPush` now mentions that a disconnected room surfaces as a push failure, and `publishDataTrack` carries your note about a timeout hinting at an SFU release without data track support when self-hosting.

Sources/LiveKit/DataTrack/RemoteDataTrack.swift
pblazej: Done in `a5e8265f`.

Sources/LiveKit/DataTrack/LocalDataTrack.swift
pblazej: yes

Sources/LiveKit/Core/Room+DataTrack.swift
pblazej: Done in `da727813` — the response queue now carries each message alongside the bytes it arrived as, handed over via a new `EncodedResponse` case, so neither the join nor the participant update is re-encoded. Worth doing for correctness rather than cost: re-encoding our decoded copy dropped every field the pinned protocol did not know (nanopb discards unknown fields), which is exactly how declared schema and frame encoding went missing before the protocol bump — room moves are the one path still building a message, since the SFU sends a `RoomMovedResponse` while the manager consumes a `ParticipantUpdate`.

Sources/LiveKit/DataTrack/DataTrackError.swift
pblazej: No — `DataTrackFrame` is a class (it has to be, for ObjC parity), so it is reference-counted regardless, and `tryPush` borrows rather than consumes it, so a direct caller never lost their reference in the first place. It is also moot at the boundary: `FfiConverterData.write` copies the payload into the RustBuffer before Rust sees the frame, so the memory Rust hands back zero-copy is already duplicated on our side. Clarified in `95ab27ce` that it is the same instance rather than an ownership transfer, and that its real consumer is `send(contentsOf:)`, where frames arrive from a sequence and the caller holds no reference.

pblazej:
Thanks @ladvoc for having an in-depth look! 2 highlights:
- RAII/memory management - I don't think this pattern is _wrong_ per se, it's more about consistency with other APIs, surprising behavior/side effects; we can treat it as "low-level" difference between SDKs/langs, simialarly to cancellation semantics, etc.
- (Re)encoding - yes I think it can be improved like this (da727813071cc6e1844cffa94caf26c201b3e8a9) which was my initial idea, probably got lost after multiple iterations
  - it's just a separate notification path, nothing is enforced here (to handle raw messages for all paths etc.)

Sources/LiveKit/Core/SignalClient.swift
```diff
@@ -350,7 +417,8 @@ private extension SignalClient {
             _delegate.notifyDetached { await $0.signalClient(self, didReceiveIceCandidate: rtcCandidate.toLKType(), target: trickle.target) }
 
         case let .update(update):
-            _delegate.notifyDetached { await $0.signalClient(self, didUpdateParticipants: update.participants) }
+            try? await _delegate.notifyAsync { await $0.signalClient(self, didUpdateParticipants: update.participants) }
+            try? await _delegate.notifyAsync { await $0.signalClient(self, didReceiveEncodedResponse: .participantUpdate(encoded)) }
```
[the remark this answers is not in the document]
pblazej: Good catch — fixed in `3d05b708`. The ordering only needs the two notifications *enqueued* in order, not awaited, so `AsyncSerialDelegate` gained `notifyDetached(inOrder:)`, which runs them from a single detached task awaiting each in turn: `_process` returns immediately as before, so participant cleanup and the app's own delegates are off the response queue's critical path, and the connect completer no longer waits behind the join handler either. A `notifyDetached` per closure cannot substitute for this, since each races to the serial runner in its own task — an undocumented trap until now, so `AsyncSerialDelegateTests` pins the guarantee (50 concurrent batches with per-batch order asserted, plus a slow first notification that must not be overtaken). Worth noting the data track e2e tests cannot cover this: park-and-reattach masks a violation, which I confirmed by breaking the ordering deliberately and watching them still pass.

Sources/LiveKit/Core/Room+DataTrack.swift
```diff
@@ -0,0 +1,399 @@
...
+    }
+
+    func handleRoomMoved(_ participants: [Livekit_ParticipantInfo], localIdentity: String) {
+        // The old room's tracks are gone for good. Dropped through the unpublish path rather than
+        // cleared outright, so anything the participant teardown didn't already report still
+        // reaches the delegates (and no-ops for participants it already removed). Then republish
+        // local tracks into the new room and surface its existing publications.
+        for track in _remoteTracks.copy() {
+            remoteTrackUnpublished(sid: track.info.sid)
+        }
```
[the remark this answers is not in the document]
pblazej: Confirmed and fixed in `d90dbf99` — `_addNewParticipants` does run before `handleRoomMoved`, so the lookup succeeded for any publisher that moved with us and the unpublish fired twice. The move now forgets the parked tracks silently rather than routing them through the unpublish path: a track still parked was never attached, so no publish was reported for it either, and the teardown has already reported the rest. My room-move test had been driving the coordinator without the teardown, so it was asserting the very event it should have forbidden — it now asserts the absence, and I checked it fails against the old shape.

Sources/LiveKit/Core/Room+DataTrack.swift
```diff
@@ -0,0 +1,399 @@
...
+        // silently dropped (`sendData` on a non-open channel fails). Gate failures are reported
+        // as publish errors, since that's what the caller asked for.
+        do {
+            try await room?.ensurePublisherConnected()
+            try await _publisherChannelOpen.wait()
+        } catch let error as LiveKitError where error.type == .timedOut {
+            throw DataTrackPublishError.timeout("Timed out establishing the publisher data track channel")
+        } catch {
+            throw DataTrackPublishError.disconnected("Lost the connection while establishing the publisher data track channel")
+        }
```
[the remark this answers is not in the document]
pblazej: Fixed in `eb49f7f3`. Both cancellation paths — `LiveKitError(.cancelled)` from `AsyncCompleter.wait` and a propagated `CancellationError` — are now rethrown unchanged ahead of the catch-all, so a cancelled publish no longer looks like a lost connection; documented on `publishDataTrack` since it widens what the call can throw beyond `DataTrackPublishError`. Laundering cancellation into a domain error seemed worse than the doc change, especially as `withDataTrack` exists to be cancellable.

PR #977, opened 2026-04-28
Support for large RPC messages using data streams

1egoman:
See this pull request for more info: https://github.com/livekit/client-sdk-js/pull/1832

@pblazej Sent you a slack message about this with some of the context!


pblazej:
The general direction looks good, need a deeper review for possible breaking changes, etc.

1egoman:
@pblazej @hiroshihorie Two things:

- I might need some help getting the ci to pass for this. Looking at the [test failures](https://github.com/livekit/client-sdk-swift/actions/runs/25077832800/job/73475117539), it looks like the test failures differ each time, and in some targets the tests pass fully while others they don't (and the same tests should be running everywhere...) - are the tests flaky?

- I have run this implementation through the rpc testing app I built which exercises a given RPC v2 implementation through all the test cases and for the cases that are easy to test, it passes fully! 🎉 . So I think this is in a good state to be reviewed at this point.

Sources/LiveKit/Protos/livekit_models.pb.swift
pblazej, line 1: The protocol version went out of sync with the submodule, have you generated that manually?

Sources/LiveKit/Types/Options/ConnectOptions.swift
```diff
@@ -74,6 +74,12 @@ public final class ConnectOptions: NSObject, Sendable {
     /// LiveKit server protocol version to use. Generally, it's not recommended to change this.
     public let protocolVersion: ProtocolVersion
 
+    /// Client-to-client protocol version advertised to other participants.
+    ///
+    /// Defaults to ``ClientProtocol/v1``, which enables RPC v2 (data-stream-based payloads
+    /// with no 15 KB size limit). Generally, it's not recommended to change this.
+    public let clientProtocol: ClientProtocol
```
pblazej, line 81: Hm, just thinking out loud: should we even expose this? (applies to `protocol` as well `Generally, it's not recommended to change this.`)

Sources/LiveKit/Core/RpcServerManager.swift
```diff
@@ -0,0 +1,283 @@
...
+    {
+        guard let handler = handlers[method] else {
+            return .failure(RpcError.builtIn(.unsupportedMethod))
+        }
+
+        do {
+            let response = try await handler(RpcInvocationData(requestId: requestId,
+                                                               callerIdentity: callerIdentity,
+                                                               payload: payload,
+                                                               responseTimeout: responseTimeout,
```
pblazej: P0: Trailing comma after `responseTimeout: responseTimeout,` is SE-0439 (Swift 6.1+) — won't build on Xcode 16.2 / Swift 6.0 (our floor) 🥲. A `swiftformat`/`swiftlint` run with the project's pinned `--swiftversion 6.0` should clean it up.

Sources/LiveKit/Core/RpcClientManager.swift
```diff
@@ -0,0 +1,246 @@
...
+            try await publishRequest(in: room,
+                                     destinationIdentity: destinationIdentity,
+                                     requestId: requestId,
+                                     method: method,
+                                     payload: payload,
+                                     responseTimeout: effectiveTimeout)
+        }
+
+        do {
+            return try await withThrowingTimeout(timeout: responseTimeout) {
```
pblazej: P0: Pending state is registered _after_ publish (lines 60-67 publish, 83-84 register inside the inner `Task`) — a fast remote can ack/respond before `addPendingAck`/`setPendingResponse` run, the response then no-ops with `Response received for unexpected RPC request`, and the call hangs to the outer `responseTimeout`.

JS sets pending state synchronously _before_ publishing (`RpcClientManager.ts:89-103`). Since `performRpc` is already actor-isolated here, we can move the calls inline (no inner `Task`):

```swift
pendingAcks.insert(requestId)
let completer = AsyncCompleter<String>(label: "rpc-\(requestId)", defaultTimeout: responseTimeout)
pendingResponses[requestId] = PendingRpcResponse(
    participantIdentity: destinationIdentity,
    onResolve: { [weak self] payload, error in
        Task { [weak self] in
            await self?.removeAllPending(requestId)
            if let error { completer.reject(error) } else { completer.resume(returning: payload ?? "") }
        }
    }
)

if useStreamTransport { try await publishRequestStream(...) }
else                  { try await publishRequest(...) }

return try await completer.wait()
```

`AsyncCompleter` from `Support/Async/` is single-resolution by construction — also fixes P0 below for free.

Unit-testable with a small internal `__test_hookAfterPublish` on the actor that runs a closure between publish and registration.

Sources/LiveKit/Core/RpcClientManager.swift
```diff
@@ -0,0 +1,246 @@
...
+                            }
+                        ))
+                    }
+
+                    Task { [weak self] in
+                        try await Task.sleep(nanoseconds: UInt64(maxRoundTripLatency * 1_000_000_000))
+                        guard let self else { return }
+                        if await self.hasPendingAck(requestId) {
+                            await self.removeAllPending(requestId)
+                            continuation.resume(throwing: RpcError.builtIn(.connectionTimeout))
```
pblazej: P0: This `continuation.resume(throwing: …)` plus the one inside `onResolve` above can both fire — if ack is delayed but the response arrives, the response Task and the ack-timeout Task both call `resume` on the same `CheckedContinuation` → `fatalError`.

Counterexample timeline:
```
T=6.999s  ack-timeout reads `hasPendingAck(id)` → true
T=7.000s  response arrives, onResolve schedules Task → continuation.resume(returning:)
          ack-timeout proceeds → continuation.resume(throwing:)
T=7.001s  💥
```

JS guards with optional-chained `completionFuture.resolve?.()` so the second resolve is a silent no-op. The `AsyncCompleter` swap above fixes this for free (already idempotent). Unit-testable with an internal `__test_forceAckTimeout(requestId:)` hook so we don't have to wait 7 s in tests.

Sources/LiveKit/Core/RpcServerManager.swift
```diff
@@ -0,0 +1,283 @@
...
+                                                               ))
+            if response.byteLength > MAX_RPC_PAYLOAD_BYTES {
```
pblazej: P0: This 15 KB cap is applied unconditionally, but `dispatchToHandler` is shared between v1 packet path _and_ v2 stream path 🥲 — so a v2 handler returning >15 KB silently maps to `responsePayloadTooLarge`, which contradicts both the docstring on this function and the comment on `MAX_RPC_PAYLOAD_BYTES` in `RPC.swift`.

JS keeps the cap inside the v1 publish path only (`RpcServerManager.ts:277-291`) — moving the check into `publishResponse` here mirrors that:

```swift
private func publishResponse(in room: Room, …, payload: String?, error: RpcError?) async throws {
    let outgoing: (String?, RpcError?)
    if let payload, payload.byteLength > MAX_RPC_PAYLOAD_BYTES {
        outgoing = (nil, RpcError.builtIn(.responsePayloadTooLarge))
    } else {
        outgoing = (payload, error)
    }
    // … build packet from `outgoing`
}
```

Unit test sketch (currently fails, passes with the fix):
```swift
@Test func v2HandlerCanReturnLargeResponse() async throws {
    try await TestEnvironment.withRoom { room in
        try await confirmation("v2 stream response sent", expectedCount: 1) { sentStream in
            let mock = MockDataChannelPair { packet in
                if case let .streamHeader(h) = packet.value, h.topic == RpcStreamTopic.response { sentStream() }
                if case .rpcResponse = packet.value { Issue.record("must not fall back to v1 packet") }
            }
            room.publisherDataChannel = mock
            try await room.registerRpcMethod("echo") { _ in String(repeating: "y", count: 20_000) }
            // … drive handleIncomingRequestStream …
        }
    }
}
```

Sources/LiveKit/Core/Room+DataStream.swift
```diff
@@ -58,6 +60,20 @@ public extension Room {
     }
 }
 
+extension Room {
+    private static let reservedStreamTopics: Set<String> = [
```
pblazej: P1: Could we make `lk.` a global reserved prefix instead of a hand-maintained set? Server convention already documents `lk.` as LiveKit's namespace, and a single `topic.hasPrefix("lk.")` check (applied on register _and_ unregister 👇) would cover any future internal topic without piecemeal additions:

```swift
static func checkReserved(topic: String) throws {
    guard !topic.hasPrefix("lk.") else {
        throw LiveKitError(.invalidParameter,
                           message: "Stream topic prefix 'lk.' is reserved for internal SDK use")
    }
}
```

Worth applying to `unregisterByteStreamHandler`/`unregisterTextStreamHandler` as well — currently a user can call `unregisterTextStreamHandler(for: "lk.rpc_request")` and silently disable our internal RPC v2 dispatch.

It is technically a breaking change for users who registered on `lk.*` topics, but they were against the convention anyway — single CHANGELOG line.

pblazej (commented):
Generally a fan of the split into `RpcClientManager`/`RpcServerManager` + `clientProtocol` negotiation — wire format matches the JS reference (`rpc-with-compression-data-streams`) byte-for-byte 🎉. There's a P0 build break on Xcode 16.2 (trailing comma) plus a few concurrency bugs around the caller-side pending state (race + potential double-resume on `CheckedContinuation`); also a 15 KB cap accidentally being applied on the v2 response path. Inline comments below cover those plus a bunch of test-side polish — should be a much tighter v2 once those are addressed.

Sources/LiveKit/Core/Room.swift
```diff
@@ -258,6 +259,25 @@ public class Room: NSObject, @unchecked Sendable, ObservableObject, Loggable {
             await metricsManager.register(room: self)
         }
 
+        // Wire RPC managers and register internal handlers for RPC v2 data streams. Topics
```
pblazej: P1: There's a race window from `Room.init` returning until this Task fires — `rpcClient`/`rpcServer` aren't attached yet and the internal handlers aren't registered, so any incoming v2 stream during that window is dropped silently 🥲. Also any failure here is just logged.

We can't `await` from `init` directly (NSObject sync init), but moving the attach + `registerTextStreamHandler` calls into the existing async `connect()` flow would close the window before the engine starts producing data — and let registration failures actually surface to the user.

Sources/LiveKit/Core/RpcClientManager.swift
```diff
@@ -0,0 +1,246 @@
...
+
+import Foundation
+
+/// Caller-side RPC.
+///
+/// Owns the in-flight bookkeeping (pending acks and pending responses) and the wire-level
+/// publishing of v1 RPC packets / v2 RPC request streams. `LocalParticipant.performRpc`
+/// is a one-line proxy that forwards into this actor.
+actor RpcClientManager: Loggable {
+    private weak var room: Room?
```
pblazej, line 25: P1: We never reap pending RPCs when a participant disconnects — caller waits up to `responseTimeout` to learn the recipient went away, vs JS surfacing `recipientDisconnected` immediately (`RpcClientManager.ts:242-256`). Could wire from `Room.engine(didMutateState:)` and `Room.cleanUp(...)` → new `handleParticipantDisconnected(_:)` that walks `pendingResponses` filtered by `participantIdentity`.

Sources/LiveKit/Core/RpcClientManager.swift
```diff
@@ -0,0 +1,246 @@
...
+                                           responseTimeout: effectiveTimeout)
+        } else {
+            try await publishRequest(in: room,
+                                     destinationIdentity: destinationIdentity,
+                                     requestId: requestId,
+                                     method: method,
+                                     payload: payload,
+                                     responseTimeout: effectiveTimeout)
+        }
+
```
pblazej, line 75: P1: Outer `withThrowingTimeout` doesn't clear `pendingResponses` when it fires — so a late response after timeout still calls `pending.onResolve` against an already-resumed continuation (another double-resume vector). The `AsyncCompleter` rewrite suggested above closes this for free; otherwise we need an explicit cleanup branch.

Tests/LiveKitCoreTests/RpcTests.swift
```diff
@@ -193,4 +195,407 @@ struct RpcTests {
             }
         }
     }
+
```
pblazej, line 46: P1: We're missing a few cases that would catch the issues flagged inline above:
- response/ack delivered _before_ registration completes (P0 race)
- late response _after_ ack-timeout fires (P0 double-resume)
- v2 handler returning >15 KB (P0 cap)
- `unregisterTextStreamHandler(for: "lk.rpc_request")` (P1 reserved-topic gap)
- participant disconnect while RPC pending → `recipientDisconnected` (P1)
- `Task.cancel()` on in-flight `performRpc` → state cleared
- 5× concurrent `performRpc` to same destination → distinct ids, no leaked pending
- v2 response stream from wrong sender identity (currently incorrectly resolves)
- `defer { #expect(rpcClient.pendingCount == 0) }` on happy paths to catch leaks

Tests/LiveKitCoreTests/RpcTests.swift
```diff
@@ -193,4 +195,407 @@ struct RpcTests {
...
+        let source = StreamReaderSource { continuation in
+            if let data = payload.data(using: .utf8) { continuation.yield(data) }
+            continuation.finish()
+        }
+        return TextStreamReader(info: info, source: source)
+    }
+}
+
+/// Simple async-safe boolean flag for cross-Task signaling in tests.
+private actor AsyncFlag {
```
pblazej: P2: This is racy — `nonisolated set()` enqueues a `Task { await self._set() }`, but `await flag.value` is a separate enqueue with no causal ordering. Under TSan / busy CI it'll flake. Swift Testing's `confirmation()` is the canonical replacement (already used elsewhere in this file):

```swift
try await confirmation("v2 stream opened", expectedCount: 1) { didStartStream in
    try await confirmation("no v1 packet", expectedCount: 0) { sawRpcRequest in
        let mock = MockDataChannelPair { packet in
            if case .rpcRequest = packet.value { sawRpcRequest() }
            if case let .streamHeader(h) = packet.value,
               h.topic == RpcStreamTopic.request { didStartStream() }
        }
        // …
    }
}
```

For `CapturedStreamHeader`, an `OSAllocatedUnfairLock<Header?>` is enough — no actor hops needed. Both helpers go away with this swap.

Tests/LiveKitCoreTests/RpcTests.swift
```diff
@@ -193,4 +195,407 @@ struct RpcTests {
...
+                payload: "Hi!",
+                timeoutMs: 8000
+            )
+            await room.rpcServer.handleIncomingRequestStream(
+                reader: reader,
+                callerIdentity: Participant.Identity(from: "v2-caller")
+            )
+
+            // Allow async send tasks a moment to complete
+            try await Task.sleep(nanoseconds: 200_000_000)
```
pblazej: P2: This 200 ms sleep is redundant — `handleIncomingRequestStream` only returns after `publishResponseStream`'s `writer.close()` completes, so by the time control returns to the test the mock has already observed both packets. With the `confirmation` swap suggested above this can just be deleted (no replacement needed).

Tests/LiveKitCoreTests/RpcTests.swift
```diff
@@ -193,4 +195,407 @@ struct RpcTests {
...
+    /// Caller times out if no ack arrives.
+    @Test func v2CallerResponseTimeout() async throws {
+        try await TestEnvironment.withRoom { room in
+            let destination = Participant.Identity(from: "v2-destination")
+            try await RpcTestSupport.installV2Remote(in: room, identity: destination)
+
+            // Do nothing in response — let the connection timeout (7s max round-trip) fire
+            room.publisherDataChannel = MockDataChannelPair { _ in }
+
+            await #expect(throws: RpcError.self) {
```
pblazej: P2: `#expect(throws: RpcError.self)` passes on _any_ `RpcError` — including the wrong code. Worth pinning to `connectionTimeout.code`:

```swift
let error = await #expect(throws: RpcError.self) {
    _ = try await room.localParticipant.performRpc(…)
}
#expect(error?.code == RpcError.BuiltInError.connectionTimeout.code)
```

Same for the large-payload fallback test below (`v2CallerV1FallbackRejectsLargePayload`).

Sources/LiveKit/Core/RpcClientManager.swift
```diff
@@ -0,0 +1,246 @@
...
+        )
+        let writer = try await room.localParticipant.streamText(options: options)
+        try await writer.write(payload)
+        try await writer.close()
+    }
+
+    // MARK: - Helpers
+
+    private func requireRoom() throws -> Room {
+        guard let room else { throw RpcError.builtIn(.applicationError) }
```
pblazej: P2: This used to throw `LiveKitError(.invalidState, message: "Room is nil")` from `Participant.requireRoom` — caller code rescuing `LiveKitError` will silently stop matching. The "no room" condition is really `invalidState` (precondition failure), not `applicationError` (handler-level RPC error), so I'd keep the original error type here. Same for `RpcServerManager.requireRoom()`.

Sources/LiveKit/Core/RpcClientManager.swift
```diff
@@ -0,0 +1,246 @@
...
+    /// otherwise v1 packet.
+    func performRpc(destinationIdentity: Participant.Identity,
+                    method: String,
+                    payload: String,
+                    responseTimeout: TimeInterval = 15) async throws -> String
+    {
+        let room = try requireRoom()
+
+        let remoteClientProtocol = room.remoteParticipants[destinationIdentity]?.clientProtocol ?? .v0
+        print("Rpc - Remote client protocol: \(remoteClientProtocol)")
```
pblazej: P2: Stray `print` — should be `log("…", .debug)` or just removed.

Sources/LiveKit/Core/RpcClientManager.swift
```diff
@@ -0,0 +1,246 @@
...
+        let useStreamTransport = remoteClientProtocol >= .v1
+
+        if !useStreamTransport, payload.byteLength > MAX_RPC_PAYLOAD_BYTES {
+            throw RpcError.builtIn(.requestPayloadTooLarge)
+        }
+
+        let requestId = UUID().uuidString
+        let maxRoundTripLatency: TimeInterval = 7
+        let minEffectiveTimeout: TimeInterval = 1
```
pblazej: P2: JS clamps the effective timeout to a minimum of 8000 ms (`rpc.ts`); we use 1 s. Functionally minor (it's just the value advertised to the receiver), but worth aligning since we're trying to be wire-compatible.

Sources/LiveKit/Core/RpcClientManager.swift
```diff
@@ -0,0 +1,246 @@
...
+            log("[Rpc] Response received for unexpected RPC request, id = \(requestId)", .error)
+            return
+        }
+        pending.onResolve(payload, error)
+    }
+
+    /// Resolve a pending RPC call from a v2 response stream on `lk.rpc_response`. Reads the
+    /// `lk.rpc_request_id` attribute to match against pending requests, then resolves
+    /// with the streamed payload.
+    func handleIncomingResponseStream(reader: TextStreamReader, senderIdentity _: Participant.Identity) async {
```
pblazej: P2: We never compare `senderIdentity` against the stored `PendingRpcResponse.participantIdentity` — defense-in-depth gap (a malicious participant could resolve someone else's pending RPC if they could spoof a stream attribute). Server-side filtering makes this hard in practice; still, a one-line guard would be nice.

Tests/LiveKitCoreTests/RpcTests.swift
```diff
@@ -102,7 +104,7 @@ struct RpcTests {
                     #expect(error is LiveKitError)
```
pblazej: P2: The `do { … } catch { #expect(error is LiveKitError) }` block above can simplify to:

```swift
await #expect(throws: LiveKitError.self) {
    try await room.registerRpcMethod("greet") { _ in "" }
}
```

Same shape as the new v2 tests below.

Tests/LiveKitCoreTests/RpcTests.swift
```diff
@@ -193,4 +195,407 @@ struct RpcTests {
...
+                }
+            }
+            room.publisherDataChannel = mockDataChannel
+
+            let response = try await room.localParticipant.performRpc(
+                destinationIdentity: destination,
+                method: "echo",
+                payload: largePayload
+            )
+            #expect(response.count == largePayload.count)
```
pblazej: P2: `response.count == largePayload.count` doesn't catch truncation/padding — `#expect(response == largePayload)` is stronger and Swift Testing's diff render handles big strings fine.

Tests/LiveKitCoreTests/RpcTests.swift
```diff
@@ -23,6 +23,8 @@ import LiveKitTestSupport
 
 @Suite(.serialized, .tags(.e2e))
```
pblazej: P3: These v2 tests don't actually need a server — they install fake remotes via `_state.mutate` and feed responses by directly calling `rpcClient.handleIncomingResponseStream`, which is a lot of plumbing to fake what `withRooms([...]) { rooms in ... }` gives us for free 🥲. Could we replace the mock-based v2 tests with real e2e ones (or have both — mocks for the unit-level edge cases like the P0 race / double-resume hooks, e2e for the happy paths)? Mirrors the existing v1 `performRpc` test below + the `RpcObjCTests.m` flow, and would actually exercise the data-stream path through real WebRTC instead of stubbing it. Then the suite stays `.serialized .tags(.e2e)` consistently.

Tests/LiveKitCoreTests/RpcTests.swift
```diff
@@ -193,4 +195,407 @@ struct RpcTests {
...
+            try await Task.sleep(nanoseconds: 200_000_000)
+
+            #expect(await didSeeStreamHeader.value == true)
+            #expect(await didSeeRpcResponsePacket.value == false)
+        }
+    }
+
+    /// Unhandled (non-RpcError) exception from a handler must be returned as a v1
+    /// `RpcResponse` packet with `APPLICATION_ERROR`, even between two v2 clients.
+    @Test func v2HandlerUnhandledErrorReturnsPacket() async throws {
```
pblazej: P3: Nit: this and `v2HandlerRpcErrorPassthroughViaPacket` could fold into one parameterized `@Test(arguments: [.genericError, .rpcErrorPassthrough])` — same body, just different `(thrown, expectedCode, expectedMessage)`.

Tests/LiveKitCoreTests/RpcTests.swift
```diff
@@ -193,4 +195,407 @@ struct RpcTests {
...
+            let mockDataChannel = MockDataChannelPair { packet in
+                if case .rpcRequest = packet.value {
+                    didSeeRpcRequestPacket.set()
+                    return
+                }
+                if case let .streamHeader(header) = packet.value, header.topic == RpcStreamTopic.request {
+                    captured.set(header)
+                    Task {
+                        // Simulate handler ack and v2 response stream
+                        let requestId = header.attributes[RpcStreamAttribute.requestId] ?? ""
```
pblazej: P3: Nit: `try #require(header.attributes[RpcStreamAttribute.requestId])` would fail fast on a regression where the SDK forgets to set the attribute — current `?? ""` makes the test time out 7 s later instead. Same pattern at line 262 and the optional-chained block at 242-247.

Tests/LiveKitCoreTests/RpcTests.swift
```diff
@@ -193,4 +195,407 @@ struct RpcTests {
...
+                )
+            }
+        }
+    }
+}
+
+// MARK: - Test support
+
+private struct RpcTestSupport {
+    /// Insert a remote participant into `room` whose `clientProtocol` advertises v2.
```
pblazej: P3: Naming inverts vs. the underlying enum (`installV2Remote` → `.v1`, `installV1Remote` → `.v0`) — a future reader will trip on this 🥲. Could be a single `installRemote(in:identity:clientProtocol:)` taking the enum directly.

Tests/LiveKitCoreTests/RpcTests.swift
```diff
@@ -193,4 +195,407 @@ struct RpcTests {
...
+        }
+    }
+
+    /// Caller times out if no ack arrives.
+    @Test func v2CallerResponseTimeout() async throws {
+        try await TestEnvironment.withRoom { room in
+            let destination = Participant.Identity(from: "v2-destination")
+            try await RpcTestSupport.installV2Remote(in: room, identity: destination)
+
+            // Do nothing in response — let the connection timeout (7s max round-trip) fire
```
pblazej: P3: Nit: comment says "let the connection timeout (7s max round-trip) fire", but with `responseTimeout: 0.05` the _outer_ `withThrowingTimeout` fires first — the 7 s ack-timeout path isn't actually exercised here. Worth either fixing the comment or splitting into two tests (one for the outer timeout mapping, one for the actual ack-timeout).

Sources/LiveKit/Core/RpcClientManager.swift
```diff
@@ -0,0 +1,246 @@
...
+    // MARK: - Outgoing wire
+
+    private func publishRequest(in room: Room,
+                                destinationIdentity: Participant.Identity,
+                                requestId: String,
+                                method: String,
+                                payload: String,
+                                responseTimeout: TimeInterval) async throws
+    {
+        guard payload.byteLength <= MAX_RPC_PAYLOAD_BYTES else {
```
pblazej, line 266: P3: Nit: this is redundant with the pre-check at line 50 in `performRpc`, and uses a different error type (`LiveKitError(.invalidParameter)` vs `RpcError.builtIn(.requestPayloadTooLarge)`) — could pick one site.

Sources/LiveKit/Core/RpcServerManager.swift
```diff
@@ -0,0 +1,283 @@
...
+                log("[Rpc] Response payload too large for \(method)", .warning)
+                return .failure(RpcError.builtIn(.responsePayloadTooLarge))
+            }
+            return .success(response)
+        } catch let error as RpcError {
+            return .failure(error)
+        } catch {
+            log("[Rpc] Uncaught error returned by RPC handler for \(method). Returning APPLICATION_ERROR instead.", .warning)
```
pblazej: P3: Nit: including `\(error)` in the log line would aid debugging — currently the underlying error message is dropped before we map to `applicationError`.

Sources/LiveKit/Types/ClientProtocol.swift
```diff
@@ -0,0 +1,48 @@
...
+    public static func < (lhs: Self, rhs: Self) -> Bool {
+        lhs.rawValue < rhs.rawValue
+    }
+}
+
+// MARK: - CustomStringConvertible
+
+extension ClientProtocol: CustomStringConvertible {
+    public var description: String {
+        String(rawValue)
```
pblazej: P3: Nit: `description` is just `String(rawValue)` — debug-friendly `"v0"`/`"v1"` would help logs. (The URL query in `Utils.swift` uses this; current numeric value is fine for the wire format, but logs would benefit from the named form.)

Sources/LiveKit/Types/ClientProtocol.swift
```diff
@@ -0,0 +1,48 @@
...
+
+import Foundation
+
+/// Client-to-client protocol version advertised to other participants.
+///
+/// This is distinct from ``ProtocolVersion``, which tracks the *signaling* protocol
+/// negotiated with the LiveKit server. ``ClientProtocol`` governs peer-to-peer feature
+/// negotiation between participants.
+@objc
+public enum ClientProtocol: Int, Sendable {
```
pblazej, line 24: P3 (ObjC parity): The `@objc` here is currently unreachable from ObjC because neither `Participant.clientProtocol` nor `ConnectOptions.clientProtocol` are marked `@objc` (consistent with `ProtocolVersion` precedent). Two options: drop `@objc` from this enum to match (smallest change), or expose the property + an ObjC-friendly `ConnectOptions` initializer overload (broader scope, would need new `ParticipantObjCTests.m` cases). Existing `RpcObjCTests.m` already covers register/perform/unregister via the deprecated `@objc` wrappers — those remain unchanged so existing ObjC coverage holds.

pblazej (changes requested):

pblazej:
Btw: the skills from `AGENTS.md` may help here

Sources/LiveKit/Protos/livekit_models.pb.swift
1egoman: Oops, the LLM added in the relevant fields manually and I didn't realize 😅 . Just updated the build protos in a1034df3.

Sources/LiveKit/Types/Options/ConnectOptions.swift
1egoman: Yea, I had the same thought. Originally `clientProtocol` wasn't in here but I moved it in here as a secondary step to be consistent with `protocol`.

Some more thinking out loud: IMO, they both should be in the same place ideally since conceptually they communicate the same type of state. Would removing `protocol` from here be a breaking change? If not, then these could both become constants (which was what `clientProtocol` was originally before I moved it in here).

One (maybe) advantage to exposing it could be that you could configure a client in an older `clientProtocol` version for end to end tests to test unusual scenarios? That's a pretty weak argument though and if it's going to actively confuse users then definitely not a good idea.

If there's no clear best path forward, I'm also open to making `protocol`/`clientProtocol` work inconsistently in the short term and making a follow up task for you or somebody else to clean it up.

Sources/LiveKit/Core/RpcServerManager.swift
1egoman: Fixed in 1ce45a71

Sources/LiveKit/Core/RpcClientManager.swift
1egoman: Ah, yes, good catch - fixed here: 24909cb9.

One small thing - I'm not sure if I implemented the `__test_hookAfterPublish` method properly. Did you mean literally adding a method with this name to let the test run code in the right phase of the lifecycle? Or did you mean something else? My depth of swift knowledge is minimal, so I apologize if this is a common pattern.

Sources/LiveKit/Core/RpcClientManager.swift
1egoman: Makes sense! As mentioned this also was fixed in 24909cb (and my same question around the `__test_` prefix applies to this point). I've also added a test in 68d17d99 and another `__test_` helper method which exercises this case specifically.

Sources/LiveKit/Core/RpcServerManager.swift
1egoman: Ah, good catch - I think I forgot this one permutation in my test case list. I will make sure to add it so other clients don't encounter this same issue.

Anyway, fixed in bf39ab15, and added a test very similar to what you suggested to exercise this case.

Sources/LiveKit/Core/Room+DataStream.swift
1egoman: Sounds good - fixed in 930e636f.

Sources/LiveKit/Core/Room.swift
1egoman: Ah, got it - makes sense! Fixed in 8948930a.

I'm curious in general though, given that from the reading I have done actor method calls are always async, I'd imagine this is probably a fairly common case to encounter when having systems of actors which all depend on each other. I just want to make sure I'm not overlooking some sort of much more elegant solution in my fix!

Sources/LiveKit/Core/RpcClientManager.swift
1egoman: Fixed as part of the above `AsyncCompleter` update: (link https://github.com/livekit/client-sdk-swift/pull/977#discussion_r3184065082, commit 24909cb).

Sources/LiveKit/Core/RpcClientManager.swift
1egoman: Ah, I definitely captured this one in my spec document, but didn't test it manually, so thanks for flagging. Fixed in ea9dde7b.

Tests/LiveKitCoreTests/RpcTests.swift
1egoman: Most of these test cases were added as part of previous updates, but the rest I have added in d425b62f.

LLM summary of the newly added tests in d425b62f:
> 1. reservedPrefixRejectedOnRegisterAndUnregister — asserts that both register{Text,Byte}StreamHandler and unregister{Text,Byte}StreamHandler throw LiveKitError on any lk.* topic, closing the previously-flagged gap where users could silently disable internal RPC v2 dispatch by unregistering lk.rpc_request.
> 2. performRpcCleansUpOnCancellation — starts a performRpc that hangs (no mock response), waits for pending state to register (pendingCount == 1), cancels the awaiting Task, asserts the call throws and that pending state is cleared (pendingCount == 0).
> 3. performRpcConcurrentRequestsHaveDistinctIds — runs 5 concurrent performRpc calls to the same destination, asserts each gets a distinct requestId
> (Set(observedIds).count == 5) and that no pending state leaks afterward.
> 4. v2ResponseStreamFromWrongSenderIsIgnored — installs a v2 remote, performs an RPC, then injects a v2 response stream from a different identity (v2-imposter). Asserts the spoofed response is ignored and the call ultimately times out with connectionTimeout rather than resolving with the spoofed payload.

Sources/LiveKit/Core/RpcClientManager.swift
pblazej: Yeah, the literal name doesn't matter — the property + `@Sendable` setter pair you went with is a perfectly fine shape (and arguably nicer than my method-only sketch). Just keep the `__test_` prefix convention so it's grep-able as test-only surface 👍

Sources/LiveKit/Core/Room.swift
pblazej: Yeah, there's no perfect pattern in the codebase right now — some places use a fire-and-forget `Task` from `init`, others wire at connect time like you did here. I may take this one and unify the other sites in a later commit 👍

pblazej:
Solid second pass — P0/P1 all closed with deterministic tests, the `AsyncCompleter` rewrite came out cleaner than I sketched (the publish-failure cleanup branch is a nice touch I didn't suggest) 🎉.

A handful of polish items still outstanding:

**P2:**
- `AsyncFlag` / `CapturedStreamHeader` still in `RpcTests.swift` — the `confirmation()` swap would let both helpers go away.
- `Task.sleep` still sprinkled across the new tests (~23 calls); most simulate fake network latency inside mock closures and could just be sequential `await`s through the actor.
- `requireRoom()` still throws `RpcError.builtIn(.applicationError)` instead of `LiveKitError(.invalidState)` — minor caller-visible behavior change.
- stray `print` at `RpcClientManager.swift:64`.
- `minEffectiveTimeout = 1` (s) still differs from JS's 8 s floor — wire-compat nit only.

**P3:** parameterize the v2 handler error tests, `try #require` instead of `?? ""`, `installV1Remote` / `installV2Remote` rename, redundant size pre-check at `publishRequest`, log `\(error)` in `dispatchToHandler`, `ClientProtocol.description` debug form, ObjC parity decision on the `@objc` annotation.

**Bigger thinking-out-loud item — unit vs e2e:** the v2 suite has grown a lot of mock-side plumbing (`MockDataChannelPair`, manual `_state.mutate`, direct `rpcClient.handleIncoming…` calls) that mostly fakes what `withRooms([...])` gives us for free against `livekit-server --dev`. Could we convert the happy-path v2 tests to `withRooms` like the v1 `performRpc` test + `RpcObjCTests.m`, and keep the mock variant only for cases that genuinely need internal `__test_*` hooks?

Tests/LiveKitCoreTests/RpcTests.swift
1egoman: Makes sense - I think this should be fixed in 684a410d.

Tests/LiveKitCoreTests/RpcTests.swift
1egoman: I think I've gotten all the `Task.sleep`s out of all of the tests - associated code here: 8b19c9c2

Sources/LiveKit/Core/RpcClientManager.swift
1egoman: Fixed in 8fb93ee8

Tests/LiveKitCoreTests/RpcTests.swift
1egoman: Fixed in 5b604a44

Sources/LiveKit/Core/RpcClientManager.swift
1egoman: Fixed in 1cdfc07e

Sources/LiveKit/Core/RpcClientManager.swift
1egoman: Fixed as part of d425b62f7

Tests/LiveKitCoreTests/RpcTests.swift
1egoman: Fixed in 8c752b72

Sources/LiveKit/Core/RpcClientManager.swift
1egoman: The web implementation of this looks like:
```ts
const maxRoundTripLatencyMs = 7000;
const minEffectiveTimeoutMs = maxRoundTripLatencyMs + 1000;
// ...
const effectiveTimeoutMs = Math.max(responseTimeoutMs, minEffectiveTimeoutMs);
```

Updated the swift implementation to have the same logic here: 53285d31

Tests/LiveKitCoreTests/RpcTests.swift
1egoman: Fixed in 24753562

Tests/LiveKitCoreTests/RpcTests.swift
1egoman: Fixed in 1ecec2b7

Sources/LiveKit/Core/RpcServerManager.swift
1egoman: Fixed in 5301dd46

Sources/LiveKit/Types/ClientProtocol.swift
1egoman: Fixed in e6a554a4.

Tests/LiveKitCoreTests/RpcTests.swift
1egoman: Fixed in d84eb0bc

Sources/LiveKit/Types/ClientProtocol.swift
1egoman: Fixed in 00c4eaf1

Tests/LiveKitCoreTests/RpcTests.swift
1egoman: Fixed in bcf1d382

Tests/LiveKitCoreTests/RpcTests.swift
1egoman: I removed the misleading comment here: 6eb7c9bd

Note that there already is another test exercising the 15s default (see 53285d3) and unless mocking global sleeps is fairly easily doable, I'm not sure that an additional test makes sense.

Sources/LiveKit/Core/RpcClientManager.swift
1egoman: Good catch - moved to the site in `publishRequest` only here: d1e25b47

Sources/LiveKit/Core/Room+DataStream.swift
1egoman: @pblazej I mentioned this to @lukasIO and he wasn't a fan.

LLM summary of lukas's perspective from our chat:
```
Breaking change cost is high — that alone is a strong reason not to do it.
Dependencies complicate a hard block — other SDKs that build on the core SDK (e.g., the agents SDK) legitimately register lk.* listeners, so a blanket prohibition doesn't cleanly work.
A suppression mechanism is the wrong fix — letting some callers opt out of the warning/block "because they know what they're doing" doesn't sit right with him.
Prefer a structural solution — if the SDK registers its own lk.* listeners at room construction time, any later user attempt to register the same name would naturally fail through the existing duplicate-registration path, without needing a special rule.
Low priority absent real reports — he asks whether anyone has actually hit this, implying the change isn't worth the cost without evidence of the problem.
```

I don't think I really have a strong perspective here, I think @pblazej if this is important to you there are some way to do this which could still permit some of the agents sdk usage patterns like warnings rather than errors. Feel free to weigh in more here if desired.

lukasIO: the LLM's summary sounds a lot harsher than my standing on this feels 👀 

And yeah, I think it's mostly important to consider the agents-sdk registering lk topics. 
Potentially this could also be other plugins of course that would do this.

pblazej: my take on that is: if we gonna break that we should break that forever for all `lk.*` topics rather than re-breaking for each small feature 😄 

It's early Mon morning, but I think LLM is exaggerating that as well 👍

1egoman: @pblazej The thing that I am still a bit conflicted on is that while I think it would be good to reserve the `lk.*` prefix for internal client sdk use, the agents sdk uses this prefix already - and since the agents sdk is indistinguishable to a client sdk from an end user's application, there's not really a good way to leave the interface open that other clients couldn't also take advantage of.

So I think we'd either need some way for a client on registration to be able to indicate it's actually a "first party livekit client" or something (brought this up to lukas last week, he also was not a fan of this), or the filtering has to either be by topic or just outright removed.

Other possible idea for data streams at least - add logic so that internal sdk uses can register a data stream handle via a different way, and which this other mechanism is uses, the data is "tee"d and sent to both the external handler and the other internal sdk feature, so an external consumer can add / remove a handler and no matter what happens they can't break internal sdk features.

Tests/LiveKitCoreTests/RpcTests.swift
1egoman: I have started on this in 932b071, but could maybe use some advice on if I am moving in a good direction here or not. Note I haven't heavily reviewed this to ensure the tests are all still testing the same things - I suspect they are a little less robust right now.

pblazej:
@1egoman I don't see anything "unresolved" atm, let me take a final read and resolve the conflicts 🌮 

BTW any manual (x-platform) tests worth performing now?

1egoman:
> BTW any manual (x-platform) tests worth performing now?

@pblazej A few platform combinations that come to mind: 1) old client-sdk-js (ie, current npm release) <-> new swift, 2) new client-sdk-js (ie, https://github.com/livekit/client-sdk-js/pull/1832) <-> new swift. Going through all the happy path cases in the [spec document](https://github.com/livekit/client-sdk-js/blob/rpc-with-compression-data-streams/RPC_SPEC.md) I think are worthwhile at a minimum.

Sources/LiveKit/Core/Room+DataStream.swift
pblazej: @1egoman @lukasIO I agree there are "semi-private" consumers of this, maybe we can just revert to `lk.rpc*` and delay this decision?

pblazej (approved):
I made some changes, notable mentions:
- test structure - to match the JS spec even closer
- RPC internal initialization logic vs `Room` - no silver bullet here as the `Room` part is pretty messy, now the (internal) topic registration is just idempotent, without changes in the public stream semantics (allowing multiple listeners, etc.); no need to store the state/use singletons/etc.
- related to the above: removing breaking change - `unregister` is still non-throwing; no-op if topic is reserved
- data channel improvements to handle transient failures leading to `dataChannelNotOpen` error; more refactors may come in separate PRs

🎉

PR #985, opened 2026-05-05
Audio: support app-audio publishing without claiming a device

sgu-bithuman:
## Summary

`AudioManager.shared.mixer.capture(appAudio:)` is the documented entry-point for feeding app-supplied PCM into the publish path (used by `MacOSScreenCapturer`, `BroadcastScreenCapturer`, and any custom audio source). In **manual rendering mode** it drops every buffer because the input chain (`appNode → appMixerNode → mainMixerNode`) is never wired up — the WebRTC ADM only invokes `engineWillConnectInput` when connecting a real device, which manual rendering by definition has none of.

This blocks several scenarios that `AudioManager.setManualRenderingMode(_:)` otherwise enables:

- **Server-side avatars / agents** that only republish app-supplied audio (no mic). Multi-process deployments on a single Mac can't share the audio device, so the existing `startLocalRecording()` path tops out at one audio publisher per machine. See `testManualRenderingModePublishAudio` in `LiveKitAudioTests` — that test demonstrates the publish flow but only one process can hold the device.
- **Screen-share with system audio but no microphone.**
- **Custom audio sources** (e.g. file playback) that don't have a real input device.

## Changes

- Add `MixerEngineObserver.wireAppAudioPath()` — public, idempotent. Wires `appNode → appMixerNode → mainMixerNode` against `engine.mainMixerNode` using the same Float32 player-node format derivation as the `src=nil` branch in `engineWillConnectInput`.
- `capture(appAudio:)` lazy-wires the path on the first call when it sees a manual-rendering engine. Existing call sites get the fix automatically; no API churn for callers.
- `engineWillRelease` clears `isInputConnected` and `mainMixerNode` so a recreated engine re-runs the wiring.

Purely additive — no public API removed, no signature changes. Engine-running and `isInputConnected` guards in `capture(appAudio:)` are preserved.

## Test plan

- [x] Builds cleanly: `swift build --target LiveKit`
- [x] Verified end-to-end with **eight server-side processes** publishing audio concurrently on one Mac. PCM byte-stream → `mixer.capture(appAudio:)` → published track. Each process: ~0.019 avg / ~0.25 peak RMS, ~36% non-zero samples vs ~2% noise floor on the previous code path.
- [x] Existing `LocalAudioTrack` mic flow unchanged (the new code only fires when `engine.isInManualRenderingMode`).
- [ ] CI / `LiveKitAudioTests` pass (would benefit from a new test mirroring `testManualRenderingModePublishAudio` but with `mixer.capture(appAudio:)` as the source, replacing the file player). Happy to add this if useful.

### DIFF
--- a/Sources/LiveKit/Audio/MixerEngineObserver.swift
+++ b/Sources/LiveKit/Audio/MixerEngineObserver.swift
@@ -169,6 +169,14 @@ public final class MixerEngineObserver: AudioEngineObserver, Loggable {
         engine.detach(micMixerNode)
         engine.detach(soundPlayerNodes)
 
+        // Reset wire-up state so a subsequent engineDidCreate +
+        // engineWillConnectInput (or wireAppAudioPath() in manual mode)
+        // re-establishes the graph on the new engine instance.
+        _state.mutate {
+            $0.isInputConnected = false
+            $0.mainMixerNode = nil
+        }
+
         return nextResult ?? 0
     }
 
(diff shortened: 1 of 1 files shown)


pblazej:
@sgu-bithuman could you add some description of the scenarios that are not covered by the current implementation?

E.g. https://github.com/webrtc-sdk/webrtc/blob/597a7f4fb2c252c8d3c1d8d8144bfb730fbacaf2/modules/audio_device/audio_engine_device.mm#L1626-L1650

https://github.com/livekit/client-sdk-swift/blob/397b800ac18584b1f24e661a003708bcd0d2909e/Tests/LiveKitAudioTests/AudioEngineTests.swift#L251-L301

Based on an initial review, it should work "as is".

PR #986, opened 2026-05-05
Single PC improvements

pblazej:
The publisher transport's `negotiate()` always runs through a 20ms debouncer to coalesce rapid renegotiations during track publishing. At connect time there is nothing to coalesce, and the delay sits directly in the critical path — most visible in single-PC mode where the client must initiate the SDP offer.

Two changes:
1. `Transport.negotiate(force:)` — when `force=true`, cancel any pending debounced action and call `createAndSendOffer()` directly.
2. Move the eager-negotiate call out of `configureTransports` and into `fullConnectSequence` *after* `signalClient.resumeQueues()`, calling with `force=true`. This is required because offers are not queueable (`Livekit_SignalRequest.canBeQueued()` returns false for `.offer`), so a forced send while the request queue is still suspended would silently drop the offer.


Sources/LiveKit/Core/Room+TransportDelegate.swift
```diff
@@ -113,10 +113,6 @@ extension Room: TransportDelegate {
         case LKRTCDataChannel.Labels.lossy: subscriberDataChannel.set(lossy: dataChannel)
         default: log("Unknown data channel label \(dataChannel.label)", .warning)
         }
-
-        if subscriberDataChannel.isOpen {
-            connectSpan?.record("dc_open")
```
pblazej, line 118: This is moved to 769386a9b9edc4a6ecd2499690b8b19b9d1fda92 to make it consistent with the spec, I'm not sure if we should drop this data point if that's not really awaitable from Swift.

I may re-evaluate the spec (and other SDKs).

pblazej:
The gap seems to be closed (within the noise level), note that the interpretation changes due to `DC_OPEN` tweaks above ⬆️ so that the total wall time also increased slightly.

https://github.com/livekit/client-sdk-swift/actions/runs/25377916039

Sources/LiveKit/Core/Room+TransportDelegate.swift
pblazej: Right now I'm more for dropping it...

lukasIO (approved):
makes sense to me!

Sources/LiveKit/Core/Room+TransportDelegate.swift
pblazej: Dropped in b37099b26e5518e85166ba424e15de39c79e5d1e

Sources/LiveKit/Core/Room+Engine.swift
```diff
@@ -286,6 +283,16 @@ extension Room {
 
         // Resume after configuring transports...
         await signalClient.resumeQueues()
+        try Task.checkCancellation()
+
+        // Eager publisher negotiation must run after `resumeQueues()` —
```
pblazej, line 288: Subtle but important: moving the offer send out of `Debounce.schedule { try? await action() }` also surfaces errors that used to be silently swallowed (createOffer / setLocalDescription / WS send failures). Pairs with #987 (typed errors through `AsyncCompleter`) in covering connect-time failures.

PR #987, opened 2026-05-06
Propagate more n/w errors

pblazej:
Small but important updates to `AsyncCompleter(Map)`, allowing custom error types.

Should not collide with (upcoming) typed throws.


hiroshihorie:
One small concern: this now turns every `.disconnected` / `.failed` peer-connection callback into `LiveKitError(.network)` before checking the room state.

Could this also happen during a normal `room.disconnect()`? In that path we move to `.disconnecting`, then `cleanUpRTC()` closes the transports, and WebRTC may emit a disconnect/failed state as part of that teardown. If a caller is concurrently waiting on the transport completers, they could now see `.network` even though the disconnect was user-initiated.

Maybe we should gate this to active connection/reconnect states, or skip the typed network reset when the room is already `.disconnecting` / `.disconnected`, so the new distinction between network failures and user cancellation stays reliable.

pblazej:
@hiroshihorie yes, I'm afraid of wrapping here; will check your case maybe adding more tests around it.

BTW I started working on typed throws #988 to see how much damage it does.

pblazej:
@hiroshihorie fixed in 56d4a8b49df0b760916e848bbc68377d83b01cda

hiroshihorie (approved):

pblazej:
cc @dfed should help with your case

PR #994, opened 2026-05-07
ci: add/fix custom CodeQL workflow

pblazej:
Tries to remove `autobuild` from the equation here.

### DIFF
--- a/.github/workflows/codeql.yaml
+++ b/.github/workflows/codeql.yaml
@@ -20,7 +20,7 @@ jobs:
   analyze:
     name: Analyze (${{ matrix.language }})
     runs-on: ${{ matrix.os }}
-    timeout-minutes: 60
+    timeout-minutes: 120
     permissions:
       security-events: write
       packages: read
--- a/.github/workflows/codeql.yaml
+++ b/.github/workflows/codeql.yaml
@@ -18,7 +18,7 @@ jobs:
   analyze:
     name: Analyze (${{ matrix.language }})
     runs-on: ${{ matrix.os }}
-    timeout-minutes: 120
+    timeout-minutes: 60
     permissions:
       security-events: write
       packages: read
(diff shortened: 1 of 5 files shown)


pblazej:
I already see the problem - compilation with instrumentation takes ages 🥱 

Not sure if it'll even make sense for every PR (just cron).

.github/workflows/codeql.yaml
```diff
@@ -0,0 +1,74 @@
...
+      packages: read
+      actions: read
+      contents: read
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          # Standard macos-26 is too slow under CodeQL's Swift extractor.
+          - language: swift
+            os: macos-26-xlarge
```
pblazej, line 33: After 1000 runs, this is a _minimal_ setup

.github/workflows/codeql.yaml
```diff
@@ -0,0 +1,74 @@
...
+
+      - name: Install swiftly
+        if: matrix.language == 'swift'
+        run: brew install swiftly && swiftly init --quiet-shell-followup --skip-install -y
+
+      # Resolve outside the tracer so binary xcframework downloads
+      # don't run under DYLD_INSERT_LIBRARIES.
+      - name: Resolve dependencies
+        if: matrix.language == 'swift'
+        run: swiftly run +xcode swift package resolve
```
pblazej, line 59: This plus single `--arch` should do the trick, however with today's GitHub stability not sure how long it's gonna last...

pblazej:
Looks like it kinda works:

```
|             Metric             |  Value  |
+--------------------------------+---------+
| Number of unresolved AST nodes |       0 |
| Number of extracted AST nodes  | 1226600 |
```

hiroshihorie (approved):

PR #995, opened 2026-05-08
Fix remote audio volume range

hiroshihorie:


xianshijing-lk (approved):

Sources/LiveKit/Track/Remote/RemoteAudioTrack.swift
```diff
@@ -21,15 +21,22 @@ internal import LiveKitWebRTC
...
     public var volume: Double {
         get {
             guard let audioTrack = mediaTrack as? LKRTCAudioTrack else { return 0 }
-            return audioTrack.source.volume / 10
+            return min(max(audioTrack.source.volume, 0), 10)
         }
         set {
             guard let audioTrack = mediaTrack as? LKRTCAudioTrack else { return }
-            audioTrack.source.volume = newValue * 10
+            audioTrack.source.volume = min(max(newValue, 0), 10)
```
xianshijing-lk: Curiously, when setting the volume, is this consistent with android volume change ? 

Do we need to update the doc if it takes 0 - 10 as value ?

hiroshihorie: Yes it should be aligned with Android now.
Need to update doc if this is mentioned somewhere.

pblazej (approved):
Nice

hiroshihorie:
Added workaround until https://github.com/webrtc-sdk/webrtc/pull/236 ships.

PR #996, opened 2026-05-11
Fix #NoUseUnstructuredThrowingTask under newest Swift

pblazej:
Add `Task.discardingErrors` helper that wraps unstructured throwing tasks, logs errors via the shared logger. Preserves caller actor isolation via `@_inheritActorContext + @isolated(any) + sending`.

Generally, it emphasizes the fact of "swallowing" errors vs using `try?, _ =` etc.
Can be gradually migrated to "handle errors" if needed...

No changes to the `detached` semantics.


Sources/LiveKit/Support/Async/AsyncDebounce.swift
```diff
@@ -32,7 +32,7 @@ actor Debounce {
         _task?.cancel()
     }
 
-    func schedule(_ action: @Sendable @escaping () async throws -> Void) {
+    func schedule(_ action: sending @escaping () async throws -> Void) {
```
pblazej: This basically allows "context-aware" sendability, not imposed on the whole type in Swift 6.

https://www.donnywals.com/sending-vs-sendable-in-swift/

Cosmetic change, mimics `Task.init`.

Sources/LiveKit/Support/Async/Task+Discarding.swift
```diff
@@ -0,0 +1,74 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import Foundation
+
+extension Task: Loggable where Success == Void, Failure == Never {}
```
hiroshihorie: ✅ LGTM. Tiny nit: can we avoid `extension Task: Loggable` here? Protocol conformances are global, so this leaks a `Task<Void, Never>` conformance from the SDK. A private logger shim should work just as well.

hiroshihorie (approved):

Sources/LiveKit/Support/Async/Task+Discarding.swift
pblazej: Yeah I think I reverted that on the way 0e09b7a5fa76bacc8f0be67a4275ae11a232653a

PR #999, opened 2026-05-11
Expose key derivation algorithm on KeyProviderOptions

### LINKED ISSUE #997: Cannot subclass BaseKeyProvider to use RTCKeyDerivationAlgorithmHKDF
BaseKeyProvider is public **final** class → cannot be subclassed.

rtcKeyProvider is let with no access modifier → module-internal, unreadable and unassignable from app code, and its type LKRTCFrameCryptorKeyProvider lives behind internal import LiveKitWebRTC → also unreachable.

Android's Kotlin SDK ships KeyProvider as an interface with val rtcKeyProvider as a contract member. iOS's wrapper is a final concrete class that owns its rtcKeyProvider privately, so the Android pattern has no direct iOS analog.

Swift version affected: 2.14.0
Android version: https://github.com/livekit/client-sdk-android/releases/tag/v2.24.0

pblazej:
## Summary
- Adds `keyDerivationAlgorithm` (`.pbkdf2` / `.hkdf`, default `.pbkdf2`) and `discardFrameWhenCryptorNotReady` (default `false`) to `KeyProviderOptions`, matching the Android SDK shape.
- Resolves #997 — HKDF was previously unreachable because `BaseKeyProvider` is `final` and `rtcKeyProvider` (typed `LKRTCFrameCryptorKeyProvider`) lives behind `internal import LiveKitWebRTC`, so subclassing couldn't bridge the WebRTC ctor parameter.
- Fully source-compatible: new fields default to the values WebRTC was already using.

## Test plan
- [x] Parametrized the 6 existing tests in `EncryptedDataChannelTests` over `[.pbkdf2, .hkdf]` — 12 cases pass against a local `livekit-server --dev` (~50s).
- [x] `swiftlint` clean, `xcodebuild build` clean on macOS.

### DIFF
--- a/.changes/e2ee-key-derivation-algo
+++ b/.changes/e2ee-key-derivation-algo
@@ -1,0 +1,1 @@
+minor type="added" "Expose `keyDerivationAlgorithm` (PBKDF2/HKDF) and `discardFrameWhenCryptorNotReady` on `KeyProviderOptions` for parity with Android"
--- a/Sources/LiveKit/E2EE/KeyProvider.swift
+++ b/Sources/LiveKit/E2EE/KeyProvider.swift
@@ -25,6 +25,17 @@ public let defaultRatchetWindowSize: Int32 = 0
 public let defaultFailureTolerance: Int32 = -1
 public let defaultKeyRingSize: Int32 = 16
 
+/// Algorithm used to derive encryption keys from the input key material.
+///
+/// - SeeAlso: ``KeyProviderOptions``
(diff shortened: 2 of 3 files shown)


hiroshihorie (approved):
nit: might be tiny breaking change for ObjC but I think it's fine.

pblazej:
@hiroshihorie good catch, that's still `@objcMembers`, but I agree that's acceptable (given the functionality gap).

PR #1003, opened 2026-05-12
Expose keyDerivationAlgorithm on KeyProviderOptions (#997)

### LINKED ISSUE #997: Cannot subclass BaseKeyProvider to use RTCKeyDerivationAlgorithmHKDF

john-rocky:
Fixes #997.

\`BaseKeyProvider\` is \`final\` and \`rtcKeyProvider\` is module-internal, so apps that need \`RTCKeyDerivationAlgorithmHKDF\` had no way to reach the underlying \`LKRTCFrameCryptorKeyProvider\` to swap the KDF. The Android SDK already exposes this via \`KeyProvider.Options.keyDerivationAlgorithm\` ([\`client-sdk-android@2.24.0\`](https://github.com/livekit/client-sdk-android/releases/tag/v2.24.0)), so this brings the iOS surface in line with the suggestion from the issue thread ("we can just parametrize it ⏫ without changing access levels").

### Changes
- Add public \`KeyDerivationAlgorithm\` enum (\`.pbkdf2\` / \`.hkdf\`) so callers don't need to touch \`LiveKitWebRTC\` types — the underlying \`internal import\` stays internal.
- Add \`keyDerivationAlgorithm: KeyDerivationAlgorithm = .pbkdf2\` to \`KeyProviderOptions\` (default keeps the previous behavior) plus matching \`isEqual\` / \`hash\` updates.
- Route both \`BaseKeyProvider\` inits through a single private \`makeRTCProvider(options:)\` helper that uses the full \`LKRTCFrameCryptorKeyProvider\` init exposing \`keyDerivationAlgorithm\`. \`discardFrameWhenCryptorNotReady\` is passed as \`false\` to preserve the previous behavior (the older inits never surfaced it).

### Usage

\`\`\`swift
let provider = BaseKeyProvider(options: KeyProviderOptions(
    sharedKey: true,
    keyDerivationAlgorithm: .hkdf
))
\`\`\`

### Test
\`\`\`
$ xcodebuild build -scheme LiveKit -destination 'platform=macOS'
** BUILD SUCCEEDED **
\`\`\`

\`xcodebuild test -only-testing LiveKitCoreTests\` requires a local \`livekit-server --dev\` per AGENTS.md, so I didn't run the integration suite. The change is API-shape only and the default keeps the existing PBKDF2 wiring unchanged.

### DIFF
--- a/Sources/LiveKit/E2EE/KeyProvider.swift
+++ b/Sources/LiveKit/E2EE/KeyProvider.swift
@@ -25,6 +25,18 @@ public let defaultRatchetWindowSize: Int32 = 0
 public let defaultFailureTolerance: Int32 = -1
 public let defaultKeyRingSize: Int32 = 16
 
+/// Key-derivation algorithm used by the underlying frame cryptor.
+///
+/// Mirrors the LiveKit Android SDK's `KeyProvider.Options.keyDerivationAlgorithm`
+/// so that cross-platform encryption setups can agree on the same KDF without
+/// reaching into `LiveKitWebRTC` directly. Defaults to PBKDF2, matching the
+/// previous behavior of this SDK.
+@objc
+public enum KeyDerivationAlgorithm: Int, Sendable {
+    case pbkdf2 = 0
+    case hkdf = 1
+}
+
 @objcMembers
 public final class KeyProviderOptions: NSObject, Sendable {
     public let sharedKey: Bool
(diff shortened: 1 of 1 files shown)


pblazej:
Is there any improvement over the (duplicated) https://github.com/livekit/client-sdk-swift/pull/999?

PR #1012, opened 2026-05-14
Fix Room.sid resolution from room updates

### LINKED ISSUE #1009: Room.sid() resolves with empty SID when room SID arrives in a later room update
**Describe the bug**

`Room.sid()` can resolve with an empty room SID if the initial `JoinResponse.room.sid` is empty. After that, later `RoomUpdate` messages do not populate `Room.sid`, so clients waiting for the server-assigned room SID never receive the real `RM_...` value.

In our app, `room.name` is populated correctly, but `room.sid?.stringValue` is empty. Calling `try await room.sid()` also returns an empty SID. This causes analytics/session correlation to miss the real LiveKit room SID.

Looking at the Swift SDK implementation, `didReceiveConnectResponse` assigns `Room.Sid(from: joinResponse.room.sid)` unconditionally, while `didUpdateRoom` updates metadata/counts but not `sid`.

Android appears to handle this case by updating `sid` from room updates:

```kotlin
override fun onRoomUpdate(update: LivekitModels.Room) {
    if (update.sid != null) {
        sid = Sid(update.sid)
    }
}
```

**SDK Version**

2.14.1

**iOS/macOS Version**

iOS Simulator / iOS 26.4

**Xcode Version**

Xcode 26.4.0  
Swift version: from Xcode 26.4 toolchain

**Steps to Reproduce**

1. Connect to a LiveKit room where the initial join response exposes a room name but has an empty `room.sid`.
2. After `room.connect(...)` completes, inspect:

   ```swift
   room.name
   room.sid?.stringValue
   ```

3. Call:

   ```swift
   let sid = try await room.sid()
   ```

4. Observe that `sid.stringValue` is empty.
5. Wait for room updates. `room.sid` remains empty, even though the server-assigned room SID exists and Android receives it via room SID flow / room update handling.

**Examples**

The problem is reproducible with:

- [ ] [`client-example-swift`](https://github.com/livekit/client-example-swift)
- [ ] [`client-example-collection-swift`](https://github.com/livekit/client-example-collection-swift)

I have not reproduced this in the public example apps yet. The behavior was observed in an app using the Swift SDK directly.

**Expected behavior**

`Room.sid` should only be populated from a non-empty server-assigned SID. If the initial join response has an empty SID, `try await room.sid()` should keep waiting until a non-empty SID is available.

If a later `RoomUpdate.room.sid` contains the SID, the Swift SDK should update `Room.sid` and resolve `room.sid()` with that value.

AndrazP:
## Summary
Fixes `Room.sid` resolution when the initial join response contains an empty room SID.

Previously, the SDK assigned `Room.Sid(from: joinResponse.room.sid)` unconditionally. If the join response SID was empty, `Room.sid` became a non-nil empty SID and `try await
room.sid()` could resolve with that empty value. Later `RoomUpdate` messages did not update `Room.sid`.

This PR:
- ignores empty `JoinResponse.room.sid` values
- updates `Room.sid` from non-empty `RoomUpdate.room.sid`
- adds coverage for an empty join SID followed by a non-empty room update SID

## Why
This matches Android SDK behavior, where room updates can populate the room SID after join.

## Test plan
```sh
xcodebuild test \
  -scheme LiveKit \
  -destination 'platform=iOS Simulator,name=iPhone 17' \
  -only-testing:LiveKitCoreTests/RoomTests \
  -parallel-testing-enabled NO
```

Fixes #1009

### DIFF
--- a/Sources/LiveKit/Core/Room+SignalClientDelegate.swift
+++ b/Sources/LiveKit/Core/Room+SignalClientDelegate.swift
@@ -112,7 +112,9 @@ extension Room: SignalClientDelegate {
             }
 
             _state.mutate {
-                $0.sid = Room.Sid(from: joinResponse.room.sid)
+                if !joinResponse.room.sid.isEmpty {
+                    $0.sid = Room.Sid(from: joinResponse.room.sid)
+                }
                 $0.name = joinResponse.room.name
                 $0.serverInfo = joinResponse.serverInfo
                 $0.maxParticipants = Int(joinResponse.room.maxParticipants)
@@ -143,6 +145,9 @@ extension Room: SignalClientDelegate {
 
     func signalClient(_: SignalClient, didUpdateRoom room: Livekit_Room) async {
         _state.mutate {
+            if !room.sid.isEmpty {
+                $0.sid = Room.Sid(from: room.sid)
+            }
             $0.metadata = room.metadata
             $0.isRecording = room.activeRecording
             $0.numParticipants = Int(room.numParticipants)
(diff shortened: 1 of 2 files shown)


pblazej:
Merged #1010

Issue #1011, opened 2026-05-14
LiveKit participant audio becomes inaudible during incoming calls or alarms


YerkeAmankul:
### Description

When using LiveKit audio/video calls on iOS, participant audio becomes muted or inaudible if the device receives:

* an incoming phone call
* an alarm
* other system audio interruptions

After the interruption starts or ends, remote participant audio does not properly recover.

### Steps to Reproduce

1. Join a LiveKit room with at least 2 participants
2. Start active voice communication
3. Trigger one of the following on the device:

   * incoming cellular call
   * alarm
   * timer
   * other system interruption
4. Return to the LiveKit call

### Expected Result

Participant audio should automatically resume after the interruption ends.

### Actual Result

Remote participant audio is not audible after interruption, or volume becomes extremely low until reconnecting/restarting audio session.

### Environment

* LiveKit SDK version: 2.9.0

hiroshihorie:
Is this on v2.9.0 ? Can you try with latest SDK ?

YerkeAmankul:
@hiroshihorie in v2.14.1 also occurred bug

Issue #886, opened 2026-01-07
Audio Recording Fails to Resume System Audio After Incoming Call Interruption in Background


willonboy:
**Describe the bug**
Problem Description: When I use broadcast to record the screen and system audio (while also capturing mic audio), if a call comes in, regardless of whether I answer the phone or not, the recording of system audio will be interrupted. Afterward, without restarting the app, it will always be unable to record system audio. I investigated the cause, and the issue lies in the method `public func capture(appAudio inputBuffer: AVAudioPCMBuffer)` in MixerEngineObserver.swift. In this method, the condition `engine.isRunning` is not met, resulting in the system audio not being mixed and sent. I tried to fix it:

1. I added `pause` and `start` methods to `MixerEngineObserver` for pausing and starting the audio engine. The specific code is as follows:

```swift
extension MixerEngineObserver {
    public func pause() {
        let appNode = _state.read {
            $0.appNode
        }
        guard let engine = appNode.engine else {
            log("Engine is not running engine == nil", .warning)
            return
        }
        try? engine.pause()
        appNode.pause()
    }
    public func start() {
        let appNode = _state.read {
            $0.appNode
        }
        guard let engine = appNode.engine else {
            log("Engine is not running engine == nil", .warning)
            return
        }
        do {
            try engine.start()
        } catch {
            log("Engine start failed \(error)", .error)
        }
        appNode.play()
    }
}
```

2. When constructing `AudioManager`, I added a listener for `AVAudioSession.interruptionNotification`. When receiving the event, I use the methods added to `MixerEngineObserver`. The specific code is as follows:

```swift
init() {
#if os(iOS) || os(visionOS) || os(tvOS)
        let engineObservers: [any AudioEngineObserver] = [audioSession, mixer]
#else
        let engineObservers: [any AudioEngineObserver] = [mixer]
#endif
    _state = StateSync(State(engineObservers: engineObservers))
    _admDelegateAdapter.audioManager = self
    RTC.audioDeviceModule.observer = _admDelegateAdapter
        NotificationCenter.default.addObserver(
            self,
            selector: #selector(handleAudioSessionInterruption),
            name: AVAudioSession.interruptionNotification,
            object: AVAudioSession.sharedInstance()
        )
    }
    @objc
    private func handleAudioSessionInterruption(_ notification: Notification) {
        guard
            let info = notification.userInfo,
            let typeValue = info[AVAudioSessionInterruptionTypeKey] as? UInt,
[...]

PR #536, opened 2025-01-06
AVAudioEngine version AudioDeviceModule

hiroshihorie:
Using lib from https://github.com/webrtc-sdk/webrtc/pull/158.
macOS will still use previous implementation at the moment. (Device selection etc not implemented yet)

### New features
[Advanced other audio ducking](https://developer.apple.com/videos/play/wwdc2023/10235/?time=199)
Lower audio from other apps dynamically when voice activity is detected.

[Muted talker detection](https://developer.apple.com/videos/play/wwdc2023/10235/?time=475)
Detect voice activity even mic is off after calling `AudioManager.shared.prepareRecording()`.

Provide custom AVAudioNode configuration for input and output through `onEngineWillConnectInput` and `onEngineWillConnectOutput` handlers.

Early recording initialization with `AudioManager.shared.prepareRecording()` without mic indicator turning on.

Early mic input buffer capturing with `AudioManager.shared.startLocalRecording()` without the need for Room and connection.

### Improvements

Correct mic muting logic without the requirement of restarting whole audio unit (previous implementation).

Correct audio session configuration timing. AudioDeviceModule requests configuration when it's required, instead of counting tracks and configuring it in advance (previous implementation).

PR #228, opened 2023-07-13
Adjust AVAudioSession CategoryOptions (include .mixWithOthers)

hiroshihorie:
Connection drops without this option when playing other audio.
https://github.com/livekit/client-sdk-swift/issues/227#issuecomment-1632802240

PR #1019, opened 2026-05-27
Audio Session Defaults

pblazej:
Reworks the iOS audio session defaults. No public API changes; observable behavior changes are listed below with rationale.

- **`.playAndRecord` is sticky for the life of the engine session.** Category selection is now a pure `selectConfiguration(state:)` driven by the engine state plus a `hasRecorded` bit: `.playAndRecord` once recording engages, kept until the WebRTC engine stops, `.playback` otherwise. Muting the microphone no longer flips the category back to `.playback` — a mid-session category change tears down Apple's Voice Processing I/O and leaves echo cancellation dead for the rest of the call. The bit is keyed to the engine's own session requirement, so external requirement holders (e.g. a prepared `SoundPlayer` sound, which lives for the lifetime of the prepared sound) don't pin `.playAndRecord` after the call ends. Covered by new hermetic unit tests (`AudioSessionCategorySelectionTests`).
- **`.mixWithOthers` dropped from the `.playAndRecord` presets** (kept on the listener `.playback` preset). On the in-call path it lets another app hold the audio device alongside the call — a known echo source (see e.g. [Agora's echo troubleshooting](https://docs.agora.io/en/help/quality-issues/echo)) — and the WebRTC ADM needs a retry loop to work around a `-66637` engine-init race when it is set (related: #1011, #886). It has a turbulent history here — added in #228, removed in #260, re-entered during the AVAudioEngine ADM rewrite (#536); this is a clean revert. This matches the Flutter SDK's `communication` preset (`allowBluetooth`/`A2DP`/`allowAirPlay`, no `mixWithOthers`). Note the user-visible consequence: publishing the mic now interrupts other apps' audio for the rest of the call; muting doesn't bring it back.
- **Docs:** `Docs/audio.md` gains an "Audio session category selection" section covering the rules above, what counts as "recording", and the mic-permission implications.

Compared to earlier revisions of this PR:

- The permission-driven selection (`canPublishMicrophone` → `.playAndRecord`) is **gone**. `canPublish` defaults to true in standard tokens, so it made `.playback` unreachable — every subscriber would have activated a `.playAndRecord` session (mic prompt; TCC kill for apps without `NSMicrophoneUsageDescription`). The viewer-mic-prompt bug it targeted is native — the ADM's teardown path touched `AVAudioEngine.inputNode`, which alone instantiates the input unit — and is fixed by webrtc-sdk/webrtc#286, shipped in 150.7871.01 (#1103).
- The **`.ambient` category reset before deactivation** (volume-rocker register fix) was dropped: no other iOS SDK we checked (LiveKit Flutter/React Native, Stream Video, Agora, Twilio, upstream `RTCAudioSession`) resets or restores the category on teardown, and `.ambient` would break apps that manage their own category (no background audio, silent-switch muting). A restore-previous-configuration approach may follow separately.
- **`.allowAirPlay` is kept.** QA1803 only documents implied options for `.voiceChat`/`.videoChat` (`allowBluetooth`, `defaultToSpeaker`); it doesn't back removing AirPlay, and Flutter keeps it too.

### DIFF
--- a/Sources/LiveKit/Audio/AudioSessionEngineObserver.swift
+++ b/Sources/LiveKit/Audio/AudioSessionEngineObserver.swift
@@ -246,12 +246,11 @@ public class AudioSessionEngineObserver: AudioEngineObserver, Loggable, @uncheck
     /// will be publishing; otherwise `.playback` (pure listener):
     ///   - `isRecordingEnabled`: a track or external acquirer needs recording now.
     ///   - `hasRecorded`: sticky — keeps `.playAndRecord` across mute toggles.
-    ///   - `wantsToPublish`: the app declared publishing intent via
-    ///     `isRecordingAlwaysPreparedMode` AND the participant has permission
-    ///     to publish a microphone (`canPublishMicrophone`).
+    ///   - `wantsToPublish`: either signal of publishing intent —
+    ///     server-issued mic permission (`canPublishMicrophone`) or the
+    ///     app-declared `isRecordingAlwaysPreparedMode`.
     private func selectConfiguration(state: State) -> AudioSessionConfiguration {
-        // Purely permission-driven for now; intent gate disabled pending validation.
-        let wantsToPublish = state.canPublishMicrophone // && AudioManager.shared.isRecordingAlwaysPreparedMode
+        let wantsToPublish = state.canPublishMicrophone || AudioManager.shared.isRecordingAlwaysPreparedMode
         let needsRecord = state.isRecordingEnabled || state.hasRecorded || wantsToPublish
         let config: AudioSessionConfiguration = needsRecord
             ? (state.isSpeakerOutputPreferred ? .playAndRecordSpeaker : .playAndRecordReceiver)
--- a/Sources/LiveKit/Audio/AudioSessionEngineObserver.swift
+++ b/Sources/LiveKit/Audio/AudioSessionEngineObserver.swift
@@ -250,7 +250,8 @@ public class AudioSessionEngineObserver: AudioEngineObserver, Loggable, @uncheck
     ///     `isRecordingAlwaysPreparedMode` AND the participant has permission
     ///     to publish a microphone (`canPublishMicrophone`).
     private func selectConfiguration(state: State) -> AudioSessionConfiguration {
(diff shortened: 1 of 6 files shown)


Sources/LiveKit/Audio/AudioSessionEngineObserver.swift
```diff
@@ -212,6 +240,25 @@ public class AudioSessionEngineObserver: AudioEngineObserver, Loggable, @uncheck
...
+    ///
+    /// `.playAndRecord` is selected when any signal indicates the user is or
+    /// will be publishing; otherwise `.playback` (pure listener):
+    ///   - `isRecordingEnabled`: a track or external acquirer needs recording now.
+    ///   - `hasRecorded`: sticky — keeps `.playAndRecord` across mute toggles.
+    ///   - `wantsToPublish`: either signal of publishing intent —
+    ///     server-issued mic permission (`canPublishMicrophone`) or the
+    ///     app-declared `isRecordingAlwaysPreparedMode`.
+    private func selectConfiguration(state: State) -> AudioSessionConfiguration {
+        let wantsToPublish = state.canPublishMicrophone || AudioManager.shared.isRecordingAlwaysPreparedMode
```
pblazej, line 253: This is a _heuristic_ ofc, open to suggestions.

pblazej:
📱 TF build 2.15.0.b20260527

hiroshihorie (approved):

pblazej:
I think I'll postpone it till minor release with better docs/changelog.

pblazej:
cc @MaxHeimbrock maybe good to match platform audio here? (or vice versa)

MaxHeimbrock:
> cc @MaxHeimbrock maybe good to match platform audio here? (or vice versa)

Thanks for the cc. I am currently improving platform audio with Unity where Unity itself also tries to configure the audio session. This might have implications that are different from the ones in Swift. Once the Unity platform audio integration is more robust I will try to see if we can align it more with Swift.

Btw how do you test this, do you do a lot of on device testing?

pblazej:
I narrowed the scope of this PR a bit, to avoid speculative changes, focusing on:
- `playAndRecord` stickiness
- `mixWithOthers` bugfix 🐞 

The impact of resetting after call is hard to estimate atm + it can be implemented on the consumer side with moderate effort.

PR #1020, opened 2026-05-27
Audio Session Interruptions

### LINKED ISSUE #886: Audio Recording Fails to Resume System Audio After Incoming Call Interruption in Background

pblazej:
Adds interruption and route-change recovery for the iOS audio session. Closes #886 (audio not resuming after a system interruption — cellular calls, alarms, Siri, FaceTime).

- **New `LKRTCAudioSessionDelegate` adapter** re-applies the active category, mode, and options on:
  - `audioSessionDidEndInterruption` with `shouldResumeSession=true` — iOS re-activates the session after an interruption but does not restore the configuration we set.
  - `audioSessionDidChangeRoute` with `.categoryChange` / `.routeConfigurationChange` — externally-driven configuration mutations (CallKit activation, system audio takeover).
- **`overrideOutputAudioPort` reasserted on resume** for `.playAndRecord` — Apple's documented workaround for a VPIO low-volume regression where audio comes back inaudibly quiet after interruption-end.
- **No-op when `isAutomaticConfigurationEnabled = false`** (CallKit apps that drive `AVAudioSession` themselves via `CXProviderDelegate`), so this does not interfere with the documented CallKit integration pattern.
- **Selection logic reused** — the re-apply path calls `selectConfiguration(state:)` from #1019, so resume picks the same category as initial activation (sticky bit, permission gate, etc.).

Depends on #1019; merge base switches to `main` once that lands.

### DIFF
--- a/.changes/audio-session-interruptions
+++ b/.changes/audio-session-interruptions
@@ -1,0 +1,1 @@
+patch type="fixed" "iOS: audio not resuming after a system interruption (incoming call, alarm, Siri)"
--- a/Docs/audio.md
+++ b/Docs/audio.md
@@ -57,6 +57,12 @@ When `isAutomaticConfigurationEnabled` is `true`, the SDK picks the audio sessio
 
 The category is reset to `.ambient` when both playout and recording stop (typically after `Room.disconnect()`).
 
+## Interruption recovery
+
+After a system interruption (incoming phone call, alarm, Siri, FaceTime), iOS re-activates your audio session but does not restore the category, mode, or options. When `isAutomaticConfigurationEnabled` is `true`, the SDK re-applies its configuration on interruption-end and externally-driven category-change events so audio resumes in a known state.
+
+If you manage `AVAudioSession` yourself (e.g., CallKit apps with `isAutomaticConfigurationEnabled = false`), the SDK does not re-apply on these events — handle interruption recovery in your `CXProviderDelegate` instead.
+
 ## Disabling Voice Processing
 
 Apple's voice processing is enabled by default, such as echo cancellation and auto-gain control.
--- a/Sources/LiveKit/Audio/AudioSessionEngineObserver.swift
+++ b/Sources/LiveKit/Audio/AudioSessionEngineObserver.swift
@@ -101,6 +101,12 @@ public class AudioSessionEngineObserver: AudioEngineObserver, Loggable, @uncheck
         set { _state.mutate { $0.next = newValue } }
     }
 
+    // Listens to iOS-driven session events (interruption-end, externally-driven
+    // category change) that may leave the session in a state different from
+    // what we configured. Held as a strong reference because `LKRTCAudioSession`
+    // stores its delegates weakly.
+    private let rtcDelegateAdapter = RTCAudioSessionDelegateAdapter()
+
     public init() {
         _state.onDidMutate = { [weak self] new, old in
             guard let self,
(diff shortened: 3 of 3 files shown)


pblazej:
Tested basic cases like phone calls, AirPods, route changes, etc.

Let's try to break it ⚒️

pblazej:
Bringing back to draft after fresh logs.

pblazej:
Closing as the resolution isn't clear atm.

PR #1023, opened 2026-05-28
Expose RPC max round trip latency in performRpc

1egoman:
Exposes a new `maxRoundTripLatency` parameter on `LocalParticipant.performRpc` (via a new `performRpc(destinationIdentity:method:payload:responseTimeout:maxRoundTripLatency:)` overload) so that a caller can configure the amount of time they'd like to be able to tolerate between a RPC request being enqueued and a RPCRequest's ACK comes back from the remote participant.

I expect this to be a fairly uncommonly tweaked value, but exposing it means that a client can be configured to tolerate situations where RPC requests are backed up behind other messages due to webrtc head of line blocking on the data channel, and RPC requests are timing out before actually being able to be sent via webrtc.

Related android change: https://github.com/livekit/client-sdk-android/pull/953

### DIFF
--- a/Sources/LiveKit/Participant/LocalParticipant+RPC.swift
+++ b/Sources/LiveKit/Participant/LocalParticipant+RPC.swift
@@ -33,11 +33,13 @@ public extension LocalParticipant {
     ///   - method: The method name to call
     ///   - payload: The payload to pass to the method
     ///   - responseTimeout: Timeout for receiving a response after the initial connection (in seconds).
-    ///     If a value less than 8s is provided, it will be automatically clamped to 8s
-    ///     to ensure sufficient time for round-trip latency buffering.
+    ///     If a value less than `maxRoundTripLatency + 1` is provided, it will be automatically
+    ///     clamped to that floor to ensure sufficient time for round-trip latency buffering.
     ///     Default: 15s.
     /// - Returns: The response payload
     /// - Throws: RpcError on failure. Details in RpcError.message
+    /// - SeeAlso: ``performRpc(destinationIdentity:method:payload:responseTimeout:maxRoundTripLatency:)``
+    ///   for an overload that lets callers customize the ack-watchdog deadline.
     func performRpc(destinationIdentity: Identity,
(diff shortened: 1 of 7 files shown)


Sources/LiveKit/Participant/LocalParticipant+RPC.swift
```diff
@@ -49,6 +51,41 @@ public extension LocalParticipant {
...
+                    payload: String,
+                    responseTimeout: TimeInterval,
+                    maxRoundTripLatency: TimeInterval) async throws -> String
+    {
+        try await requireRoom().rpcClient.performRpc(destinationIdentity: destinationIdentity,
+                                                     method: method,
+                                                     payload: payload,
+                                                     responseTimeout: responseTimeout,
+                                                     maxRoundTripLatency: maxRoundTripLatency)
+    }
```
1egoman, line 56: Can this be combined into `performRpc(destinationIdentity:method:payload:responseTimeout:)` with an optional default for `maxRoundTripLatency`? A LLM seemed to think this would break obj-c compatibility so I left them as separate definitions.

pblazej: We had this case before; ObjC backwards compatibility is not the top priority - feel free to add the new arg as described 👍

pblazej:
Could you revert protobuf changes? It's failing due to an unrelated change in swift-protobufs (root cause is #1018)

1egoman:
Reverted in 8477bc9

Sources/LiveKit/Participant/LocalParticipant+RPC.swift
1egoman: Sweet, fixed in 4a470f16

1egoman:
@pblazej Assuming that the ci not fully passing is fine due to your previously surfaced point, anything else you'd like to see on this before it could be merged / released?

pblazej (approved):
LGTM, protocol failure is handled in #1018

Issue #998, opened 2026-05-11
App Store rejection on iOS 26: deprecated UIKit APIs in LiveKitWebRTC binary (sharedApplication, mainScreen)


n0vakovic:
Our iOS app, which depends on `client-sdk-swift` 2.14.0 (transitively, via `elevenlabs-swift-sdk` 3.1.4), was rejected by Apple App Review citing "non-public or deprecated APIs." Auditing the final binary points the deprecated UIKit selectors to the `LiveKitWebRTC.framework` shipped by `livekit/webrtc-xcframework`. Filing here because `livekit/webrtc-xcframework` has issues disabled.

> **Cross-reference:** also filed at the ElevenLabs Swift SDK, since that is our direct dependency that pulls in `client-sdk-swift`: https://github.com/elevenlabs/elevenlabs-swift-sdk/issues/176. The deprecated symbols themselves are in the binary this repo's stack ships, hence this is the more upstream report.

### Dependency / version snapshot

```
our app
  → elevenlabs-swift-sdk 3.1.4
    → client-sdk-swift 2.14.0   (this repo)
      → webrtc-xcframework 144.7559.4   (LiveKitWebRTC.framework)
```

### otool / strings evidence on `LiveKitWebRTC.framework/LiveKitWebRTC` (ios-arm64)

```
0xa2476d handleApplicationDidBecomeActive:
0xa2478f sharedApplication
...
0xa26023 setMultipleTouchEnabled:
0xa2603c mainScreen
0xa26047 scale
```

Specific call sites (from string analysis):
- `-[LKRTCCameraVideoCapturer handleApplicationDidBecomeActive:]_block_invoke` — uses `[UIApplication sharedApplication]`
- `-[LKRTCAudioSession handleApplicationDidBecomeActive:]` — uses `[UIApplication sharedApplication]`
- A class implementing `setMultipleTouchEnabled:` (Metal video view path) — uses `[[UIScreen mainScreen] scale]`

### Root cause

`UIScreen.main` / `[UIScreen mainScreen]` was deprecated in iOS 26.0 (Apple's multi-scene migration). Apple's automated review scanner now appears to flag binaries that reference it. The selectors originate in Google's upstream `libwebrtc` iOS code (e.g., `sdk/objc/components/capturer/RTCCameraVideoCapturer.m` on `main` still uses `[UIApplication sharedApplication]`). LiveKit's `webrtc-xcframework` builds Google's source, so the symbols carry through.

[...]

PR #1024, opened 2026-05-29
Fix display pixel density in adaptive stream

hiroshihorie:
## Summary

- scale adaptive-stream sizes from logical points to physical pixels on high-density displays
- add `VideoView.AdaptiveStreamPixelDensity` with `.auto` and `.fixed(Double)` controls
- resolve display scale per platform and guard invalid densities before capping to `maxDensity`

## Testing

- `swift test --filter AdaptiveStreamPixelDensityTests`

### DIFF
--- a/Sources/LiveKit/Views/VideoView.swift
+++ b/Sources/LiveKit/Views/VideoView.swift
@@ -67,6 +67,32 @@ public class VideoView: NativeView, Loggable {
         case flip
     }
 
+    /// Controls how the view's logical (point) size is scaled to physical pixels
+    /// when reporting the adaptive-stream size to the server. Server simulcast/SVC
+    /// layers are sized in physical pixels, so this avoids under-requesting (and
+    /// thus a soft/upscaled layer) on retina / high-density displays.
+    public enum AdaptiveStreamPixelDensity: Sendable, Equatable {
+        /// Use the view's own screen scale (`UIScreen.scale` on iOS/tvOS,
+        /// `NSScreen.backingScaleFactor` on macOS, `traitCollection.displayScale`
+        /// on visionOS). This is the default.
+        case auto
+        /// A fixed multiplier (e.g. `1.0`, `1.5`, `2.0`). Capped at ``maxDensity``.
+        case fixed(Double)
+
+        /// Upper bound applied to the resolved density to keep bandwidth in check.
+        public static let maxDensity: Double = 3.0
+
+        /// Resolves the effective multiplier, capped at ``maxDensity``. For
+        /// ``auto``, falls back to the supplied screen scale.
+        func resolve(screenScale: CGFloat) -> CGFloat {
+            let density: CGFloat = switch self {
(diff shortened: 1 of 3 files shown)


pblazej:
Is it a Flutter backport? How was it discovered?

hiroshihorie:
@pblazej The thought process was:
1. Video "crisp-ness" issues reported for Flutter SDK
2. Discovered that pixel density logic was not implemented in Flutter / Swift SDK but JS SDK has it
3. Implemented in Flutter SDK → Ported to Swift SDK

So if the render view size is not large enough to trigger F layer (High quality), typically H layer (mid quality) gets requested more often. (Especially on mobile devices).

Sources/LiveKit/Views/VideoView.swift
```diff
@@ -423,6 +464,24 @@ public class VideoView: NativeView, Loggable {
...
+        #elseif os(visionOS)
+        let scale = traitCollection.displayScale
+        return scale > 0 ? scale : 2.0
+        #elseif os(iOS) || os(tvOS)
+        let scale = traitCollection.displayScale
+        if scale > 0 { return scale }
+        return window?.screen.scale ?? UIScreen.main.scale
+        #else
+        return 1.0
+        #endif
```
pblazej, line 483: Per [webrtc commit 23c7882](https://github.com/livekit/webrtc/commit/23c7882d182690402428242377c7aff9dcccaa54) (driven by [#998](https://github.com/livekit/client-sdk-swift/issues/998) — App Store rejection on iOS 26 for deprecated UIKit APIs), drop the `UIScreen.main.scale` and `NSScreen.main` fallbacks here too — they're the same audit risk. Two more changes to match the webrtc fix:

- Use `window?.windowScene?.screen.scale` instead of `window?.screen.scale` — `UIWindow.screen` is itself deprecated since iOS 13.
- Use `max(scale, 1.0)` as the safety net rather than a sentinel default; Flutter shipped a `scale == 0` crash from the not-yet-attached case in [flutter#162785](https://github.com/flutter/flutter/pull/162785) and had to [revert in #166080](https://github.com/flutter/flutter/pull/166080) for exactly this reason.

```suggestion
        #if os(macOS)
        var scale = window?.backingScaleFactor ?? 0
        if scale <= 0 { scale = window?.screen?.backingScaleFactor ?? 0 }
        return max(scale, 1.0)
        #elseif os(visionOS)
        return max(traitCollection.displayScale, 1.0)
        #elseif os(iOS) || os(tvOS)
        var scale = traitCollection.displayScale
        if scale <= 0 { scale = window?.windowScene?.screen.scale ?? 0 }
        return max(scale, 1.0)
        #else
        return 1.0
        #endif
```

The existing `UIScreen.main.bounds` in `BroadcastScreenCapturer.swift:36` is the same class of issue but out of scope for this PR.

pblazej (approved):
◀️ 📹 ▶️

PR #1025, opened 2026-06-02
Fix #IsolatedConformances warning in subscribeOnMainActor

pblazej:
Run the `for try await` loop on `@concurrent` instead of `@MainActor` so Swift 6.2 (SE-0470) doesn't warn that the iterator's `AsyncIteratorProtocol` conformance may be MainActor-isolated; hop to MainActor per element and drop the unused stateful overload.


pblazej:
Fixes https://github.com/livekit/client-sdk-swift/actions/workflows/swift-snapshot.yaml

hiroshihorie (approved):

PR #1028, opened 2026-06-03
DataChannel: keep reliable seq monotonic across concurrent sends

pblazej:
## Summary

`DataChannelPair.send(dataPacket:)` assigned the reliable sequence number under `_state.mutate` but yielded onto the AsyncStream event loop **outside** that critical section. Two concurrent callers could pick `(seq=N, seq=N+1)` and yield in the reverse order — the SFU's per-publisher dedup gate then silently dropped the lower-seq packet.

Fix: move sequence assignment and serialization **into** the event-loop subscriber via a new `.sendPending` event variant, so the sequence carried on the wire matches the FIFO order in which packets reach `channel.sendData(...)`. The retry-buffer replay path keeps using `.sendRequested(PublishDataRequest)` for already-serialized entries.

## Discovery

Pre-existing issue (predates RPC v2), but surfaced during RPC v2 stress testing with large payloads — concurrent stream chunks raced for sequence numbers and one chunk's worth (~15 KB) consistently went missing:

```
🔴 [test-2] FAIL: sent=76.38 KB received=46.38 KB
🔴 [test-4] error (53.10 KB): RpcError(code: 1501, message: "Connection timeout", data: "")
🔴 [test-1] FAIL: sent=79.75 KB received=64.75 KB
```

## Test plan

- [x] New regression test `concurrentReliableSendsDeliverExactlyOnce` fires 256 reliable user-packets concurrently and asserts exact-once delivery
  - Reverting the SDK fix on this branch: **217/256** delivered (39 dropped)
  - With the fix: **256/256** delivered
- [x] Standalone CLI stress (10 concurrent `performRpc` × 1KB–1MB × 10 rounds against `wss://blaze-playground-…livekit.cloud`): **100/100** pass; pushed to 20 × 10: **200/200** pass

### DIFF
--- a/.changes/concurrent-dc-sends
+++ b/.changes/concurrent-dc-sends
@@ -1,0 +1,1 @@
+patch type="fixed" "Reliable data-channel sequence stays monotonic under concurrent sends"
--- a/Sources/LiveKit/Core/DataChannelPair.swift
+++ b/Sources/LiveKit/Core/DataChannelPair.swift
@@ -33,10 +33,15 @@ protocol DataChannelDelegate: AnyObject, Sendable {
 /// and serializes all outgoing sends through a private FIFO event loop.
 ///
 /// ## Send flow
-/// `send(dataPacket:)` builds a `PublishDataRequest` and yields a `.sendRequested`
-/// event onto the internal `AsyncStream`. The event loop's single observer is the
-/// only mutator of `Buffers` (lossy / reliable send buffers + reliable retry
-/// buffer), so all bookkeeping is race-free without locks.
+/// `send(dataPacket:)` encrypts the packet and yields a `.sendPending` event onto
+/// the internal `AsyncStream` together with its waiting continuation. The event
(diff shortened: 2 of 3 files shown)


pblazej:
Tested "manually" with swift-example between iOS and macOS (1KB-1MB payload, 15 concurrent calls):

```
🟢 [test-1] OK: 581.72 KB → participant-ydrb4kx1k8
🟢 [test-5] OK: 620.68 KB → participant-ydrb4kx1k8
🟢 [test-10] OK: 692.03 KB → participant-ydrb4kx1k8
🟢 [test-2] OK: 787.74 KB → participant-ydrb4kx1k8
🟢 [test-8] OK: 813.78 KB → participant-ydrb4kx1k8
🟢 [test-12] OK: 946.26 KB → participant-ydrb4kx1k8
```

pblazej:
Test code for the curious: https://github.com/livekit-examples/swift-example/compare/main...blaze/rpc-stress-test

hiroshihorie (approved):

PR #1029, opened 2026-06-05
Use dev server action

ladvoc:
Uses [livekit/dev-server-action](https://github.com/livekit/dev-server-action) to run SFU for end-to-end testing.


pblazej (commented):
Can we mechanically copy that to other yamls? (I believe the only one is `benchmarks`)

pblazej (approved):

PR #1031, opened 2026-06-08
DataChannel: reject oversized data messages before they break the channel

pblazej:
## Summary

Sending more than the negotiated SCTP max-message-size makes libwebrtc return success from `sendData` and then abruptly tear down the publisher data channel — every subsequent publish fails until a full reconnect.

- Parse `a=max-message-size` (RFC 8841) from the publisher answer SDP, clamp to a 64000-byte default.
- Reject in `DataChannelPair.makeRequest` after serialization (encoded byte count = what goes on the wire). `0` means "no limit".
- Log an error when the publisher channel transitions to `.closed` without an SDK-initiated reset.

Ports the fix from [livekit/rust-sdks#1137](https://github.com/livekit/rust-sdks/pull/1137). Android does the simpler hardcoded-constant variant at [`LocalParticipant.kt:984-986`](https://github.com/livekit/client-sdk-android/blob/46da6784bec2ac124972738764a8e18702a78ee3/livekit-android-sdk/src/main/java/io/livekit/android/room/participant/LocalParticipant.kt#L984-L986).

## Test plan

- [x] Unit tests: parameterized SDP parser (6 cases) + size-guard tests (`oversizedSendRejectedWithInvalidParameter`, `zeroLimitDisablesTheSizeGuard`)
- [x] `xcodebuild build` macOS clean
- [x] SwiftLint clean
- [ ] CI matrix

### DIFF
--- a/Sources/LiveKit/Core/DataChannelPair.swift
+++ b/Sources/LiveKit/Core/DataChannelPair.swift
@@ -107,6 +107,18 @@ class DataChannelPair: NSObject, @unchecked Sendable, Loggable {
         var reliableReceivedState: TTLDictionary<String, UInt32> = TTLDictionary(ttl: reliableReceivedStateTTL)
         var e2eeManager: E2EEManager?
 
+        /// Negotiated SCTP max message size (bytes), parsed from the publisher
+        /// answer SDP (`a=max-message-size`, RFC 8841). `0` means "no limit".
+        /// Defaults to ``DataChannelPair/defaultMaxMessageSize`` until the
+        /// first SDP answer is received.
+        var maxMessageSize: UInt64 = DataChannelPair.defaultMaxMessageSize
+
+        /// Set by ``DataChannelPair/reset(throwing:)``; cleared when a new
+        /// channel is assigned via ``DataChannelPair/setChannel(_:kind:)``.
+        /// Distinguishes SDK-initiated teardown from an unexpected close
+        /// (e.g. libwebrtc tearing the channel down after an oversized send).
+        var wasReset: Bool = false
+
         var isOpen: Bool {
             guard let lossy, let reliable else { return false }
             return reliable.readyState == .open && lossy.readyState == .open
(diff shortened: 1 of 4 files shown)


Sources/LiveKit/Core/DataChannelPair.swift
```diff
@@ -660,3 +717,25 @@ private extension LKRTCDataChannel {
...
+func parseSDPMaxMessageSize(_ sdp: String) -> UInt64? {
+    // `components(separatedBy: .newlines)` splits on Unicode scalars, which is
+    // what we want here — `String.split` treats CRLF as a single grapheme and
+    // would leave a stray `\r` on every line of a typical SDP.
+    for line in sdp.components(separatedBy: .newlines) {
+        let trimmed = line.trimmingCharacters(in: .whitespaces)
+        let prefix = "a=max-message-size:"
+        guard trimmed.hasPrefix(prefix) else { continue }
+        let value = trimmed.dropFirst(prefix.count).trimmingCharacters(in: .whitespaces)
+        return UInt64(value)
```
xianshijing-lk, line 738: from your testing, what is the value that is in the SDP ?

Sources/LiveKit/Core/Room+SignalClientDelegate.swift
```diff
@@ -338,6 +338,14 @@ extension Room: SignalClientDelegate {
     func signalClient(_: SignalClient, didReceiveAnswer answer: LKRTCSessionDescription, offerId: UInt32) async {
         log("Received answer for offerId: \(offerId)")
 
+        // Clamp to the SDK default — libwebrtc advertises larger (~256 KiB)
+        // than LiveKit/pion can deliver end-to-end (~64 KiB), so we trust
+        // the answer no more than our compiled-in ceiling.
+        let parsed = parseSDPMaxMessageSize(answer.sdp) ?? DataChannelPair.defaultMaxMessageSize
+        let maxMessageSize = min(parsed, DataChannelPair.defaultMaxMessageSize)
```
xianshijing-lk, line 345: do we still need the min(parsed, DataChannelPair.defaultMaxMessageSize) ? or we should just use the value from the SDP  instead ?

Sources/LiveKit/Core/DataChannelPair.swift
```diff
@@ -318,22 +330,38 @@ class DataChannelPair: NSObject, @unchecked Sendable, Loggable {
...
         } catch {
             pending.continuation.resume(throwing: error)
             return nil
         }
+
+        // Encoded size is what actually goes on the wire — sending more than
+        // the negotiated max-message-size makes libwebrtc abruptly tear the
+        // data channel down (returns success from `sendData`, then closes),
+        // breaking every subsequent publish. Reject here instead.
+        let maxMessageSize = _state.maxMessageSize
```
xianshijing-lk, line 351: should you do guard maxMessageSize > 0 else {
Log and error ? 
}
Then
guard UInt64(bytes.count) <=  maxMessageSize else {
            pending.continuation.resume(throwing: LiveKitError(
                .invalidParameter,
                message: "data packet size (\(bytes.count) bytes) exceeds the negotiated max-message-size (\(maxMessageSize) bytes)",
            ))
            return nil
        }

xianshijing-lk (commented):
some questions / nits

Sources/LiveKit/Core/DataChannelPair.swift
pblazej: `a=max-message-size:262144` in prod

Sources/LiveKit/Core/Room+SignalClientDelegate.swift
pblazej: Yes — libwebrtc/Chrome advertise ~256 KiB in their SDP answer, but LiveKit (pion) cannot deliver messages that large end-to-end, so we cap at the SDK default. Same approach as [rust-sdks](https://github.com/livekit/rust-sdks/blob/b9a700837c8be346b494b200f7dd91b8d44e9754/livekit/src/rtc_engine/rtc_session.rs#L1241-L1244).

Sources/LiveKit/Core/DataChannelPair.swift
pblazej: `maxMessageSize == 0` is the RFC 8841 "no limit" sentinel (mirrored by [rust-sdks](https://github.com/livekit/rust-sdks/blob/b9a700837c8be346b494b200f7dd91b8d44e9754/livekit/src/rtc_engine/rtc_session.rs#L1062-L1063) and pinned by the `zeroMeansNoLimit` parser test) — so the check has to skip when zero, not error. The current `if maxMessageSize != 0, UInt64(bytes.count) > maxMessageSize` expresses exactly that.

pblazej:
@xianshijing-lk the questions refer to the contract defined in https://github.com/livekit/rust-sdks/pull/1137 rather than implementation details, so I'm delaying this PR until the above is merged.

Sources/LiveKit/Core/DataChannelPair.swift
xianshijing-lk: thanks for the clarification.

pblazej:
Confirmed parity with merged [livekit/rust-sdks#1137](https://github.com/livekit/rust-sdks/pull/1137) — 64000-byte default, clamped to that ceiling when the answer SDP advertises more (e.g. libwebrtc's 256 KiB), `0` honored as RFC 8841 "no limit", and unexpected-close logged at error level. The only diff is N/A for Swift: the Rust SDK now also wires the close listener for its third `_data_track` channel, which doesn't exist here (yet).

xianshijing-lk (approved):
lgtm

hiroshihorie (approved):

PR #1042, opened 2026-06-16
ci: add App Size check job

pblazej:
Adds an `App Size` CI job that archives an empty vs LiveKit-integrated iOS app on every PR and reports the SDK's app-size impact to the job summary (per-library breakdown + a baseline budget gate that warns within tolerance and fails when exceeded).


hiroshihorie (approved):
✅ Nice

Issue #1022, opened 2026-05-28
EXC_BAD_ACCESS in LocalParticipant.set(source:enabled:) ObjC completion bridge on macOS 26.1 (mic; sibling of #1016)


utkarsh-socratic:
## Summary

On macOS 26.1, calling `room.localParticipant.set(source: .microphone, enabled: true)` reliably crashes the process with `EXC_BAD_ACCESS` inside `UnsafeContinuation.resume`. The crashing frame is the `@objc` completion-handler block that bridges the ObjC `(LocalTrackPublication?, NSError?) -> ()` callback to a Swift `UnsafeThrowingContinuation`, and the crash reproduces on both the success path (`resume(returning:)`) and the error path (`resume(throwing:)`), suggesting the continuation is being resumed more than once.

This is the microphone variant of livekit/client-sdk-swift#1016 (camera). Both reports share the same SDK version, same OS version, same `UnsafeContinuation.resume` top frame, and the same ObjC-completion-to-Swift-continuation pattern — only the source (`.camera` vs `.microphone`) and the inner closure (`setCamera` vs `set(source:enabled:)`) differ.

## Environment

| Component | Version |
| -- | -- |
| OS | macOS 26.1 (build `25B78`) |
| Hardware | `Mac16,7`, Apple Silicon (ARM-64) |
| LiveKit Swift SDK | `2.14.1` |
| WebRTC xcframework | `144.7559.06` (bundled by 2.14.1) |
| `livekit-uniffi-xcframework` | `0.0.6` |
| Swift concurrency runtime | Stock `libswift_Concurrency.dylib` shipped with macOS 26.1 |
| Build | SwiftPM executable, ad-hoc codesigned |

## Reproduction

1. SwiftPM macOS executable depending on `livekit-client` exact `2.14.1`.
2. `Info.plist` includes `NSMicrophoneUsageDescription`. Microphone permission granted.
3. Connect to a room:

   ```swift
   try await room.connect(url: serverURL, token: token)
   ```
4. Once `state == .connected`, publish the microphone:

   ```swift
   try await room.localParticipant.set(
       source: .microphone,
       enabled: true,
       captureOptions: AudioCaptureOptions(),
       publishOptions: AudioPublishOptions()
   )
   ```
5. Process is killed by `SIGSEGV` somewhere between this call entering and returning — either right after the call (success path) or after a \~10s `Code=101 "Timed out"` (error path).

[...]

Issue #1016, opened 2026-05-23
Camera publish SIGBUS in `UnsafeContinuation.resume` from `LKRTCCameraVideoCapturer` completion on macOS 26.1


zdrohC:
## Summary

On macOS 26.1 (25B78), the first call to `LocalParticipant.setCamera(enabled: true)` causes a hard `EXC_BAD_ACCESS (SIGBUS) / KERN_PROTECTION_FAILURE` crash on the `org.webrtc.RTCDispatcherCaptureSession` dispatch queue when LiveKit's `LKRTCCameraVideoCapturer` invokes its async completion bridge.

The crash reproduces on **every LiveKit version we've tested (2.10.0, 2.13.0, 2.14.1)**, on **arm64 native and x86_64 under Rosetta**, and with **all `captureOptions` / `publishOptions` combinations** (including no options at all).

The fault address is consistently wild memory (e.g. `0x6000000030` inside `GPU Carveout` reserved region; `0x305300002a` inside `SwiftUICore` `__TEXT` read-only segment). Our hypothesis: the prebuilt `LiveKitWebRTC.framework` was linked against a `libswift_Concurrency.dylib` whose `AsyncTask` layout no longer matches the runtime that ships in macOS 26.

## Environment

- **OS:** macOS 26.1 (build `25B78`), Release Type: User
- **Hardware:** `MacBookPro18,1` (Apple M1 Pro, 16 GB)
- **Swift toolchain:** the Xcode 26 / macOS 26.2 SDK toolchain (`MacOSX26.2.sdk` per stack frames)
- **LiveKit Swift SDK versions tested:**
  - `2.10.0` → bundles `LiveKitWebRTC.xcframework` `137.7151.10` → **crashes**
  - `2.13.0` → bundles `144.7559.03` → **crashes**
  - `2.14.1` → bundles `144.7559.06` → **crashes**
- **Build system:** SwiftPM, executable target, ad-hoc codesigned with `--entitlements` (sandbox off for local dev; entitlements include `cs.disable-library-validation` + `get-task-allow`)

## Reproduction

1. Create a SwiftPM macOS executable target depending on `livekit-client` exact `"2.14.1"`.
2. Configure `Info.plist` with `NSCameraUsageDescription`.
3. Connect to a room: `try await room.connect(url: serverURL, token: token)`.
4. Wait for `state == .connected`.
5. Call `try await room.localParticipant.setCamera(enabled: true)`.
[...]

PR #1044, opened 2026-06-17
Wrap ObjC auto-async calls in explicit checked continuations

### LINKED ISSUE #1036: EXC_BAD_ACCESS in Room.publisherShouldNegotiate / Room.fullConnectSequence on every connect (2.14.0 + 2.14.1, macOS arm64)
**SDK version:** `client-sdk-swift` 2.14.0 and 2.14.1 (both reproducible)
**Platform:** macOS 26.5 (Tahoe), arm64
**App:** menubar SwiftUI client embedding LiveKit via a thin wrapper SDK

## Symptom

Every connect to a `Room` crashes with `EXC_BAD_ACCESS` within ~5–8 s of `Room.connect(...)`. Two distinct stacks observed in the same SDK version, both deep inside `Room`:

### Stack A — on first SDP offer

```
SignalClient._process(signalResponse:)
 → Room.signalClient(_:didReceiveOffer:offerId:)
 → Room.publisherShouldNegotiate(force:)
 → direct field offset for Room._regionManager
 → EXC_BAD_ACCESS (KERN_INVALID_ADDRESS at 0x28)
```

### Stack B — during initial connect sequence

```
Room.fullConnectSequence(_:_:)
 → @objc closure #1 in Room.fullConnectSequence(_:_:)
 → EXC_BAD_ACCESS (KERN_INVALID_ADDRESS at 0x20)
```

The crashing thread in stack A is the main thread doing a SwiftUI layout pass; the corrupting work is on a webrtc dispatch thread (`webrtc::Thread::ProcessMessages → PhysicalSocketServer::WaitPoll`) that the `Room` delegate hop returns into.

## Memory corruption, not a null deref

Bad-access addresses differ across reproductions: `0x28`, `0x10001002a`, `0x7a326d303934685d` (ASCII garbage, looks like a stomped value), `0x20`. Same SDK version, same call site. That pattern points at a use-after-free of a `Room`-owned object (likely `_regionManager`, which appears as a `direct field offset for` symbol in stack A) rather than a static null pointer.

## What rules out our deployment

Reproduced identically against three independent LiveKit servers:

1. LiveKit Cloud (production environment)
2. LiveKit Cloud (staging environment)
3. Self-hosted `livekit-server` via the project's own dev docker-compose, fronted by a local realtime worker registering against `ws://localhost:7880`

Same SDK version, same call stack, same `0x28` subtype on stack A regardless of which server is on the other end. Server-side ICE/TURN/SFU is therefore not the variable.

## Reproducibility

100% — five `.ips` files from a single afternoon, every voice session attempted ended this way. The connect path always reaches `Room.publisherShouldNegotiate` (stack A) or dies during `fullConnectSequence` (stack B); there is no successful session on this build.

## Tried

- 2.14.1 → crash (stack A consistently)
- 2.14.0 → crash (stack A primarily, stack B once)
- 2.13.0 → could not test; SDK removed `LiveKitSDK.setTracing(_:)` and the `Span(label:)` initializer used by our `ConnectTracing`, so it does not compile against our codebase without source changes.

Happy to share `.ips` files privately if useful — they contain workspace IDs / user identifiers that I'd rather not paste in a public issue.

## Environment

- SwiftPM, Swift 6 language mode
- LiveKit dependency surface used: `Room.connect(...)`, `ConnectOptions(enableMicrophone: true)` (publish-during-join), `Room.prepareConnection` for prewarm, `LiveKitSDK.setLogLevel`, `LiveKitSDK.setTracing`
- Transitive: `livekit-uniffi-xcframework` 0.0.6, `webrtc-xcframework` 144.7559.6

elesahich:
## Summary

Wraps two compiler-synthesized ObjC *auto-async* calls in explicit `withCheckedThrowingContinuation`:

- `Transport._iceCandidatesQueue` → `_pc.add(iceCandidate.toRTCType())` (`addIceCandidate:completionHandler:`)
- `CameraCapturer.startCapture()` → `capturer.startCapture(with:format:fps:)` (`startCaptureWithDevice:format:fps:completionHandler:`)

Both now call the completion-handler API directly inside an explicit checked continuation, matching the pattern already used in this file for `setRemoteDescription` / `setLocalDescription` / `createOffer`.

## Motivation

When an app statically links/merges modules built in **mixed Swift language modes** (some Swift 5, some Swift 6), the compiler-generated *unsafe* continuation thunk for an imported ObjC completion-handler method is a signature-keyed shared symbol. The unsafe thunk emitted by a Swift-5 module can win symbol deduplication over the checked thunk this SDK (built in Swift 6) intends to use. At runtime the affected resume paths are silently downgraded to unsafe.

Under concurrent peer-connection negotiation, the downgraded unsafe continuation corrupts the Swift task allocator, surfacing as `EXC_BAD_ACCESS (SIGSEGV)`:

```
UnsafeContinuation.resume(returning:)
  ← @objc completion handler block (NSError?) -> ()
  ← LiveKitWebRTC
... StackAllocator::getSlabForAllocation / flagAsAndEnqueueOnExecutor
```

We hit this reliably on real devices (iOS 18 and iOS 26) on `connect` (addIceCandidate) and on enabling the camera (startCapture). The two sites here were the only remaining `_pc.`/`capturer.` calls going through the auto-async overload; every other peer-connection call in `Transport` already uses an explicit continuation, which is why those were unaffected.

## Why this fix

Calling the completion-handler API explicitly means the shared auto-async thunk is **never generated or referenced**, so there is no symbol for cross-module dedup to collapse. The resume stays checked regardless of how downstream modules are built — no behavior change for callers, just a safer continuation path.

## Testing

Verified on device (iPhone 12 / iOS 18, iPhone 17 Pro / iOS 26): connect succeeds and camera enables without the crash; previously both reproduced consistently.

### DIFF
--- a/.changes/objc-async-checked-continuation
+++ b/.changes/objc-async-checked-continuation
@@ -1,0 +1,1 @@
+patch type="fixed" "Wrap ObjC completion-handler async calls in explicit checked continuations (fixes mixed Swift 5/6 continuation-bridge EXC_BAD_ACCESS)"
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -130,6 +130,7 @@ private static let playAndRecordOptions: AVAudioSession.CategoryOptions = [.mixW
 - Only `Task.discarding` preserves caller actor isolation in its closure; other `Support/Async/` helpers (`Task.retrying`, `AsyncSerialDelegate.notifyAsync`, etc.) run their bodies nonisolated — hop explicitly with `await MainActor.run { ... }` when UI-bound work is needed
 - Use async primitives in `Support/Async` and `Support/Schedulers` when operation order matters
 - Prefer native Swift async/await over `Combine` for new code
+- Until the minimum supported compiler is Swift 6.3, wrap calls to imported Objective-C completion-handler methods made via their synthesized `async` overload in an explicit `withCheckedThrowingContinuation`, since the bare auto-bridge hits a mixed Swift 5/6 thunk-coalescing crash (swiftlang/swift#81846) fixed in 6.3
 
 ### Error Handling
 
--- a/Sources/LiveKit/Core/Transport.swift
+++ b/Sources/LiveKit/Core/Transport.swift
@@ -68,7 +68,12 @@ actor Transport: NSObject, Loggable {
         guard let self else { return }
 
         do {
-            try await _pc.add(iceCandidate.toRTCType())
+            let rtcCandidate = iceCandidate.toRTCType()
+            try await withCheckedThrowingContinuation { (continuation: CheckedContinuation<Void, Error>) in
+                _pc.add(rtcCandidate) { error in
+                    if let error { continuation.resume(throwing: error) } else { continuation.resume() }
+                }
+            }
         } catch {
             log("Failed to add(iceCandidate:) with error: \(error)", .error)
         }
--- a/Sources/LiveKit/Core/Transport.swift
+++ b/Sources/LiveKit/Core/Transport.swift
@@ -70,7 +70,7 @@ actor Transport: NSObject, Loggable {
         do {
             let rtcCandidate = iceCandidate.toRTCType()
             try await withCheckedThrowingContinuation { (continuation: CheckedContinuation<Void, Error>) in
-                _pc.add(rtcCandidate) { error in
+                self._pc.add(rtcCandidate) { error in
                     if let error { continuation.resume(throwing: error) } else { continuation.resume() }
                 }
             }
--- a/Sources/LiveKit/Track/Capturers/CameraCapturer.swift
+++ b/Sources/LiveKit/Track/Capturers/CameraCapturer.swift
@@ -309,7 +309,11 @@ public class CameraCapturer: VideoCapturer, @unchecked Sendable {
         // Already stopped
         guard didStop else { return false }
 
-        await capturer.stopCapture()
+        await withCheckedContinuation { (continuation: CheckedContinuation<Void, Never>) in
+            capturer.stopCapture {
+                continuation.resume()
+            }
+        }
 
         // Update internal vars
         set(dimensions: nil)
--- a/Sources/LiveKit/Track/Capturers/CameraCapturer.swift
+++ b/Sources/LiveKit/Track/Capturers/CameraCapturer.swift
@@ -285,7 +285,11 @@ public class CameraCapturer: VideoCapturer, @unchecked Sendable {
 
         log("starting camera capturer device: \(device), format: \(selectedFormat), fps: \(selectedFps)(\(fpsRange))", .info)
 
-        try await capturer.startCapture(with: device, format: selectedFormat.format, fps: selectedFps)
+        try await withCheckedThrowingContinuation { (continuation: CheckedContinuation<Void, Error>) in
+            capturer.startCapture(with: device, format: selectedFormat.format, fps: selectedFps) { error in
+                if let error { continuation.resume(throwing: error) } else { continuation.resume() }
+            }
+        }
 
         // Update internal vars
         _cameraCapturerState.mutate {
--- a/Sources/LiveKit/Track/Capturers/InAppCapturer.swift
+++ b/Sources/LiveKit/Track/Capturers/InAppCapturer.swift
@@ -39,11 +39,15 @@ public class InAppScreenCapturer: VideoCapturer, @unchecked Sendable {
         guard didStart else { return false }
 
         // TODO: force pixel format kCVPixelFormatType_420YpCbCr8BiPlanarFullRange
-        try await RPScreenRecorder.shared().startCapture { [weak self] sampleBuffer, type, _ in
-            guard let self else { return }
-            // Only process .video
-            if type == .video {
-                capture(sampleBuffer: sampleBuffer, capturer: capturer, options: options)
+        try await withCheckedThrowingContinuation { (continuation: CheckedContinuation<Void, Error>) in
+            RPScreenRecorder.shared().startCapture { [weak self] sampleBuffer, type, _ in
+                guard let self else { return }
+                // Only process .video
+                if type == .video {
+                    capture(sampleBuffer: sampleBuffer, capturer: capturer, options: options)
+                }
+            } completionHandler: { error in
+                if let error { continuation.resume(throwing: error) } else { continuation.resume() }
             }
         }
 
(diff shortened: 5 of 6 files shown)


pblazej:
Thanks @elesahich — confirmed the root cause, reproduced it, and pinned the affected range. The fix is right; a few things before merge.

## Root cause

swiftlang/swift#81846: calling an imported ObjC completion-handler method via its
synthesized `async` overload emits a weak, signature-keyed bridge thunk. **Swift 5
mode emits the *unchecked* variant, Swift 6 the *checked* one — under the same
symbol name.** A mixed Swift 5/6 static link coalesces them → resume through the
wrong layout → `EXC_BAD_ACCESS` in `swift_continuation_resumeImpl`. An explicit
`withCheckedContinuation` emits no such thunk, so wrapping fixes it.

## Reproduced + affected range

Repro: an ObjC `()->()` method bridged from a `.swiftLanguageMode(.v5)` and a
`.v6` caller, statically linked.

| Swift | `auto` (unwrapped) | wrapped |
|---|---|---|
| 6.1.2 / 6.2.0 / 6.2.3 | **SIGSEGV** | clean |
| 6.3.0 | clean | clean |

Affected on **Swift 6.1.x–6.2.3 (Xcode 16.3–26.x)**; fixed in **6.3 (Xcode 26.4+)** — no 6.2 backport.

## Suggestions

1. **Wrap all auto-bridge consumption sites, not just two** — every `await` into an
   imported ObjC completion method is exposed. There are 7:
   - `Transport.swift:71` `_pc.add(iceCandidate)` ✅ in PR
   - `CameraCapturer.swift:288` `capturer.startCapture(...)` ✅ in PR
   - `CameraCapturer.swift:308` `capturer.stopCapture()` ❌ (the `()->()` shape; #1016's workaround included it)
   - `MacOSScreenCapturer.swift:121` `stream.startCapture()` ❌
   - `MacOSScreenCapturer.swift:143` `stream.stopCapture()` ❌
   - `MacOSScreenCapturer.swift:452` `SCShareableContent.excludingDesktopWindows(...)` ❌
   - `InAppCapturer.swift:42` `RPScreenRecorder.shared().startCapture { … }` ❌

   Fine to take the four Apple-framework ones in a fast-follow if you'd rather keep this WebRTC-scoped.

2. **Add an AGENTS.md rule** under "Concurrency and State":
   > Until the minimum supported compiler is Swift 6.3, wrap calls to imported Objective-C completion-handler methods made via their synthesized `async` overload in an explicit `withCheckedThrowingContinuation`, since the bare auto-bridge hits a mixed Swift 5/6 thunk-coalescing crash (swiftlang/swift#81846) fixed in 6.3.

3. **Add a `.changes` entry** `.changes/objc-async-checked-continuation`:
   ```
   patch type="fixed" "Wrap ObjC completion-handler async calls in explicit checked continuations (fixes mixed Swift 5/6 continuation-bridge EXC_BAD_ACCESS)"
   ```

4. **Confirm the build toolchain** — the trigger is the compiler the app is built
   with, not the device OS. Which Xcode did you build the crashing app with?

## Linked issues

- **#1016 (camera)** → the `startCapture` site, fixed here.
- **#1022 (mic)** → no separate site; reaches a bridge only via the shared `addIceCandidate` path, covered here.
- **#1036** → reported on Xcode 26.5 (Swift 6.3, the *fixed* compiler), where this can't occur.

Relates to #1016, #1022, #1036.

pblazej (changes requested):
Happy to merge after addressing the above 🎉

elesahich:
Thanks for the review — the repro table and affected range were helpful. Addressed all of it:

**All auto-bridge sites wrapped** (pushed): added the remaining five alongside the two WebRTC ones —
- `CameraCapturer.swift` `capturer.stopCapture()` (void completion → `withCheckedContinuation`)
- `MacOSScreenCapturer.swift` `stream.startCapture()`
- `MacOSScreenCapturer.swift` `stream.stopCapture()`
- `MacOSScreenCapturer.swift` `SCShareableContent.getExcludingDesktopWindows(...)`
- `InAppCapturer.swift` `RPScreenRecorder.shared().startCapture(handler:completionHandler:)`

Each calls the completion-handler API inside an explicit checked continuation, so no synthesized async-bridge thunk is emitted.

**AGENTS.md** — added the rule under "Concurrency and State".

**`.changes/objc-async-checked-continuation`** — added.

**Toolchain** — the crashing app was built with Xcode 26.2 (Swift 6.2.3), in the affected range. Our app statically merges ~170 Swift-5-mode modules with this SDK (Swift 6), which is the mixed-mode coalescing condition. The wrapping stays harmless once we reach 6.3.

elesahich:
Update on the remaining sites:

Wrapped the iOS-relevant bridges — `Transport.addIceCandidate`, `CameraCapturer.startCapture` / `stopCapture`, and `InAppCapturer` (ReplayKit). 

I backed the three `MacOSScreenCapturer` sites out for now:
- `getExcludingDesktopWindows` returns a non-Sendable `SCShareableContent`; a manual continuation surfaces a region-isolation "sending 'content' risks data races" error that the synthesized bridge handles for us.
- Wrapping `SCStream.startCapture` tips the already-long `startCapture()` past SwiftLint's `function_body_length` (54 > 50).

Both are cold macOS screen-capture paths — happy to take them as a focused fast-follow (the SCShareableContent one likely needs a small `nonisolated(unsafe)` hand-off, and `startCapture()` a short extract or an explicit disable). Let me know if you'd prefer them in this PR instead.

Toolchain (from before): the crashing app builds with Xcode 26.2 / Swift 6.2.3, statically merged with ~170 Swift-5-mode modules — the mixed-mode coalescing condition.

CI is green except `Build & Test (iOS Simulator, iPhone 16 Pro, OS 18.5)`, which fails in `RpcClientTests` with `RpcError(1502, "Response timeout")` — unrelated to this change (the `RoomTests` suite passes, the iPhone 17 Pro / iOS 26.5 sim passes, and the same job passed on the first revision). I can't re-run it from the fork; a re-run should clear it.

pblazej (approved):
Looks great — approving; just to close the loop, can you confirm the crash no longer reproduces once the app is rebuilt with Swift 6.3 (Xcode 26.4+)?

elesahich:
Thanks for approving. On the 6.3 confirmation — I can't verify it directly yet; our toolchain is pinned at Xcode 26.2 / Swift 6.2.3 and we don't have 26.4+ here.

That said, your repro table already shows the bare (unwrapped) call is clean on 6.3.0, which matches the swiftlang/swift#81846 fix, and the explicit continuations here stay correct on 6.3 — so the wrapping is a harmless no-op there rather than a behavior change. I'll confirm on-device once we move to Xcode 26.4+ and report back.

Issue #815, opened 2025-10-14
Blocking engine availability method call when microphone permission is not determined


silviudeac:
Our phone application uses CallKit and the LiveKit iOS Client SDK for making and receiving phone calls.

Calling AudioManager's "setEngineAvailability(.default)" when the microphone permission is not determined (AVAuthorizationStatus.notDetermined) will block the calling thread.

Example usage:
```
class CallKitDelegate: CXProviderDelegate {

    func provider(_ provider: CXProvider, didActivate audioSession: AVAudioSession) {
        do {
            try AudioManager.shared.setEngineAvailability(.default)
        } catch {
            print("Error setting engine availability: \(error)")
        }
    }
}
```

This will block until the user gives his permission (a permission dialog will appear if the application is running in the foreground).
This is acceptable for when the app is in foreground, but consider that the application may be running in the background when an incoming call arrives in which case the permission dialog will not appear and the application will be blocked.

We can implement a workaround but I believe it would be preferable for the method to throw rather than block.

**SDK Version**
2.8.1

**iOS/macOS Version**
iOS 26.0.1

**Xcode Version**
Xcode 26.0, Swift 6.2

**Steps to Reproduce**
Requirements:
- an iOS application that uses the CallKit SDK 
- microphone permission not determined (user was not asked yet)
- call 'AudioManager.shared.setEngineAvailability(.default)' after CallKit activates the AVAudioSession and reports it in it's delegate.

**Expected behavior**
Throw an error if the setEngineAvailability cannot complete successfully due to the missing microphone permission. 

**Screenshots**
<img width="1365" height="670" alt="Image" src="https://github.com/user-attachments/assets/79dbc656-7c26-41d9-bb50-5ab767c4d191" />

hiroshihorie:
Thank you for using this new API and reporting this issue.
I will patch this, in addition to [other improvements](https://github.com/livekit/client-sdk-swift/pull/808).

hiroshihorie:
Update: will remove this from the WebRTC lib and move it to the SDK with app foreground checks
https://github.com/webrtc-sdk/webrtc/pull/204

pblazej:
@hiroshihorie I think we can merge it and resolve ⏫ 🙃

silviudeac:
Opened this [PR](https://github.com/livekit/client-sdk-swift/pull/1047) for the SDK side.

[...]

PR #204, opened 2023-04-12
[🤖 readme-manager] Update README

Ocupe:
This PR has been automatically generated by the readme manager. Please check it and if all is well, merge it. Thank you, my human friend. 🤖

PR #1047, opened 2026-06-22
Request microphone permission in foreground before enabling recording.

silviudeac:
This is the SDK-side follow-up to **webrtc-sdk/webrtc#204**. That change makes the WebRTC Audio Device Module (ADM) stop implicitly requesting microphone permission — the blocking request that hangs a worker thread when the app is woken in the background (#815) — and replaces it with a passive authorized-check. With the ADM no longer prompting, the SDK now does it: `LocalParticipant` proactively requests access at its async mic-enable path, but only while the app is foregrounded so the system dialog can actually appear.
  
The two are meant to land together — #204 removes the hang, this PR restores the prompt.
  
## What changed
  
- **`LiveKit+DeviceHelpers.swift`** — new `LiveKitSDK.ensureDeviceAccessIfForegrounded(for:)`: requests access only when the app is `.active`, reusing the existing non-blocking `ensureDeviceAccess(for:)`. A no-op on non-active / non-UIKit contexts.
- **`LocalParticipant.set(source:enabled:)`** — in the `.microphone` create branch, before `LocalAudioTrack.createTrack(...)`:  
    
```swift
if AVCaptureDevice.authorizationStatus(for: .audio) == .notDetermined {
    _ = await LiveKitSDK.ensureDeviceAccessIfForegrounded(for: [.audio])
}
```
 
Only .notDetermined triggers a request; backgrounded/extension contexts never prompt; muting/unmuting an existing track is untouched. No change to existing API signatures.

Refs #815 (full fix together with webrtc-sdk/webrtc#204).

### DIFF
--- a/.changes/mic-permission-foreground
+++ b/.changes/mic-permission-foreground
@@ -1,0 +1,1 @@
+patch type="fixed" "Request microphone permission in foreground before enabling recording (#815)"
--- a/Sources/LiveKit/LiveKit+DeviceHelpers.swift
+++ b/Sources/LiveKit/LiveKit+DeviceHelpers.swift
@@ -16,6 +16,10 @@
 
 import AVFoundation
 
+#if canImport(UIKit) && (os(iOS) || os(visionOS) || os(tvOS))
+import UIKit
+#endif
+
 public extension LiveKitSDK {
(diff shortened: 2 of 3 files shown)


pblazej:
Closing in favor of #1085 cc @hiroshihorie

PR #1048, opened 2026-06-22
feat(audio): add AudioProcessingOptions

hiroshihorie:
## Summary

Adds an explicit, per-effect audio processing API so apps can choose how each effect runs, either Apple's platform Voice Processing I/O on the device or WebRTC's software processing, instead of relying on implicit toggles. This also adopts the webrtc-sdk audio processing state v2 contract for accurate, engine-wide diagnostics.

### Motivation

Until now, software voice processing was reached indirectly by disabling platform voice processing, and the exact behavior changed across releases. Some apps deliberately avoid Apple VPIO (for consistent hardware volume, faster call start, screen recording with audio, and no system mute sounds) while still wanting echo cancellation and noise suppression in software. This PR makes that an explicit, stable choice.

## New API

The new surface is Swift-only. The shipped Objective-C surface of `AudioCaptureOptions` (its `Bool` properties and `initWithEchoCancellation:...` initializer) is unchanged.

### Per-effect mode types

Each effect has its own mode type: `EchoCancellationMode`, `NoiseSuppressionMode`, `AutoGainControlMode`, and `HighpassFilterMode`, selecting the implementation:

- `.automatic` (default): prefer platform processing when available, fall back to WebRTC software otherwise.
- `.platform`: use platform processing only. Rejected if the platform implementation is unavailable.
- `.software`: force WebRTC software processing and disable the matching platform effect when possible.

There is no platform high-pass filter, so `HighpassFilterMode` offers only `.automatic` and `.software` — requesting a platform HPF is unrepresentable at compile time.

### `AudioProcessingOptions`

An immutable `Hashable`, `Sendable` struct describing all four effects (echo cancellation, noise suppression, auto gain control, high-pass filter) with an enabled flag and a mode each. The default initializer produces the communication profile (EC/AGC/NS enabled, HPF disabled, all modes `.automatic`); an `AudioProcessingOptions.noProcessing` preset disables everything.

### `AudioCaptureOptions`

Gains matching `*Mode` fields (`echoCancellationMode`, `autoGainControlMode`, `noiseSuppressionMode`, `highpassFilterMode`, following the existing Swift `highpassFilter` spelling). Existing call sites keep working since modes default to `.automatic`.

### Runtime control

`LocalAudioTrack.setAudioProcessingOptions(_:)` applies options to an already-published track. On success it returns `AudioProcessingOptionsResult` (`.applied`, or `.stored` when no sender is active yet); on failure it throws `AudioProcessingOptionsError`, whose `code` carries the reason (`.invalidState`, `.invalidCombination`, `.platformUnavailable`, `.applyFailed`). Success and failure states are separate types, so neither can express the other.

### Diagnostics

`AudioManager.audioProcessingState` exposes the v2 state, following the `requested -> resolved -> active -> effective` vocabulary, read from the factory-owned, engine-wide module rather than a single peer connection. Each component reports its last typed request, per-path detail (`software` and `platform` resolved/active flags, with `platform` being `nil` on devices without a built-in implementation), and the `effective` implementation.

`AudioManager.platformVoiceProcessingState` separately exposes device-level Apple Voice Processing I/O state: per-effect availability/requested/active, coupling topology, and the VPIO unit's enabled/bypassed/AGC flags as requested/active pairs.

### Renamed API

`AudioManager.setVoiceProcessingEnabled(_:)` and `isVoiceProcessingEnabled` are renamed to `setPlatformVoiceProcessingAllowed(_:)` and `isPlatformVoiceProcessingAllowed`, matching the underlying ADM accessors and the "allowed" policy meaning. The old names remain as deprecated, renamed forwarders, so existing code keeps compiling with a fix-it.

## Usage

Publish the microphone with all effects forced to software:

```swift
let options = AudioCaptureOptions(
    echoCancellation: true,
    autoGainControl: true,
    noiseSuppression: true,
    highpassFilter: true,
    echoCancellationMode: .software,
    autoGainControlMode: .software,
    noiseSuppressionMode: .software,
    highpassFilterMode: .software
)
try await room.localParticipant.setMicrophone(enabled: true, captureOptions: options)
```

Set it as the room default so it applies whenever the mic track is published:

```swift
try await room.connect(url: url, token: token,
                       roomOptions: RoomOptions(defaultAudioCaptureOptions: options))
```

Guarantee Apple VPIO is never used, then rely on software processing:

```swift
try AudioManager.shared.setPlatformVoiceProcessingAllowed(false)
```

Verify what each effect resolved to:

```swift
let state = AudioManager.shared.audioProcessingState
print(state.echoCancellation.effective) // Software
print(state.noiseSuppression.effective)  // Software
```

## Documentation

`Docs/audio.md` gains a new section, "Audio Processing Modes (software, platform, automatic)", covering publish-time, room-default, and runtime configuration, plus how to verify the resolved implementation. The "Disallowing Platform Voice Processing" section is updated to the renamed API.

## Example app

A companion example demonstrates the full surface, including a live "Voice Processing" and "Runtime Audio Processing" panel with effective-state diagnostics:

- Branch: https://github.com/livekit-examples/swift-example/tree/hiroshi/runtime-vp
- Audio controls UI: [`Multiplatform/Views/AudioControlsPanel.swift`](https://github.com/livekit-examples/swift-example/blob/hiroshi/runtime-vp/Multiplatform/Views/AudioControlsPanel.swift)
- Wiring: [`Multiplatform/Controllers/AppContext.swift`](https://github.com/livekit-examples/swift-example/blob/hiroshi/runtime-vp/Multiplatform/Controllers/AppContext.swift)

That branch pins its SDK dependency to this branch (`hiroshi/runtime-vp`).

## Notes

- Requires `LiveKitWebRTC` `144.7559.11` (already on `main`).
- Room defaults apply at track creation. An already-published mic track is updated through `setAudioProcessingOptions(_:)`, not by toggling the mic.
- Runtime options apply to the engine-wide audio processing module; conflicting updates from multiple local audio tracks are last-writer-wins.

## Test plan

- [x] macOS build passes
- [x] `AudioProcessingOptionsTests` and Objective-C compatibility tests pass
- [x] Generated Objective-C header inspected: shipped `AudioCaptureOptions` selector preserved, no new Swift-only types exposed
- [ ] Example app builds and runs on macOS against the final API

### DIFF
--- a/Docs/audio.md
+++ b/Docs/audio.md
@@ -46,6 +46,68 @@ AudioManager.shared.audioSession.isAutomaticDeactivationEnabled = false
 
 When set to `false`, the audio session remains active after the LiveKit call ends, preserving your app's audio state.
 
+## Audio Processing Modes (software, platform, automatic)
+
+Each audio processing effect (echo cancellation, noise suppression, auto gain control, and high-pass filter) can run in one of two ways. It can use Apple's platform Voice Processing I/O on the device, or WebRTC's software processing. You choose per effect with an `AudioProcessingMode`:
+
+- `.automatic` (default): prefer platform voice processing when available, and fall back to WebRTC software processing otherwise.
+- `.platform`: use platform voice processing only. If the platform implementation is unavailable, the request is rejected.
+- `.software`: force WebRTC software processing and disable the matching platform effect when possible.
+
+Configure the modes when you publish the microphone:
+
+```swift
+let options = AudioCaptureOptions(
+    echoCancellation: true,
+    autoGainControl: true,
+    noiseSuppression: true,
+    highpassFilter: true,
+    echoCancellationMode: .software,
+    autoGainControlMode: .software,
+    noiseSuppressionMode: .software,
+    highPassFilterMode: .software
+)
+try await room.localParticipant.setMicrophone(enabled: true, captureOptions: options)
+```
+
+You can also set them as the room default, so they apply whenever the mic is published:
+
+```swift
+let room = Room()
+try await room.connect(url: url, token: token,
+                       roomOptions: RoomOptions(defaultAudioCaptureOptions: options))
+```
+
+To change the modes at runtime on an already-published track, use `setAudioProcessingOptions`:
+
+```swift
+let result = try localAudioTrack.setAudioProcessingOptions(
+    AudioProcessingOptions(
+        echoCancellation: true,
+        autoGainControl: true,
+        noiseSuppression: true,
+        highPassFilter: true,
+        echoCancellationMode: .software,
+        autoGainControlMode: .software,
+        noiseSuppressionMode: .software,
+        highPassFilterMode: .software
+    )
+)
+print(result.code) // .applied or .stored on success
+```
+
+To confirm which implementation each effect resolved to, read the engine-wide state:
+
+```swift
+let state = AudioManager.shared.audioProcessingState
+print(state.echoCancellation.effective) // Software
+print(state.noiseSuppression.effective)  // Software
+```
+
+> **NOTE**: There is no platform high-pass filter, so `.platform` is rejected for `highPassFilterMode`. Use `.software` instead.
+
+To guarantee Apple Voice Processing I/O is never used at all, for example to keep hardware volume consistent or to allow screen recording with audio, also disallow platform voice processing as described below.
+
 ## Disallowing Platform Voice Processing
 
 Apple's platform voice processing is allowed by default, such as echo cancellation and auto-gain control.
--- a/Docs/audio.md
+++ b/Docs/audio.md
@@ -46,25 +46,25 @@ AudioManager.shared.audioSession.isAutomaticDeactivationEnabled = false
 
 When set to `false`, the audio session remains active after the LiveKit call ends, preserving your app's audio state.
 
-## Disabling Voice Processing
+## Disallowing Platform Voice Processing
 
-Apple's voice processing is enabled by default, such as echo cancellation and auto-gain control.
+Apple's platform voice processing is allowed by default, such as echo cancellation and auto-gain control.
 
-If your app doesn't require voice processing at all, you can disable it entirely:
+If your app must not use Apple Voice Processing I/O, disable voice processing:
 
 ```swift
 try AudioManager.shared.setVoiceProcessingEnabled(false)
 ```
 
-This restarts the internal `AVAudioEngine` to apply the change. It can cause a short audio glitch, so it is recommended to set it once before connecting to a Room. Disabling voice processing also disables muted speaker detection.
+This restarts the internal `AVAudioEngine` when an Apple VPIO path is active. It is recommended to set it once before connecting to a Room. Runtime `AudioProcessingOptions` with `automatic` mode will fall back to WebRTC software processing while platform voice processing is disallowed.
 
-If your app requires toggling voice processing at run-time, it is recommended to use:
+For per-track or per-capture software processing, use `AudioProcessingOptions` with `.software` modes. The lower-level bypass API remains available when you need to directly control Apple VPIO:
 
 ```swift
 AudioManager.shared.isVoiceProcessingBypassed = true
 ```
 
-Set it back to `false` to re-enable processing. This uses `AVAudioEngine`'s [isVoiceProcessingBypassed](https://developer.apple.com/documentation/avfaudio/avaudioinputnode/isvoiceprocessingbypassed) and works seamlessly at run-time.
+Set it back to `false` to re-enable the Apple path. This uses `AVAudioEngine`'s [isVoiceProcessingBypassed](https://developer.apple.com/documentation/avfaudio/avaudioinputnode/isvoiceprocessingbypassed). Runtime `AudioProcessingOptions` can overwrite this Apple-specific state when capture starts or options are reapplied.
 
 ## Other audio ducking
 
--- a/Docs/audio.md
+++ b/Docs/audio.md
@@ -65,7 +65,7 @@ let options = AudioCaptureOptions(
     echoCancellationMode: .software,
     autoGainControlMode: .software,
     noiseSuppressionMode: .software,
-    highPassFilterMode: .software
+    highpassFilterMode: .software
 )
 try await room.localParticipant.setMicrophone(enabled: true, captureOptions: options)
 ```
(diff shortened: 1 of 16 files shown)


Docs/audio.md
```diff
@@ -46,25 +46,87 @@ AudioManager.shared.audioSession.isAutomaticDeactivationEnabled = false
...
 
-## Disabling Voice Processing
+## Audio Processing Modes (software, platform, automatic)
 
-Apple's voice processing is enabled by default, such as echo cancellation and auto-gain control.
+Each audio processing effect (echo cancellation, noise suppression, auto gain control, and high-pass filter) can run in one of two ways. It can use Apple's platform Voice Processing I/O on the device, or WebRTC's software processing. You choose per effect with an `AudioProcessingMode`:
 
-If your app doesn't require voice processing at all, you can disable it entirely:
+- `.automatic` (default): prefer platform voice processing when available, and fall back to WebRTC software processing otherwise.
+- `.platform`: use platform voice processing only. If the platform implementation is unavailable, the request is rejected.
```
pblazej, line 54: Nit: I'd just call it `.platformOnly`/`.softwareOnly` (self-explanatory)

Docs/audio.md
```diff
@@ -46,25 +46,87 @@ AudioManager.shared.audioSession.isAutomaticDeactivationEnabled = false
...
+- `.software`: force WebRTC software processing and disable the matching platform effect when possible.
+
+Configure the modes when you publish the microphone:
+
+```swift
+let options = AudioCaptureOptions(
```
pblazej, line 60: Maybe it's worth approaching in a way similar to e2ee options? Deprecate the old ones and use explicit enums for each set of values `EchoCancellationMode` etc.?

```swift
public enum EchoCancellationMode: Sendable { case off, automatic, platform, software }
```

Then you also avoid this problem (inability to model):

> There is no platform high-pass filter

Also static helpers for `.software`, `.platform` etc.

Docs/audio.md
pblazej: Or maybe `.hardwareOnly` is even better.

Sources/LiveKit/Track/Local/LocalAudioTrack.swift
```diff
@@ -90,12 +94,32 @@ public class LocalAudioTrack: Track, LocalTrackProtocol, AudioTrackProtocol, @un
         try await super._unmute()
     }
 
+    /// Updates this local track's voice processing options without restarting capture.
+    ///
+    /// If this track is already published, WebRTC reapplies the updated options through
+    /// the active sender. Effective APM configuration is shared by the WebRTC voice engine,
+    /// so conflicting updates from multiple local audio tracks are last-writer-wins.
+    @discardableResult
+    public func setAudioProcessingOptions(_ options: AudioProcessingOptions) throws -> AudioProcessingOptionsResult {
```
pblazej, line 106: `throws` + Result can look a bit awkward, some options:
  - non-throwing -> AudioProcessingOptionsResult, caller inspects .code, or
  - keep throwing but carry the code on the thrown error (typed reason), not just interpolated into a string.

Sources/LiveKit/Types/AudioProcessingOptions.swift
```diff
@@ -0,0 +1,379 @@
...
+        case .platform: .platform
+        case .softwareAndPlatform: .softwareAndPlatform
+        @unknown default: .unknown
+        }
+    }
+}
+
+extension AudioProcessingOptionsResultCode {
+    init(_ code: LKRTCAudioProcessingOptionsResultCode) {
+        self = Self(rawValue: code.rawValue) ?? .applyFailed
```
pblazej: Nit: switch would be more robust.

Docs/audio.md
```diff
@@ -46,25 +46,87 @@ AudioManager.shared.audioSession.isAutomaticDeactivationEnabled = false
...
+    echoCancellation: true,
+    autoGainControl: true,
+    noiseSuppression: true,
+    highpassFilter: true,
+    echoCancellationMode: .software,
+    autoGainControlMode: .software,
+    noiseSuppressionMode: .software,
+    highPassFilterMode: .software
```
pblazej: Nit: `highpassFilter` above

pblazej (commented):
LGTM, my biggest concern: https://github.com/livekit/client-sdk-swift/pull/1048#discussion_r3472553143

xianshijing-lk (approved):
lgtm, should we land this and make a new release ?

Sources/LiveKit/Track/Local/LocalAudioTrack.swift
```diff
@@ -90,12 +94,37 @@ public class LocalAudioTrack: Track, LocalTrackProtocol, AudioTrackProtocol, @un
...
+    /// - Throws: ``AudioProcessingOptionsError`` when the options cannot be applied.
+    @discardableResult
+    public func setAudioProcessingOptions(_ options: AudioProcessingOptions) throws -> AudioProcessingOptionsResult {
+        guard let audioTrack = mediaTrack as? LKRTCAudioTrack else {
+            throw AudioProcessingOptionsError(
+                code: .applyFailed,
+                message: "Media track is not an audio track",
+            )
+        }
+        let result = audioTrack.setAudioProcessingOptions(options.toRTCType()).toLKType()
```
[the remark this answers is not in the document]
hiroshihorie: Guarded in webrtc-sdk level.

Sources/LiveKit/Track/Local/LocalAudioTrack.swift
```diff
@@ -66,6 +66,10 @@ public class LocalAudioTrack: Track, LocalTrackProtocol, AudioTrackProtocol, @un
             "googNoiseSuppression": options.noiseSuppression.toString(),
             "googTypingNoiseDetection": options.typingNoiseDetection.toString(),
             "googHighpassFilter": options.highpassFilter.toString(),
+            "echoCancellationMode": options.echoCancellationMode.toConstraintValue(),
+            "autoGainControlMode": options.autoGainControlMode.toConstraintValue(),
+            "noiseSuppressionMode": options.noiseSuppressionMode.toConstraintValue(),
+            "highPassFilterMode": options.highpassFilterMode.toConstraintValue(),
```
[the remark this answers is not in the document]
hiroshihorie: "highPassFilterMode" is correct in the WebRTC-SDK

PR #1050, opened 2026-06-29
fix: improve initial video quality by setting x-google-start-bitrate for all video codecs

xianshijing-lk:
## Summary
- Add `x-google-start-bitrate` SDP munging for all video codecs (VP8, VP9, AV1, H264, H265)
- Use 90% of target bitrate as start bitrate to prevent initial blurriness
- Default `degradationPreference` to `.maintainResolution` for video tracks

## Problem
Video starts blurry for 5-15 seconds before improving. This is caused by WebRTC's bandwidth estimator starting at ~300kbps and slowly ramping up to the target bitrate.

## Solution
1. **x-google-start-bitrate**: Tell WebRTC to start at 90% of target bitrate instead of ramping up from ~300kbps. Applied consistently to all video codecs.
2. **MaintainResolution**: When bandwidth is constrained, prefer dropping frames over reducing resolution. This maintains video clarity at the cost of smoothness.

## Changes
- Added `TrackBitrateInfo` struct and tracking in `Transport`
- Added `setTrackBitrateInfo()` method to register bitrate info when publishing video tracks
- Added `mungeStartBitrate()` SDP munging function
- Updated negotiate sequence to apply start bitrate munging
- Changed default `degradationPreference` from `.auto` to `.maintainResolution`

## Test plan
- [ ] Verify video quality is sharp from the start when publishing
- [ ] Test with VP8, VP9, H264, AV1 codecs
- [ ] Verify bandwidth estimator adapts properly if network can't handle the start bitrate

### DIFF
--- a/Sources/LiveKit/Core/Transport.swift
+++ b/Sources/LiveKit/Core/Transport.swift
@@ -61,6 +61,14 @@ actor Transport: NSObject, Loggable {
     private var _isRestartingIce: Bool = false
     private var _latestOfferId: UInt32 = 0
 
+    // Track bitrate info for x-google-start-bitrate SDP munging
+    private var _trackBitrates: [String: TrackBitrateInfo] = [:]
+
+    struct TrackBitrateInfo {
+        let codec: String
+        let maxBitrateKbps: Int
+    }
+
     // forbid direct access to PeerConnection
     private let _pc: LKRTCPeerConnection
 
@@ -116,6 +124,11 @@ actor Transport: NSObject, Loggable {
 
     func setIsRestartingIce() {
         _isRestartingIce = true
+    }
+
+    /// Register bitrate info for a track to enable x-google-start-bitrate SDP munging
+    func setTrackBitrateInfo(cid: String, codec: String, maxBitrateKbps: Int) {
+        _trackBitrates[cid] = TrackBitrateInfo(codec: codec, maxBitrateKbps: maxBitrateKbps)
     }
 
     func add(iceCandidate candidate: IceCandidate) async throws {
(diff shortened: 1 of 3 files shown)


Sources/LiveKit/Core/Transport.swift
```diff
@@ -253,6 +275,88 @@ extension Transport {
...
+    /// Why 90%: Gives ~10% headroom for bandwidth estimation while starting close to target.
+    /// Why same for all codecs: Target bitrate already accounts for codec efficiency
+    /// (e.g., users set lower targets for VP9/AV1 knowing they're more efficient).
+    private static let startBitrateMultiplier = 0.9
+
+    private static let videoCodecs = ["vp8", "vp9", "av1", "h264", "h265"]
+
+    /// Munge SDP to add x-google-start-bitrate for video codecs.
+    /// This helps prevent initial video blurriness by starting at a higher bitrate.
+    func mungeStartBitrate(_ sdp: String, trackBitrates: [String: TrackBitrateInfo]) -> String {
```
pblazej, line 289: A few suggestions:

- **Per-codec, not global max.** We compute `max(maxBitrateKbps)` across all tracks and stamp that single value onto every video payload. The consensus in the thread was "0.9× of *that codec's* target." It happens to work because one codec is negotiated at a time, but it's not what it says on the tin.
- **H264 loses payloads.** `codecPayloads[codecName] = payload` keeps only the last payload per codec name. H264 commonly advertises several profile payloads; only one gets the hint. (Low impact since H264 doesn't benefit anyway, but it shows the matching is loose.)
- **Can be `static`** like `mungeInactiveToRecvOnlyForMedia` — it uses no instance state.
- **Needs a test.** This is non-trivial string parsing (CRLF detection, fmtp insert-vs-append, payload matching) with no coverage. Please add a unit test that feeds a representative VP9/AV1 offer and asserts the resulting `a=fmtp` line.

pblazej:
Changing the `degradationPreference` default to `.maintainResolution` is the right call — it's the consensus from the investigation thread and the change that actually fixed the reported VP8 case (the `x-google-start-bitrate` munging is confirmed to be a no-op for VP8/H264 and only helps SVC codecs). Please add a changeset entry for it, since it's a public default-behavior change for camera publishing. Also needs a lint/format pass — CI is red.

pblazej:
@xianshijing-lk covered by https://github.com/livekit/client-sdk-swift/pull/1121?

Issue #368, opened 2024-04-22
Reconnecting State not activated while quick reconnects


buttjer:
**Describe the bug**
The room event roomIsReconnecting seems to be called late. The system is already trying to reconnect while the there is no network at all. Also the state remains .connected.

But the event is called if the network is reestablished. Followed immedetiary by roomDidReconnect . 
Is this on purpose? I think it should reflect the actual state. The room connect state seems also remain in .connected. I would expect reconnecting or at least disconnected .

Otherwise, how to show the user that the systems is actively trying to reconnect?
No internet could also mean that the system gave up trying to reconnect at all. I can’t see a way to distinguish that.
I also think in the version 1 of the SDK this also behaved differently. 

**SDK Version**
2.0.5

**iOS/macOS Version**
iOS 17.4.1

**Steps to Reproduce**
Disable network while in a call.

**Expected behavior**
Connection State changes to `reconnecting`.
`roomIsReconnecting` called.

**Logs**
```2024-04-22T15:05:40+0200 warning LiveKitSDK : [LiveKit] Engine.startReconnect(reason:nextReconnectMode:) [Connect] Reconnect already in progress...
2024-04-22T15:05:40+0200 error LiveKitSDK : [LiveKit] Engine.transport(_:didUpdateState:) Failed calling startReconnect, error: Error Domain=io.livekit.swift-sdk Code=104 "Invalid state" UserInfo={NSLocalizedDescription=Invalid state}
2024-04-22T15:05:49+0200 warning LiveKitSDK : [LiveKit] Engine.startReconnect(reason:nextReconnectMode:) [Connect] Reconnect already in progress...
2024-04-22T15:05:49+0200 error LiveKitSDK : [LiveKit] Engine.transport(_:didUpdateState:) Failed calling startReconnect, error: Error Domain=io.livekit.swift-sdk Code=104 "Invalid state" UserInfo={NSLocalizedDescription=Invalid state}
Task <B4FFE66E-82E2-4D5B-9767-7C60D4CD1DA1>.<1> finished with error [-1001] Error Domain=NSURLErrorDomain Code=-1001 "The request timed out." UserInfo={NSErrorFailingURLStringKey=wss://...&protocol=12&sdk=swift&version=2.0.5&os=iOS&os_version=17.4.1&device_model=iPhone16,1&network=wifi&reconnect=1&auto_subscribe=0&adaptive_stream=0, _NSURLErrorRelatedURLSessionTaskErrorKey=(
[...]

Issue #367, opened 2024-04-19
ConnectionState does not match the connection state within LivekitSDK (didUpdateConnectionState not updating)


michaelbutler1998:
**Describe the bug**

connectionState still appears to be .connected (didUpdateConnectionState is not called to alert about .disconnect) when you lose connection for a prolonged period (around 15+ seconds), but the LiveKitSDK can see that you should be in .disconnected

**SDK Version**
2.0.5

**iOS Version**
17.4.1

**Steps to Reproduce**

1. Join a call, with another participant active and listening
2. Turn off your cellular network for 15 seconds.

Whilst you're in this state you can see livekit logs (see log #1 below). When checking connectionState from room, connectionState == .connected.

didUpdateConnectionState is not called to notify that state is now .disconnected

1. Quickly, toggle mute / unmute when cellular is back on

You now end up in a situation where the listening user cannot hear or communicate with you. If the listening user, or the glitched user rejoins, you can hear each other again.

**Expected behavior**

1. During the reconnect, the state should be .reconnecting
2. connectionState is updated when Livekit is aware of .disconnection, so that we can prevent users from turning their mic on/off when not connected

**Logs**
Please provide logs if you can.

```
Log #1 - \`\`\`
2024-04-19T14:23:14+0100 error LiveKitSDK : \[LiveKit\] SignalClient.*sendRequest(*:) connectionState is .disconnected
2024-04-19T14:23:14+0100 error LiveKitSDK : \[LiveKit\] Engine.transport(\_:didGenerateIceCandidate:) Failed to send iceCandidate, error: Error Domain=io.livekit.swift-sdk Code=104 "Invalid state(connectionState is .disconnected)" UserInfo={NSLocalizedDescription=Invalid state(connectionState is .disconnected)}
```

michaelbutler1998:
Possibly related bug:

1. Connect to call with two phones, cellular only on 4G/5G.
2. With one phone, in your phone's settings, set your connection to 2G
3. Whilst in 2G, toggle your microphone
4. Change your phones settings back to 4G
5. Attempt to communicate / toggle mic etc.

You cannot be heard by the other participant, the same as in the initial bug report.

If you are not using Callkit, you get this error:
[...]

PR #1052, opened 2026-06-30
Fix ping/pong timeout never firing: a lost signaling connection stays `.connected` forever

silviudeac:
## Summary
  
When the signaling WebSocket dies but media stays up, the SDK gets stuck in a zombie state: `Room.connectionState` stays `.connected`, `didUpdateConnectionState` never fires, and the client neither reconnects nor disconnects. The ping/pong liveness check that should catch this never fires, because its timeout is re-armed on every ping.

This has happened in practice to us when the device switches networks silently in the background.
  
Related: #367, #368.

 ## Root cause

  `SignalClient._onPingIntervalTimer()` arms the ping timeout with `_pingTimeoutTimer.restart()` on **every** ping interval:

  ```swift
  func _onPingIntervalTimer() async throws {
      guard let jr = _state.lastJoinResponse else { return }
      try await _sendPing()
      _pingTimeoutTimer.setTimerInterval(TimeInterval(jr.pingTimeout))
      _pingTimeoutTimer.setTimerBlock { /* cleanUp(.serverPingTimedOut) */ }
      _pingTimeoutTimer.restart()   // cancels the in-flight countdown and starts a new one
  }
  ```

  `restart()` cancels the running countdown and starts a fresh one. With the server's typical `pingInterval` (5s) `< pingTimeout` (15s), every ping pushes the
  deadline out before it can elapse, so it **never** fires:

  ```
  t=0   ping -> arm timeout, deadline t+15
  t=5   ping -> restart,      deadline t+20   (t+15 cancelled)
  t=10  ping -> restart,      deadline t+25   (t+20 cancelled)
  ...   never fires
  ```
  
The timeout is meant to be cleared by a **pong** (`_onReceivedPong` / `_onReceivedPongResp` -> `cancel()`), not by sending another ping. Healthy connections hide the bug because a pong cancels the timer before the next ping touches it, so `restart()` and a non-resetting start are equivalent there — they only diverge once pongs stop, which is exactly the case the timeout exists to catch. (Ping sends failing doesn't help either: the send error is swallowed in the
  request queue, so `_onPingIntervalTimer` runs to completion and re-arms every interval regardless.)                                       

## Why nothing else catches it

The transport/ICE-failure path (`Room+TransportDelegate` -> `startReconnect(reason: .transport)`) only fires when the **media** peer connection dies. A signaling-only outage — signaling dead, media alive — leaves that path silent, so the ping/pong timeout is the *only* watchdog for it.
  
## Fix

Add `AsyncTimer.startIfStopped()`, which starts the timer only if it isn't already running and otherwise leaves the in-flight countdown untouched, and use it to arm the ping timeout.

 The first un-answered ping arms the deadline and subsequent pings leave it intact; when pongs stop, it elapses once -> `cleanUp(.serverPingTimedOut)` -> `.disconnected` -> the existing reconnect path runs.

## Tests
  
Added `AsyncTimerTests`:
  - `startIfStoppedFiresWhileRepeatedlyArmed` — arms the timer far more often than its interval; asserts the block still fires (fix behavior).
  - `restartNeverFiresWhileRepeatedlyArmed` — same with `restart()`; asserts it never fires while being re-armed (regression guard documenting the bug).

### DIFF
--- a/.changes/fix-ping-timeout
+++ b/.changes/fix-ping-timeout
@@ -1,0 +1,1 @@
+patch type="fixed" "Ping/pong timeout was re-armed on every ping and never fired, leaving a lost signaling connection stuck at .connected"
--- a/Sources/LiveKit/Core/SignalClient.swift
+++ b/Sources/LiveKit/Core/SignalClient.swift
@@ -677,7 +677,8 @@ private extension SignalClient {
             await cleanUp(withError: LiveKitError(.serverPingTimedOut))
         }
 
-        _pingTimeoutTimer.restart()
+        // Arm without resetting a running countdown, else every ping pushes the deadline out and it never fires.
+        _pingTimeoutTimer.startIfStopped()
     }
 
     func _onReceivedPong(_: Int64) async {
--- a/Sources/LiveKit/Support/Async/AsyncTimer.swift
+++ b/Sources/LiveKit/Support/Async/AsyncTimer.swift
@@ -89,4 +89,10 @@ final class AsyncTimer: Sendable, Loggable {
 
         Task { await scheduleNextInvocation() }
     }
+
+    /// Starts the timer only if not already running, leaving an in-flight countdown untouched.
+    func startIfStopped() {
+        if _state.read({ $0.isStarted }) { return }
+        restart()
+    }
 }
--- a/Tests/LiveKitCoreTests/AsyncTimerTests.swift
+++ b/Tests/LiveKitCoreTests/AsyncTimerTests.swift
@@ -23,7 +23,6 @@ import LiveKitTestSupport
 
 @Suite(.tags(.concurrency))
 struct AsyncTimerTests {
-
     @Test func startIfStoppedFiresWhileRepeatedlyArmed() async throws {
         let counter = ConcurrentCounter()
         let timer = AsyncTimer(interval: 0.2)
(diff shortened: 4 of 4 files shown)


pblazej:
LGTM — the fix is correct, and it matches what [client-sdk-android already does](https://github.com/livekit/client-sdk-android/blob/c8637747ef2d74c1a2a65cbbbd340d5ffd9b0947/livekit-android-sdk/src/main/java/io/livekit/android/room/SignalClient.kt#L865-L874): `startPingTimeout` arms the timeout only if one isn't already running (`if (pongJob != null) return`), which is exactly the semantics `startIfStopped()` adds here. (JS/Rust/Flutter sidestep the same trap by never re-arming the timeout from the ping-send path at all.)

Two small things before merge:

- [ ] **Fix lint** — `swiftformat --lint` fails on `Tests/LiveKitCoreTests/AsyncTimerTests.swift:25` (`blankLinesAtStartOfScope`); drop the blank line right after `struct AsyncTimerTests {`.
- [ ] **Add a `.changes` file** so this lands in the CHANGELOG. e.g. `.changes/fix-ping-timeout`:
  ```
  patch type="fixed" "Ping/pong timeout never fired, leaving a lost signaling connection stuck at .connected"
  ```

Thanks!

pblazej (approved):
Approving — lint and changeset both addressed. Fix is correct, minimal, and matches the existing client-sdk-android behavior. 🚢

PR #1065, opened 2026-07-26
Remove non-public buttonPressed: selector from broadcast picker activation

### LINKED ISSUE #1064: App Store submission blocked by RPSystemBroadcastPickerView.buttonPressed: private selector
**Describe the bug**

Apple's automated App Review blocks iOS submissions that include LiveKit because the final application binary references the non-public selector:

```text
RPSystemBroadcastPickerView.buttonPressed:
```

The rejection says:

> The app uses or references the following non-public or deprecated APIs:
> Selectors:
> - RPSystemBroadcastPickerView.buttonPressed:
>
> Revise the binary and remove all references to any non-public or deprecated APIs.

The downstream app does not use LiveKit screen sharing or ReplayKit. Merely linking LiveKit causes these public SDK source paths to be compiled into the application binary:

- `Sources/LiveKit/Broadcast/BroadcastManager.swift:125`
- `Sources/LiveKit/Track/Support/Extensions.swift:114`

Both construct and perform the private selector:

```swift
let selector = NSSelectorFromString("buttonPressed:")
if view.responds(to: selector) {
    view.perform(selector, with: nil)
}
```

`RPSystemBroadcastPickerView` is public, but programmatically invoking its private `buttonPressed:` selector is not. Deprecating the public `show(...)` extension in #586 did not remove the selector: `BroadcastManager.requestActivation()` uses the same private implementation.

This is especially problematic for apps that do not use screen sharing because there is currently no package product or compile-time option that excludes this code from the linked binary.

**SDK Version**

- Rejected binary: `client-sdk-swift` 2.13.0
- Confirmed still present in current release: 2.15.2

**iOS/macOS Version**

- iOS App Store submission; rejection occurs during Apple's automated static review rather than at runtime
- Build host: macOS 26.4

**Xcode Version**

- Xcode 26.6 (17F113)

**Steps to Reproduce**

1. Add `client-sdk-swift` to an iOS application through Swift Package Manager.
2. Link the `LiveKit` product. The app does not need to call `setScreenShare`, `BroadcastManager`, or any ReplayKit API.
3. Create a Release archive and submit it to App Store Connect.
4. Apple's automated review blocks processing and identifies `RPSystemBroadcastPickerView.buttonPressed:` as a non-public/deprecated API reference.

The source reference can also be confirmed directly in 2.15.2:

```sh
rg 'buttonPressed:' Sources/LiveKit
```

**Examples**

The problem is reproducible with:

- [ ] [client-example-swift](https://github.com/livekit/client-example-swift)
- [ ] [client-example-collection-swift](https://github.com/livekit/client-example-collection-swift)

We reproduced it in a production downstream application archive.

**Expected behavior**

The LiveKit package should not place references to non-public Apple selectors in downstream application binaries.

Possible compliant directions include:

1. Replace programmatic activation with a public API design that exposes/configures an `RPSystemBroadcastPickerView` for the application to place in its visible UI and for the user to tap.
2. Separate broadcast-screen-sharing support into an opt-in package product so apps that do not use it do not link the ReplayKit activation implementation.
3. At minimum, add a compile-time option that completely excludes both `buttonPressed:` paths.

Could this be treated as a release-blocking issue? Updating from 2.13.0 to 2.15.2 does not resolve the App Store rejection.

**Screenshots**

Apple's rejection text is included above. A screenshot can be provided if needed.

**Logs**

No runtime logs apply; this is an App Store static-analysis rejection.

hiroshihorie:
## Summary

Apple's automated App Review now flags the literal `buttonPressed:` selector string in application binaries (ITMS-90338 / Guideline 2.5.1) and blocks submissions of any app that links LiveKit — even apps that never use screen sharing or ReplayKit (#1064).

Two code paths embedded the private selector:

- `BroadcastManager.requestActivation()` (`Sources/LiveKit/Broadcast/BroadcastManager.swift`) — the current activation path, invoked automatically from `LocalParticipant.setScreenShare(enabled:)` when `useBroadcastExtension` is enabled
- The deprecated `RPSystemBroadcastPickerView.show(for:showsMicrophoneButton:)` extension (`Sources/LiveKit/Track/Support/Extensions.swift`) — deprecated in #586, but the selector still shipped in the binary

Both now activate the picker using only public API: locate the picker's `UIButton` subview and fire it via `sendActions(for: .touchUpInside)`. This is the same approach used by other RTC SDKs (Tencent TUIRoomKit, Nextcloud Talk, Agora examples, 100ms). If the button is ever not found (e.g. a future OS changes the view hierarchy), a warning is logged instead of silently no-opping.

No public API changes; behavior is identical at runtime.

## Verification

- `rg buttonPressed Sources/` returns no matches, and `strings` over the compiled objects and a device build of the example app confirms the selector string no longer appears in downstream binaries
- Tested on a physical iPhone with the swift-example app: the system broadcast picker opens via `requestActivation()`, shows the app's broadcast extension preselected, and the broadcast starts and publishes normally
- `xcodebuild build -scheme LiveKit -destination 'generic/platform=iOS Simulator'` succeeds; swiftformat / swiftlint clean

Fixes #1064

### DIFF
--- a/.changes/broadcast-picker-private-selector
+++ b/.changes/broadcast-picker-private-selector
@@ -1,0 +1,1 @@
+patch type="fixed" "Remove non-public buttonPressed: selector from broadcast picker activation"
--- a/Sources/LiveKit/Broadcast/BroadcastManager.swift
+++ b/Sources/LiveKit/Broadcast/BroadcastManager.swift
@@ -23,6 +23,8 @@ import Foundation
 import ReplayKit
 #endif
 
+import UIKit
+
 /// Manages the broadcast state and track publication for screen sharing on iOS.
 public final class BroadcastManager: Sendable {
     /// Shared broadcast manager instance.
@@ -115,16 +117,22 @@ public protocol BroadcastManagerDelegate: Sendable {
     func broadcastManager(didChangeState isBroadcasting: Bool)
 }
 
-private extension RPSystemBroadcastPickerView {
+extension RPSystemBroadcastPickerView {
     /// Convenience function to show broadcast picker.
     static func showPicker(for preferredExtension: String?) {
         let view = RPSystemBroadcastPickerView()
         view.preferredExtension = preferredExtension
         view.showsMicrophoneButton = false
+        view.triggerPicker()
+    }
 
-        let selector = NSSelectorFromString("buttonPressed:")
-        guard view.responds(to: selector) else { return }
-        view.perform(selector, with: nil)
+    /// Simulates a tap on the picker's button using only public API.
+    func triggerPicker() {
+        guard let button = subviews.compactMap({ $0 as? UIButton }).first else {
+            sharedLogger.log("Unable to find button in RPSystemBroadcastPickerView", .warning, type: RPSystemBroadcastPickerView.self)
+            return
+        }
+        button.sendActions(for: .touchUpInside)
     }
 }
 
--- a/Sources/LiveKit/Track/Support/Extensions.swift
+++ b/Sources/LiveKit/Track/Support/Extensions.swift
@@ -111,10 +111,7 @@ public extension RPSystemBroadcastPickerView {
         let view = RPSystemBroadcastPickerView()
         view.preferredExtension = preferredExtension
         view.showsMicrophoneButton = showsMicrophoneButton
-        let selector = NSSelectorFromString("buttonPressed:")
-        if view.responds(to: selector) {
-            view.perform(selector, with: nil)
-        }
+        view.triggerPicker()
     }
 }
 #endif


hiroshihorie:
## Follow-up: cleaner path on iOS 27+

This PR keeps the tap-simulation approach because it's the only way to programmatically launch the picker on iOS ≤ 26 — Apple never exposed a public trigger for `RPSystemBroadcastPickerView`. But as of iOS 27, there finally is an official route:

- The entire ReplayKit broadcast stack is deprecated at 27.0: `RPSystemBroadcastPickerView` ("Use ScreenCaptureKit SCContentSharingPicker instead"), and notably `RPBroadcastSampleHandler` is marked "No longer supported".
- ScreenCaptureKit arrives on iOS/iPadOS/tvOS/visionOS in 27.0, and [`SCContentSharingPicker`](https://developer.apple.com/documentation/screencapturekit/sccontentsharingpicker) provides a fully public, programmatic `present()` — exactly what this workaround has been emulating.

Suggested direction:

1. **Short term (this PR):** ship the `sendActions` activation — required for every device on iOS ≤ 26 for years to come.
2. **Next:** add an `SCContentSharingPicker`-based activation path behind `if #available(iOS 27, *)` so newer devices use the sanctioned API and drop the view-hierarchy assumption entirely.
3. **Larger effort (tracking issue):** the `RPBroadcastSampleHandler` deprecation means the broadcast-upload-extension architecture itself (`BroadcastScreenCapturer`, IPC channel, sample handler) has a designated successor on iOS 27+ — a ScreenCaptureKit capture path could eventually remove the extension requirement and its app-group/IPC setup for apps targeting 27+.

Note: using deprecated *public* API doesn't trigger App Review rejections — only the non-public selector did — so there's no submission risk in keeping ReplayKit support for older OS versions.

xianshijing-lk (approved):

PR #1070, opened 2026-07-30
feat(e2ee): add BaseKeyProvider.setKey(keyData:) for raw binary keys

eleboucher:
Accept arbitrary key bytes instead of a UTF-8 String so binary key material from external E2EE layers can be installed without string conversion. The String variant now delegates to it.


pblazej:
LGTM, can you add a small note to the `.changes` (patch)?

pblazej:
Test flakiness addressed elsewhere.

pblazej (approved):

PR #1071, opened 2026-07-30
Fix CI flakiness: data-stream races, QoS inversions, deterministic tests

pblazej:
Deep investigation of CI flakiness (61 failed runs sampled across Jun–Jul, plus per-run test/server log capture). Every fix below is root-caused from CI evidence; several turned out to be production bugs rather than test issues.

## Production code changes

**`IncomingStreamManager` — dropped and reordered incoming data streams**
Senders reuse a stream ID for consecutive streams (transcription does); descriptor cleanup ran in a per-reader task that raced the reopening header and silently dropped the new stream (`confirmed 2 of 5` in CI). Cleanup is now synchronous in the trailer and chunk-error paths and generation-guarded so a stale cleanup can never remove a successor; open streams are failed on sender disconnect and room cleanup. Cross-stream handler ordering was also a scheduler race (one detached task per stream); topics registered with `ordered: true` now order handlers by wire happens-before — a handler waits for handlers of streams that closed before its own opened, while concurrently open streams run concurrently, so a long-lived stream never delays overlapping ones (e.g. a user's transcript during barge-in). Opt-in, used by the transcription receiver.

**`AsyncCompleter` — timeout timers could be deferred indefinitely**
Timeout blocks ran on a `.background` queue, the QoS tier the system may defer without bound under CPU pressure. A waiter whose only resume path is the timeout then never wakes: two CI legs burned full 30-minute attempts hung this way, and RPC callers saw `1501` instead of `1502` when the 0.05 s timer lost to a 2 s watchdog. Now `.utility`.

**`DeviceManager` — same QoS inversion on camera start**
Device-discovery KVO registration ran on the `.background` global queue and is the only resumer of the completers `CameraCapturer.startCapture` awaits. Now `.utility`, matching the completer timers.

**`AsyncTimer` — injectable sleep**
Internal `sleep:` parameter (defaulted, all call sites unchanged) so timing tests drive ticks deterministically instead of asserting against the scheduler.

**SwiftLint: `no_background_qos`** — errors on any `.background` queue QoS or task priority to keep the inversion class out.

## Test changes

**`withRooms` teardown on every exit path** — a thrown body previously skipped `disconnect()`, leaking live Rooms (WebSocket, peer connections, timers) into every later suite in the process; this was the amplifier that turned one flake into a run-wide cascade.

**Deterministic waits instead of fixed sleeps** — AsyncTimer tests drive ticks via the injected sleep (`ManualSleeper`); transcription tests wait for expected update counts; reliable-data tests wait for delivery *inside* `withRooms` (waiting after teardown lost in-flight packets: 56–127 of 128 delivered).

**RPC tests inject via the awaited `afterPublish` hook** — unstructured `Task {}` injection raced response timeouts under load (`1502` flakes).

**ObjC tests retry connects** (like `withRooms` always has) — the first `Room` in a fresh process pays one-time WebRTC/audio init that can exceed the connect timeout on simulators (`testSendFile`, 40 historical failures).

**Buffer-track tests resolve dimensions directly** — captured frames resolve dimensions on the capture queue, which lags past the publish timeout under load.

**`WebSocketTests` disconnect cancelled connects** — a cancelled connect can still complete the server-side join; 21 leaked participants per run otherwise.

**New regression tests** — stream-ID reuse delivery and ordered-topic FIFO (unit-level, drive the manager's event loop directly; the reuse test reproduces the drop deterministically without the fix), transcription burst-ordering, AsyncTimer semantics.

**Guard against process aborts** — transcription validation indexed past `#expect` count mismatches, crashing the whole test process (52 aborts historically).

## CI changes

- **Job-level retries replaced with in-run test retries**: the `nick-fields/retry` wrapper masked every flake above behind full rebuild-and-rerun attempts, with failures buried in superseded logs. Now `-retry-tests-on-failure -test-iterations 2` re-runs failures inside the same xcodebuild invocation (XCTest re-runs only the failed cases; Swift Testing runs a second iteration) — a flaky leg costs one extra test pass, and per-iteration results stay visible in the log and xcresult.
- **Failure artifacts**: raw `xcodebuild` output (xcbeautify swallows parameterized-test failure details) and the dev-server log (close reasons prove client vs server fault) upload on failure.
- Xcode 26.6 on macos-26 legs; new `xcode-27` preview legs (macOS, Catalyst, iOS/tvOS/visionOS 27.0 sims).

### DIFF
--- a/.changes/fix-background-qos-inversions
+++ b/.changes/fix-background-qos-inversions
@@ -1,0 +1,1 @@
+patch type="fixed" "Completer timeouts and camera device discovery no longer run at background QoS, which the system may defer indefinitely under load"
--- a/.changes/fix-data-stream-ordering
+++ b/.changes/fix-data-stream-ordering
@@ -1,0 +1,1 @@
+patch type="fixed" "Incoming data streams could be dropped on stream-ID reuse or handled out of order; transcription handlers now run in wire order and open streams are failed on sender disconnect and room cleanup"
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -87,6 +87,7 @@ jobs:
       - uses: actions/checkout@v7
 
       - name: Run LiveKit Server
+        id: livekit-server
         uses: livekit/dev-server-action@61e2b4dcb170dd3591e0c9b0db3c3fe5db93b500
         with:
           github-token: ${{ github.token }}
@@ -130,6 +131,17 @@ jobs:
         with:
           name: test-log-${{ strategy.job-index }}
           path: ${{ runner.temp }}/xcodebuild.log
+          retention-days: 5
+
+      # Client-side logs can't distinguish "we gave up" from "the SFU hung up on
+      # us": a failed e2e test just shows `Connection reset by peer`. The server's
+      # JSONL log records the participant close reason, so capture both sides.
+      - name: Upload Server Log
+        if: ${{ failure() }}
+        uses: actions/upload-artifact@v7
+        with:
+          name: server-log-${{ strategy.job-index }}
+          path: ${{ steps.livekit-server.outputs.log-path }}
           retention-days: 5
 
       - name: Build for Release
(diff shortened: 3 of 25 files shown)


hiroshihorie (approved):

pblazej:
Most of the systematic failures that can be mapped to either client/server logs should be fixed now, the rest is timeouts/runner congestion without specific case.

xianshijing-lk (approved):

PR #1072, opened 2026-07-30
Fix quiet remote audio with software voice processing on iOS

hiroshihorie:
## Problem

With software voice processing (`echoCancellationMode: .software` etc.) and the microphone published, remote audio plays out noticeably quieter than with Apple voice processing. Reported by a customer and reproduced on device.

## Root cause

iOS applies a reduced, call tuned speaker gain when the audio session mode is `.videoChat` or `.voiceChat` while capture is active. Apple Voice Processing I/O adds its own loudness stage on the playout path that compensates for this. WebRTC software processing has no VPIO, so the SDK's chat mode session config exposes the raw call tuned gain.

Verified with a pure AVFoundation reproduction app (no WebRTC): with capture active and VPIO off, flipping only the session mode between `.videoChat` and `.default` toggles playout between quiet and loud. Everything else was ruled out on device: category options (mixWithOthers, Bluetooth options), mono output connects, the full ADM graph shape, VPIO teardown residue, and the WebRTC render path (bit identical between modes).

## Change

`AudioSessionEngineObserver` now selects the session config from the voice processing implementation:

- Apple voice processing active: `.playAndRecordSpeaker` / `.playAndRecordReceiver` (chat modes, unchanged)
- Software processing: new `.playAndRecordSpeakerMedia` / `.playAndRecordReceiverMedia` presets using `.default` mode (plus `.defaultToSpeaker` for the speaker preference)

The ADM commits its engine state only at the end of a transition, but the session is configured during it (`engineWillEnable`), so the observer cannot read the resolved state back. Instead, `AudioManager.startLocalRecording`, `setRecordingAlwaysPreparedMode`, and `LocalAudioTrack.setAudioProcessingOptions` inform the observer of the expected implementation before the transition starts, via the new internal `AudioProcessingOptions.requestsPlatformEchoNoisePath`.

The behavior follows the voice processing implementation unconditionally, no new public API. Apps needing the legacy chat mode config can disable automatic session configuration and manage the session directly.

## Companion example

livekit-examples/agent-starter-swift#48 adds an audio options panel to the agent example with a voice processing mode picker (automatic, platform, software), which is the UI used to validate and demo this fix on device.

## Validated on device

iPhone 17 Pro Max, agent-starter-swift against this branch: software mode with mic published is now loud, platform and automatic modes unchanged. Platform VP state confirmed off via `platformVoiceProcessingState` diagnostics during the test.

## Simplification once webrtc-sdk/webrtc#275 lands

The expectation push (`updateExpectedPlatformVoiceProcessing`, `setPlatformVoiceProcessingExpected`, `requestsPlatformEchoNoisePath`, and the `LocalAudioTrack` call site) exists only because the resolved voice processing state is not readable when `engineWillEnable` fires. webrtc-sdk/webrtc#275 passes that state through the `willEnableEngine` delegate callback. Once it ships in an xcframework release, all of the above gets deleted and the observer reads the value directly from the callback, with no Swift side resolution logic. This PR intentionally keeps the mechanism minimal and internal so that swap stays small.

## Known limitations / follow ups

- Separate bug found during investigation: runtime `setAudioProcessingOptions` switching platform to software on a published track does not tear down VPIO until the mic is republished (the reverse direction works). Tracked separately.
- Louder playout gives software AEC more echo to cancel. Echo behavior should be evaluated before enabling this by default in a release; the opt out exists for that reason.

### DIFF
--- a/.changes/software-vp-session-mode
+++ b/.changes/software-vp-session-mode
@@ -1,0 +1,1 @@
+patch type="fixed" "Use media session mode when recording with software voice processing to fix quiet remote audio"
--- a/Sources/LiveKit/Audio/AudioSessionEngineObserver.swift
+++ b/Sources/LiveKit/Audio/AudioSessionEngineObserver.swift
@@ -64,12 +64,39 @@ public class AudioSessionEngineObserver: AudioEngineObserver, Loggable, @uncheck
         set { _state.mutate { $0.isSpeakerOutputPreferred = newValue } }
     }
 
+    /// Controls whether the session mode follows the voice processing implementation.
+    ///
+    /// iOS applies a reduced, call-tuned speaker gain when the session mode is
+    /// `.videoChat` or `.voiceChat` while the microphone is active. Apple's
+    /// Voice Processing I/O compensates with its own loudness stage, WebRTC
+    /// software processing does not, which makes remote audio noticeably
+    /// quieter in software mode.
+    ///
+    /// - When `true`: `.default` mode is used while recording with software
+    ///   voice processing, keeping media playback loudness. Chat modes are
+    ///   used when Apple voice processing is active.
+    /// - When `false`: Chat modes are always used while recording (legacy behavior).
+    ///
+    /// > Note: This value is only used when `isAutomaticConfigurationEnabled` is `true`.
+    ///
+    /// Default value: `true`
+    public var isSoftwareProcessingMediaModeEnabled: Bool {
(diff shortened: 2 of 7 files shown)


xianshijing-lk (approved):

pblazej:
Tested against agent example PR and I think it works smoothly (as far as the volume matches on both settings).

pblazej (approved):

PR #1073, opened 2026-07-30
Add WsClient and HttpClient implementations for livekit-net.

jhugman:
Initialize `livekit-net` with a Swift implementation of the `WsClient` and `HttpClient`. Production versions of these use `URLSession`.

This requires https://github.com/livekit/rust-sdks/pull/1290 to be built and released.

For testing, there is a self-test which calls through Rust into dummy implementations of `WsClient` and `HttpClient`.

### DIFF
--- a/Sources/LiveKit/Core/Room.swift
+++ b/Sources/LiveKit/Core/Room.swift
@@ -257,6 +257,7 @@ public class Room: NSObject, @unchecked Sendable, ObservableObject, Loggable {
                 connectOptions: ConnectOptions? = nil,
                 roomOptions: RoomOptions? = nil)
     {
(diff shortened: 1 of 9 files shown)


pblazej:
Nice, but how it's consumed on the Rust side now? It's not?

pblazej:
@jhugman I'll try to incorporate that soon into the telemetry crate, then maybe add some small tweaks to the swift side 👍

pblazej:
Decided to wrap the existing clients instead: https://github.com/livekit/client-sdk-swift/pull/1106

so they get some historical workarounds/some degree of "safety".

Regardless of the impl, it works pretty well 👏

PR #1078, opened 2026-08-03
Encapsulate SDP munging in a tested, line-preserving SDP module

pblazej:
Adds an internal `SDP` module (`Sources/LiveKit/Support/SDP/`) — a lossless, line-oriented SDP document model with typed accessors for only the attributes we munge (mid, direction, rtpmap, fmtp). Every other line passes through verbatim, so `parse ∘ write` is the identity for any input. The existing munging (`mungeInactiveToRecvOnlyForMedia`) is migrated onto it; PR #1068 and the start-bitrate work can follow.

Also adds a munge fallback: libwebrtc m144 actively validates munged SDP and can reject it outright (`IsSdpMungingAllowed`, expanding via field-trial kill switches; our munges are tracked upstream as `kDirection` / `kAudioCodecsFmtpOpusStereo`). Rejection happens before apply and leaves the peer connection untouched, so `Transport` now retries with the unmunged offer instead of failing negotiation — proven against a real peer connection in `TransportMungeFallbackTests`.

Tests: fixtures ported from sdp-transform's suite (MIT) plus crafted ones (data channel, unknown lines, malformed lines); 28 tests, 100% line coverage of `SDP.swift`.

Resolves CLT-3189.

### Differences from client-sdk-js (sdp-transform) behavior — all deliberate
- Exact-token fmtp parameter matching: JS uses a substring check, so an existing `sprop-stereo=1` defeats adding `stereo=1`; we match exact parameters.
- Uniformly case-insensitive codec lookup (RFC 4855): JS is case-insensitive for video codecs but exact-match for opus.
- `mid` is always a `String`: sdp-transform parses numeric mids as numbers (the reason for JS's `getMidString` workaround).
- Unknown lines are preserved verbatim and in place: sdp-transform drops unknown line types (e.g. `y=`) and re-emits unknown `a=` lines after the recognized ones.
- Round-trip is byte-exact including EOL style (`\r\n`/`\n`/`\r`): sdp-transform rewrites lines in canonical order.

### Parser alternatives considered
- **Parser generated from the spec (ANTLR etc.)** — SDP is line-oriented with attribute-value grammars scattered across many RFCs; generated parsers are strict where munging must be lenient, and the runtime is a new dependency.
- **Existing Swift SDP libraries** — nothing adoptable: abandoned, incomplete, or unlicensed.
- **Rust crate via LiveKitUniFFI** — best candidate measured ~286 KB compiled (over the size budget), adds release coupling with rust-sdks, and marshaling a mutable document model across FFI is clumsy for ops this small.
- **Full Swift port of sdp-transform** (grammar table + writer + fixture suite) — viable, but 51 of its 64 grammar rules cover attributes we never touch, and its writer reorders output. This remains the escalation path if our typed accessors grow past ~15 line kinds; the fixture suite carries over unchanged.

### DIFF
--- a/.changes/sdp-munging-module
+++ b/.changes/sdp-munging-module
@@ -1,0 +1,1 @@
+patch type="changed" "Encapsulate SDP munging in a tested SDP module and fall back to the unmunged offer if libwebrtc rejects munged SDP"
--- a/Sources/LiveKit/Core/Transport.swift
+++ b/Sources/LiveKit/Core/Transport.swift
@@ -190,13 +190,13 @@ actor Transport: NSObject, Loggable {
         func _negotiateSequence() async throws {
             _latestOfferId += 1
             var offer = try await createOffer(for: constraints)
-            if singlePCMode {
-                let mungedSDP = Self.mungeInactiveToRecvOnlyForMedia(offer.sdp)
-                if mungedSDP != offer.sdp {
-                    offer = RTC.createSessionDescription(type: offer.type, sdp: mungedSDP)
-                }
+            let mungedSDP = singlePCMode ? Self.mungeInactiveToRecvOnlyForMedia(offer.sdp) : offer.sdp
+            if mungedSDP != offer.sdp {
+                offer = try await set(mungedLocalDescription: RTC.createSessionDescription(type: offer.type, sdp: mungedSDP),
+                                      fallingBackTo: offer)
+            } else {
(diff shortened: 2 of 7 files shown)


pblazej:
@lukasIO tagging you just in case of interesting diffs with JS.

hiroshihorie (approved):

PR #1082, opened 2026-08-04
Add support for a custom video encoder factory

sergeyphi:
This PR adds `LiveKitSDK.set(videoEncoderFactory:)`, a static API that lets apps supply a custom `LKRTCVideoEncoderFactory` to replace the SDK's default VideoToolbox-backed encoder used when building the underlying WebRTC PeerConnectionFactory. 

This mirrors the equivalent capability already available on the Android SDK via [LiveKitOverrides.videoEncoderFactory](https://github.com/livekit/client-sdk-android/blob/main/livekit-android-sdk/src/main/java/io/livekit/android/LiveKitOverrides.kt#L47-L50).

### DIFF
--- a/Sources/LiveKit/Core/RTC+VideoEncoderFactory.swift
+++ b/Sources/LiveKit/Core/RTC+VideoEncoderFactory.swift
@@ -1,0 +1,37 @@
+/*
+ * Copyright 2026 LiveKit
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// This file intentionally exposes `LiveKitWebRTC` publicly: providing a custom video
+// encoder factory necessarily means accepting an `LKRTCVideoEncoderFactory` from the caller.
+// It is the only public surface that leaks the WebRTC module.
+public import LiveKitWebRTC
+
+public extension LiveKitSDK {
+    /// Sets a custom video encoder factory to be used by the underlying WebRTC
+    /// `PeerConnectionFactory`, replacing the default VideoToolbox-backed factory.
+    ///
+    /// Pass `nil` to restore the default factory.
+    ///
+    /// This must be called **before** the peer connection factory is initialized (i.e. before
+    /// the first `Room` connection / track creation). Calling it afterwards throws, because the
+    /// factory is created once and cached for the lifetime of the process.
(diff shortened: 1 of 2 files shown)


pblazej:
@sergeyphi thank you for opening the PR, could you please describe the actual use case?

Is it forcing a particular (software?) codec? Or completely replacing one?

Our main concern is about exposing WebRTC types directly ofc, the preferable way would be:
- full and transperent wrapper, similar to `AudioCustomProcessingDelegate`
- using ` @_spi(WebRTC)` etc.

The actual shape of this "injection" API is slightly different than Android's, so let's try to minimize the impact on the majority of consumers.

sergeyphi:
@pblazej Thank you for the review 🙏  I completely understand your concern about exposing WebRTC internals.

Our use case: not codec forcing, the track already negotiates H.264. We need to swap the encoder implementation behind it: a modified software libopenH264 `RTCVideoEncoder` in place of the VideoToolbox one. Our receiver is an embedded intercom device whose hardware H.264 decoder is very finicky - basically designed to only decode its own video. It doesn't handle the bitstream iOS's VideoToolbox encoder produces, and the software encoder we control fixes it.

I see @hiroshihorie  already pushed `hiroshi/video-encoder-factory-wrapper`, which looks like exactly the right shape! Happy to close this PR in favor of that one.

—

More on @hiroshihorie's encoder factory wrapper:

I checked that wrapper branch against our encoder and it covers essentially everything: settings, QP thresholds, `CodecSpecificInfo.h264`, `VideoCodecInfo` parameters, and `I420VideoBuffer`'s plane pointers (which is what we hand to openH264). Two gaps:

1. RTP timestamp (blocking). We echo `RTCVideoFrame.timeStamp` back into `RTCEncodedImage.timeStamp`. The public `VideoFrame` only carries `timeStampNs`, and `toLKType()` drops `timeStamp` — but my understanding the two aren't interchangeable. Could `VideoFrame` surface it (and `EncodedVideoFrame.rtpTimestamp`'s doc point at that rather than at `timeStampNs`)?
2. `toI420()` on `VideoBuffer` (also blocking). openH264 needs I420 planes, and frames actually arrive as native NV12 `CVPixelBuffers` — we call `[frame.buffer toI420]` per frame. `supportsNativeHandle = false` doesn't change that (in m144 it only gates software cropping and simulcast pass-through vs Scale(), which returns another native buffer). So `toLKType()` hands us a `CVPixelVideoBuffer`, which exposes only `pixelBuffer`, and `I420VideoBuffer` has no public init — no route to the planes. Could `VideoBuffer` gain a `toI420()`, mirroring `RTCVideoFrameBuffer.toI420`?

Minor: `EncodedVideoFrame` omits ntpTimeMs/flags/encodeStartMs/encodeFinishMs, and VideoEncoderSettings has no public init (useful for tests).

Also one note on the ergonomics: the protocols are Swift-only (Sendable, payload enums, closure typealias), so ObjC/ObjC++ encoders like ours need a Swift shim. Fine by us, just flagging it in case an @objc-friendly variant is worth considering.

Both changes look small, `rtpTimestamp` is a line in `toLKType()`, `toI420()` is a protocol requirement plus a `CVPixelVideoBuffer` implementation — happy to send them as a PR against that branch, and to test the result end-to-end with our openH264 encoder!

pblazej:
@hiroshihorie @sergeyphi I think we're on the same page, can we address the above comments/ergonomics in the other PR? 👍

PR #1084, opened 2026-08-11
Add ExternalAudioSource for publishing app audio as an independent track

hiroshihorie:
Adds `ExternalAudioSource`, a pushable audio source the app feeds PCM into, so app or screen-share audio can be published as an independent track instead of being mixed into the microphone track inside the engine.

Built on webrtc-sdk/webrtc#282, which adds the native `RTCExternalAudioSource` and keeps ADM audio out of externally fed send streams. Draft until a LiveKitWebRTC release containing that lands and the pins here can be bumped. CI is expected to fail against `144.7559.11`, which lacks the new API and the updated ADM delegate signature.

## API

- `ExternalAudioSource`: push `AVAudioPCMBuffer` or `CMSampleBuffer`, foreign formats are converted via a cached `AudioConverter`. Buffered pacing with overflow rejection, `bufferedDurationMs`, `clearBuffer()`.
- `LocalAudioTrack.createTrack(externalSource:)`: publishes as `.screenShareAudio`, and `startCapture()` short-circuits so the ADM recording path never starts.
- `ScreenShareCaptureOptions.appAudioPublishMode`: `.mix` (default, existing behavior) or `.separateTrack`. With `.separateTrack`, screen share publishes app audio as its own track and unpublishing the video takes the audio track with it.
- Capturer routing: `BroadcastScreenCapturer` (iOS) and `MacOSScreenCapturer` feed the source directly. In `.separateTrack` mode a missing sink drops audio with a single warning instead of silently falling back into the mic mix. iOS in-app capture has no audio path, so track prep bails with a warning instead of publishing a permanently silent track.

## Commit guide

- `0babffe6` the feature itself (wrapper, track creation, options, capturer routing, lifecycle, E2E test)
- `d794daa6` unpublish all `.screenShareAudio` publications, not just the first match
- `6518a07b` fail track prep when the capturer has no audio path, tag the E2E test `.e2e`, assert the ADM recording path never started
- `1bad4614` never route separate-track app audio into the mic mixer, log app-audio publish failures and let the share continue
- `8795a447` skip the publish frame-wait gate from #690 for externally fed tracks. The watcher observes the ADM capture path, which external sources bypass by design, so the publish always timed out
- `bc739c86` adapt to the new ADM `willEnableEngine` delegate signature (webrtc-sdk#275). Required by any framework built past `144.7559.11`, rides with the pin bump

## Testing

- `ExternalAudioSourceTests` 4/4 against `livekit-server --dev`, including publish to a remote participant with the ADM recording path asserted never started
- Verified on device (macOS): screen share via `set(source:)` publishes video plus an independent `screen_share_audio` track, remote side gets separate volume control per track, stop unpublishes both
- Known follow-up: webrtc-sdk#265 made the engine's mic permission check passive, and the request side is not implemented in the SDK yet. With TCC undetermined, mic publish fails with no prompt unless the app calls `ensureDeviceAccess` first. Needs a decision whether the SDK requests automatically in the publish path

## Merge gates

- webrtc-sdk/webrtc#282 merged and released as a LiveKitWebRTC version
- Bump `Package.swift`, `Package@swift-6.2.swift`, `LiveKitClient.podspec`, and `Package.resolved` to that version in this PR

### DIFF
--- a/Sources/LiveKit/Audio/AudioDeviceModuleDelegateAdapter.swift
+++ b/Sources/LiveKit/Audio/AudioDeviceModuleDelegateAdapter.swift
@@ -40,7 +40,10 @@ class AudioDeviceModuleDelegateAdapter: NSObject, LKRTCAudioDeviceModuleDelegate
         return entryPoint?.engineDidCreate(engine) ?? 0
     }
 
-    func audioDeviceModule(_: LKRTCAudioDeviceModule, willEnableEngine engine: AVAudioEngine, isPlayoutEnabled: Bool, isRecordingEnabled: Bool) -> Int {
+    func audioDeviceModule(_: LKRTCAudioDeviceModule, willEnableEngine engine: AVAudioEngine, isPlayoutEnabled: Bool, isRecordingEnabled: Bool, isVoiceProcessingEnabled _: Bool) -> Int {
+        // isVoiceProcessingEnabled is new in the ADM delegate (webrtc-sdk
+        // PR 275). Ignored for now, exposing it through AudioEngineObserver
+        // is part of the framework version bump, not this feature.
         guard let audioManager else { return 0 }
         let entryPoint = audioManager.buildEngineObserverChain()
         return entryPoint?.engineWillEnable(engine, isPlayoutEnabled: isPlayoutEnabled, isRecordingEnabled: isRecordingEnabled) ?? 0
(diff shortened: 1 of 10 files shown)


serkan-korkac:
Hi @hiroshihorie , thanks for working on this! We’re waiting for this feature to publish ReplayKit app audio separately from the microphone on iOS. I noticed the SDK is now on WebRTC 150.7871.01. Does it include the required RTCExternalAudioSource support? Are you planning to rebase this PR, and is there a rough ETA? Thanks!

hiroshihorie:
I'm actively working on this one (in WebRTC M150), maybe I can bring it to a mergeable state in about a week.

PR #1086, opened 2026-08-13
Configure the audio session from the resolved voice processing state

hiroshihorie:
`willEnableEngine` now carries `isVoiceProcessingEnabled` (webrtc-sdk/webrtc#275, shipped since `144.7559.12`). The adapter hands it to `AudioSessionEngineObserver` before the observer chain runs, so the session mode is always configured from the state the ADM actually resolved, never a Swift-side guess. This is the simplification planned in #1072.

Deleted with that: `updateExpectedPlatformVoiceProcessing` including its simulator special case and the `requestsPlatformEchoNoisePath && isPlatformVoiceProcessingAllowed` mirror of the ADM's resolution, its call sites in `setRecordingAlwaysPreparedMode`, `startLocalRecording`, and `LocalAudioTrack.setAudioProcessingOptions`, and `AudioProcessingOptions.requestsPlatformEchoNoisePath` with its tests. Net -80 lines.

Kept: the direct push in `setPlatformVoiceProcessingAllowed(false)`, since VPIO teardown there happens without an engine transition, so no callback would fire to update the session. The observer state is renamed from `isPlatformVoiceProcessingExpected` to `isPlatformVoiceProcessingActive` to match the resolved semantics.

No public API changes. Passing the state through the public `AudioEngineObserver` chain was deliberately deferred: adding a parameter to `engineWillEnable` breaks existing conformers, and a defaulted overload silently drops the value whenever a chain contains an old-style observer. A context-struct redesign is the right public evolution if consumers need the state.

### DIFF
--- a/.changes/data-tracks
+++ b/.changes/data-tracks
@@ -1,0 +1,1 @@
+minor type="added" "Data tracks: publish and subscribe to ordered frame streams over a dedicated unreliable channel, with optional schema and frame-encoding metadata"
--- a/.changes/vp-resolved-session-mode
+++ b/.changes/vp-resolved-session-mode
@@ -1,0 +1,1 @@
+patch type="changed" "Audio session mode now follows the voice processing state resolved by the audio device module"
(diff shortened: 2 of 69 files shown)


pblazej (approved):

hiroshihorie:
Converting to draft for now, the delivery logic here may change (mid-capture VP toggles don't re-fire `willEnableEngine`, likely addressed via an upstream event in a future WebRTC build).

PR #1097, opened 2026-08-24
Room dependency stages

pblazej:
Part of CLT-3228 (ad-hoc DI in `Room`) — the smallest slice that buys static safety. No public API change.

`Room`'s per-connection and per-join subsystems lived in an optional `State.transport`, nullable side-slots (`_e2eeManager`, `_dataTracks`), and reset lists ordered by comments and `isFullReconnect:` flags. They are now one dependency stage inside `Room.State`:

```mermaid
stateDiagram-v2
    idle --> connecting: connect() stages ConnectionDependencies (data tracks, E2EE manager)
    connecting --> connected: JOIN builds JoinDependencies (transports + channels)
    connected --> connecting: full reconnect retires the join, keeps the connection
    connecting --> idle: disconnect retires the connection
```

- `JoinDependencies` is only constructible from its `ConnectionDependencies`, by one factory (the old `configureTransports` body): cross-tier init order is compiler-checked, and transports exist iff the stage is `.connected` — the optionals are gone.
- Rebuild = new payload, so nothing leaks across a reconnect by being missing from a reset list; a stage transition and its state reset are one atomic mutation under the existing lock.
- `State.transport`, `Room.dataTracks`, and `Room.e2eeManager` forward into the stage — read sites and the public API are unchanged. `e2eeManager` is now genuinely connection-scoped (nil before connect and after disconnect).
- A duplicate JOIN throws `invalidState` instead of silently no-oping.

Fixes riding along (changeset included), all instances of the forgotten-reset class:

- The legacy `e2eeOptions` branch never installed the crypto manager on the data-channel pairs — such rooms couldn't decrypt incoming encrypted packets, and a reused `Room` kept the previous connection's released manager. Install is now uniform across the options branches (sending still gates on `isDataChannelEncryptionEnabled`), and connection teardown detaches it.
- `DataChannelPair.reset()` restores the default `maxMessageSize` instead of carrying the old session's negotiated value into the next one.

Tested with the Room, DataChannel, DataStream, DataTrack, and RPC suites against a local server — including the new `DependencyStageTests`, `encryptionSurvivesFullReconnect`, and `resetRestoresDefaultMaxMessageSize`.


pblazej:
Planned follow-ups, in rough dependency order (each is its own PR; 1–2 are independent of each other):

1. **State data tiers** — split `Room.State`'s plain data fields into `Connection` ⊃ `Join` sub-structs with forwarders, so `cleanUp`'s hand-maintained keep-list disappears the same way the subsystem reset lists did here. Mostly cherry-pickable from `checkpoint/room-di-tiers-v1`; rider fixes: `connectSpan` surviving region failover, publish-codecs/track-permissions resetting with their tier.

2. **Composition root** — replace the six `lazy var`s with self-less `let` construction wired right after `super.init()`, before `self` escapes. Supersedes the Room half of #1088 (the non-Room fixes in #1089 are needed regardless); requires swapping the `MockDataChannelPair` assignment seam for a send interceptor on the real pair. A validated version exists on the checkpoint branch and applies with minor adaptation.

3. **`runConnection` single-owner task** — one long-lived task owns the connection; full reconnect becomes a loop iteration and teardown gets exactly two call sites. This makes "gates re-arm before the `transport == nil` window opens" structural (currently sequenced by hand in `cleanUpRTC`), and replaces `isFullReconnect:` flags with a cancellation-vs-graceful-shutdown split. Largest behavioral risk; wants a TSan soak.

4. **Finish the join factory** — `Transport` takes its offer handler as an init parameter (the one remaining wire-later gap inside `JoinDependencies.make`), and channel/cryptor composition becomes pure initializer parameters.

5. **Small refinements** — per-join generation IDs so late events from a dead session are dropped by one comparison; a continuation-based intent queue (gRPC-style) replacing `_queuedBlocks`; a serial state-effects stream so `onDidMutate` becomes a pure yield.

hiroshihorie (approved):

PR #1106, opened 2026-09-02
Register the SDK's own transport with livekit-net

pblazej:
`livekit-net` leaves the WebSocket and HTTP clients to the host, so this **wraps the SDK's existing `WebSocket`/`HTTP` stack** in ~180 lines of adapter rather than adding a second, parallel `URLSession` implementation — the Rust side inherits `.callSignaling` QoS, multipath handover and the iOS 26 simulator TLS workaround for free. `WebSocket` also keeps the HTTP status off a rejected upgrade, which `livekit-net` carries as `TransportError::Http` and `livekit-signaling` maps to `SignalError::Handshake`, where a 404 gates the v1 → v0 path fallback.

The tests drive the seam from Rust back through the FFI using `livekit-net`'s own `self_test_*` exports against a live `livekit-server`, so nothing is asserted against a mock.

**Draft:** needs `livekit-uniffi` 0.1.10 — 0.1.9 has no transport seam, so CI can't build this until that release lands (blocked on livekit/rust-sdks#1362).

### DIFF
--- a/Sources/LiveKit/Support/Network/RustTransport.swift
+++ b/Sources/LiveKit/Support/Network/RustTransport.swift
@@ -1,0 +1,182 @@
+/*
+ * Copyright 2026 LiveKit
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+internal import LiveKitUniFFI
+import Foundation
+
+/// Registers the SDK's own network stack as `livekit-net`'s transport, so anything the
+/// Rust side sends goes out over the same ``WebSocket``/``HTTP`` session the Swift path
+/// uses — one place tuning `networkServiceType`, multipath, TLS workarounds and logging.
+///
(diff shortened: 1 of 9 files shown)


pblazej:
Blocked by UniFFI release (soon) 🚫 works locally tho

1egoman (approved):
Generally makes sense to me from a 10k foot view level, but I'm unable to effectively review the lower level swift idiosyncrasies

pblazej:
@1egoman are you going to change the rust side? It's not a high-priority to me (no consumers atm)

1egoman:
> are you going to change the rust side?

No rust changes are planned to this interface AFAIK at the moment, unless you are planning some 😄

Sources/LiveKit/Support/Network/RustTransport.swift
```diff
@@ -0,0 +1,182 @@
...
+// MARK: - HTTP
+
+final class HttpClientAdapter: HttpClient {
+    /// Translate one `livekit-net` HTTP call into a `URLRequest`. Cache policy and timeout
+    /// belong to ``HTTP/request(_:cachePolicy:timeoutInterval:)``, not here.
+    static func makeRequest(method: HttpMethod, url: URL, headers: [Header], body: Data?) -> URLRequest {
+        var request = URLRequest(url: url)
+        request.httpMethod = switch method {
+        case .get: "GET"
+        case .post: "POST"
```
xianshijing-lk, line 96: curiously, do we need to support other https methods ? like PUT, PATCH, DELETE ?

xianshijing-lk (approved):
one question

Sources/LiveKit/Support/Network/RustTransport.swift
pblazej: There are 0 rust consumers at the moment (1st would be telemetry), so this is more like speculation based on full rust feature set, we can double check and maybe update the interface.

Issue #934, opened 2026-03-05
kAudioUnitErr_TooManyFramesToProcess (-10874)


Young1006:
**Describe the bug**
When I started a meeting and turned on the microphone, there was a continuous warning log output from the console. The error seemed to be that the frame buffer size configuration of the core audio unit did not match. The error message indicates that the system requires processing 1440 frames, but the current maximum processing capacity of the audio unit configuration is only 1156 frames.

**SDK Version**
2.12.1

**iOS/macOS Version**
iOS

**Xcode Version**
16.2

**Additional Information**
iPhone 14 (iOS 17.4.1)

**Steps to Reproduce**
1. Start a meeting
2. Turn on the microphone

**Logs**
kAudioUnitErr_TooManyFramesToProcess : inFramesToProcess=1440, mMaxFramesPerSlice=1156
  from AU (0x15921ac70): aumx/mcmx/appl, render err: -10874
  from AU (0x15920f540): augn/sspl/appl, render err: -10874
  from AU (0x15921ac70): aumx/mcmx/appl, render err: -10874
  from AU (0x15920f540): augn/sspl/appl, render err: -10874
  from AU (0x15921ac70): aumx/mcmx/appl, render err: -10874
  from AU (0x15920f540): augn/sspl/appl, render err: -10874
  from AU (0x15921ac70): aumx/mcmx/appl, render err: -10874
  from AU (0x15920f540): augn/sspl/appl, render err: -10874
  from AU (0x15921ac70): aumx/mcmx/appl, render err: -10874
  from AU (0x15920f540): augn/sspl/appl, render err: -10874
  from AU (0x15921ac70): aumx/mcmx/appl, render err: -10874
  from AU (0x15920f540): augn/sspl/appl, render err: -10874
  from AU (0x15921ac70): aumx/mcmx/appl, render err: -10874
  from AU (0x15920f540): augn/sspl/appl, render err: -10874
kAudioUnitErr_TooManyFramesToProcess : inFramesToProcess=1440, mMaxFramesPerSlice=1156
  from AU (0x15921ac70): aumx/mcmx/appl, render err: -10874
  from AU (0x15920f540): augn/sspl/appl, render err: -10874
  from AU (0x15921ac70): aumx/mcmx/appl, render err: -10874
  from AU (0x15920f540): augn/sspl/appl, render err: -10874
  from AU (0x15921ac70): aumx/mcmx/appl, render err: -10874
[...]

Issue #893, opened 2026-01-19
Linker warnings: liblivekit_uniffi.a was built for newer iOS version (26.1)


bguidolim:
## Description

After updating to the latest version of the SDK, Xcode produces numerous linker warnings during the build process. The warnings indicate that object files in `liblivekit_uniffi.a` were built for iOS 26.1, but the project is targeting a lower iOS version.

This appears to be related to PR #822 (Integrate UniFFI). The framework that needs to be rebuilt with a lower deployment target is: https://github.com/livekit/livekit-uniffi-xcframework

## Environment

- **Xcode**: 26.1.1
- **Project deployment target**: iOS 16.0
- **LiveKit SDK version**: 2.11.0

## Warnings

```
ld: object file ([PROJECT_PATH]/DerivedData/Build/Intermediates.noindex/ArchiveIntermediates/[SCHEME]/BuildProductsPath/Release-iphoneos/liblivekit_uniffi.a[558](25ac62e5b3c53843-curve25519.o)) was built for newer 'iOS' version (26.1) than being linked (16.0)
ld: object file ([PROJECT_PATH]/DerivedData/Build/Intermediates.noindex/ArchiveIntermediates/[SCHEME]/BuildProductsPath/Release-iphoneos/liblivekit_uniffi.a[559](0bbbd18bda93c05b-aes_nohw.o)) was built for newer 'iOS' version (26.1) than being linked (16.0)
ld: object file ([PROJECT_PATH]/DerivedData/Build/Intermediates.noindex/ArchiveIntermediates/[SCHEME]/BuildProductsPath/Release-iphoneos/liblivekit_uniffi.a[560](00c879ee3285a50d-montgomery.o)) was built for newer 'iOS' version (26.1) than being linked (16.0)
ld: object file ([PROJECT_PATH]/DerivedData/Build/Intermediates.noindex/ArchiveIntermediates/[SCHEME]/BuildProductsPath/Release-iphoneos/liblivekit_uniffi.a[561](00c879ee3285a50d-montgomery_inv.o)) was built for newer 'iOS' version (26.1) than being linked (16.0)
ld: object file ([PROJECT_PATH]/DerivedData/Build/Intermediates.noindex/ArchiveIntermediates/[SCHEME]/BuildProductsPath/Release-iphoneos/liblivekit_uniffi.a[562](a0330e891e733f4e-ecp_nistz.o)) was built for newer 'iOS' version (26.1) than being linked (16.0)
[...]

PR #1107, opened 2026-09-02
drop signal messages while waiting for reconnect response

lukasIO:
in an effort to align connection behaviour consistently across SDKs, this PR adopts what the rust SDK is currently doing, which is drop all signal messages during a reconnect until a ReconnectResponse is received.

### DIFF
--- a/Sources/LiveKit/Core/SignalClient.swift
+++ b/Sources/LiveKit/Core/SignalClient.swift
@@ -106,6 +106,8 @@ actor SignalClient: Loggable {
         var messageLoopTask: AnyTaskCancellable?
         var lastJoinResponse: Livekit_JoinResponse?
         var rtt: Int64 = 0
+        // Set while a reconnect awaits its Join/ReconnectResponse; other messages in that window are stale.
+        var isAwaitingReconnectResponse: Bool = false
         // Tracks whether the v0 signal path (/rtc) is in use, set during connect.
         // Reused by reconnect to avoid re-attempting the unsupported v1 path.
         var useV0SignalPath: Bool = false
@@ -183,7 +185,10 @@ actor SignalClient: Loggable {
             log("Connecting with url: \(url)")
         }
 
-        _state.mutate { $0.connectionState = (isReconnect ? .reconnecting : .connecting) }
+        _state.mutate {
+            $0.connectionState = (isReconnect ? .reconnecting : .connecting)
+            $0.isAwaitingReconnectResponse = isReconnect
+        }
 
         do {
             let socket = try await WebSocket(url: url,
@@ -282,6 +287,7 @@ actor SignalClient: Loggable {
             $0.socket?.close()
             $0.socket = nil
             $0.lastJoinResponse = nil
+            $0.isAwaitingReconnectResponse = false
         }
 
         _connectResponseCompleter.reset(throwing: disconnectError)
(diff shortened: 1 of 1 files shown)


pblazej:
| SDK | Before `ReconnectResponse` | How staleness is actually scoped |
|---|---|---|
| **Rust** | drops all but `Reconnect`/`Leave` ([`get_reconnect_response`](https://github.com/livekit/rust-sdks/blob/c5844f414fe27a26c1fd87f12ca8a2b7a02270ab/livekit-signaling/src/lib.rs#L1081-L1107)) | [`restart()`](https://github.com/livekit/rust-sdks/blob/c5844f414fe27a26c1fd87f12ca8a2b7a02270ab/livekit-signaling/src/lib.rs#L548-L604) closes the old stream under a write lock; the gate reads the **new** stream's own channel |
| **JS** | **forwards** any non-`leave` first message ([`validateFirstMessage`](https://github.com/livekit/client-sdk-js/blob/e775b6a7fd8e59212481b81263ec9a1e4e539558/src/api/SignalClient.ts#L1218-L1249)) | [`attemptId`](https://github.com/livekit/client-sdk-js/blob/e775b6a7fd8e59212481b81263ec9a1e4e539558/src/api/SignalClient.ts#L1186-L1205) generation counter — a superseded attempt is discarded |
| **Android** | **forwards** first message, buffers + replays the rest ([`handleSignalResponse`](https://github.com/livekit/client-sdk-android/blob/7357a39d670c740d68f7d7bde70246d485901693/livekit-android-sdk/src/main/java/io/livekit/android/room/SignalClient.kt#L668-L741)) | [`ws != currentWs`](https://github.com/livekit/client-sdk-android/blob/7357a39d670c740d68f7d7bde70246d485901693/livekit-android-sdk/src/main/java/io/livekit/android/room/SignalClient.kt#L297-L331) in five places |
| **Flutter** | no gate ([`_onSocketData`](https://github.com/livekit/client-sdk-flutter/blob/b0e5db286a2f175f2f6027c796318cdc900d248c/lib/src/core/signal_client.dart#L274)) | — |
| **Swift (this PR)** | drops by message type | one `Bool` in state shared across connections |

The "align across SDKs" claim holds for Rust only, and in every SDK that has a gate the thing doing the real work is connection scoping, not the message-type filter. Swift's `Bool` outlives the connection and `cleanUp()` clears it *before* `await _responseQueue.clear()`, so a straggler from the dead socket still passes the gate — a per-connection generation counter (JS's `attemptId` is exactly this) gives the real invariant and makes the deny-list unnecessary. Worth noting both prior fixes in this area *narrowed* a drop ([rust #893](https://github.com/livekit/rust-sdks/pull/893), [android #934](https://github.com/livekit/client-sdk-android/pull/934)), and `refreshToken`/`trackSubscribed`/`trackUnpublished`/`roomMoved` have no server replay path.

lukasIO:
> gives the real invariant and makes the deny-list unnecessary. 

how so? if the server sends updates out of order or rather other messages prior to a ReconnectResponse on a fresh socket connection, that would still be an issue?

pblazej:
On a fresh socket they aren't applied early either: `connect()` starts with [`await cleanUp()`](https://github.com/livekit/client-sdk-swift/blob/e86e8e4c81862091ea58cbf1343ccebfae55bb07/Sources/LiveKit/Core/SignalClient.swift#L156), which [`clear()`s `_responseQueue` to `.suspended`](https://github.com/livekit/client-sdk-swift/blob/e86e8e4c81862091ea58cbf1343ccebfae55bb07/Sources/LiveKit/Support/Schedulers/QueueActor.swift#L61-L68) before the new socket is even opened, so everything but `join`/`reconnect`/`leave` is enqueued and replayed in wire order by `resumeQueues()` after `configureTransports`. So the deny-list isn't changing *when* those messages are applied — it's changing buffered → discarded.

That said, I think you're pointing at something real, just one layer down: the ordering isn't actually guaranteed today, because each message is dispatched with its own `Task.detached`, so enqueue order isn't tied to wire order — a message that arrived before the `ReconnectResponse` can have its task run after `resumeQueues()` and get processed live, out of order against the replay. The deny-list does paper over that during reconnect, but it's pre-existing and applies in steady state too, so I'd rather fix the dispatch than filter by type. (The three data-track types are the one genuinely live path — they bypass the queue by design.)

PR #1109, opened 2026-09-04
Add custom video encoder factory support via SDK-owned protocols

hiroshihorie:
Adds `LiveKitSDK.set(videoEncoderFactory:)` with SDK-owned `VideoEncoderFactory` / `VideoEncoder` protocols, so apps can supply their own H264 or H265 encoder (e.g. a software openH264 encoder) without `LiveKitWebRTC` types in the public API. Supersedes #1082, thanks @sergeyphi for the PR and the detailed testing notes.

Addresses the requests from that thread:
- `VideoFrame.rtpTimestamp` carries the 90kHz RTP timestamp WebRTC assigns before `encode`, and `EncodedVideoFrame.rtpTimestamp` must be copied from it.
- `VideoBuffer.toI420()` / `VideoFrame.toI420()` give access to I420 planes for NV12 camera frames. `supportsNativeHandle` only gates crop/scale in WebRTC, so this is the route to planar data.
- `VideoEncoderSettings` has a public init for testing.
- Encode timing and NTP fields were requested but are intentionally not exposed: `FrameEncodeMetadataWriter` overwrites them for every frame it matches by RTP timestamp, so encoder-provided values never reach stats or the wire.

Scope for v1, kept to what demonstrably reaches WebRTC (verified against the fork source):
- H264 and H265 only. VP8/VP9 need per frame layer info and AV1 a dependency descriptor, none of which the ObjC `RTCCodecSpecificInfo` bridge can carry, so `set` rejects them. AV1 can follow once the bridge can emit a descriptor.
- `EncodedVideoFrame` carries `data`, `dimensions`, `rtpTimestamp`, `frameType`, `qp` and an optional H264 `packetizationMode`. Capture time, rotation and content type are filled by WebRTC from its own record of the source frame, so they are not part of the type.
- Codec specific info is always built for the codec the encoder was created for. Frames without a mode follow the negotiated `packetization-mode`; an H264 format advertised without that parameter is normalized to mode 1, matching the built in factory and how frames are packetized.
- `VideoEncoderStatus` names only codes WebRTC acts on. Positive codes are omitted since the simulcast adapter stops encoding the remaining layers on anything other than OK.
- `resolutionAlignment` and `scalabilityModes` are left out: the ObjC bridge always reports alignment 1 (fix in webrtc-sdk/webrtc#289) and the bridge reports every encoder as hardware accelerated. Both return as additive changes once a WebRTC release carries the fixes.
- The built in VideoToolbox encoders remain the simulcast fallback, both for codecs the factory declines and when an encoder returns `.fallbackSoftware`. `set` throws if either the encoder factory or the peer connection factory has already been initialized.

Testing: 12 unit tests exercise the bridge with fake factories and encoders (validation and normalization, codec gate, frame type arity, callback lifecycle, codec info selection). Compile verified on macOS and iOS Simulator. No real encoder has run through the path yet, and H265 is compile verified only. Swift only for now; ObjC encoders need a thin Swift shim.

### DIFF
--- a/Sources/LiveKit/Core/LiveKitSDK+VideoEncoderFactory.swift
+++ b/Sources/LiveKit/Core/LiveKitSDK+VideoEncoderFactory.swift
@@ -1,0 +1,45 @@
+/*
+ * Copyright 2026 LiveKit
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import Foundation
+
+public extension LiveKitSDK {
+    /// Sets a custom ``VideoEncoderFactory`` used to create video encoders,
+    /// replacing the SDK's default VideoToolbox backed encoders.
+    ///
+    /// The factory is wrapped in WebRTC's simulcast encoder adapter, so each
+    /// simulcast layer is encoded by an encoder created from this factory.
+    ///
+    /// Pass `nil` to restore the default factory.
+    ///
+    /// ```swift
+    /// try LiveKitSDK.set(videoEncoderFactory: MyEncoderFactory())
+    /// let room = Room()
+    /// try await room.connect(url: url, token: token)
+    /// ```
+    ///
+    /// - Warning: Must be called before the first `Room` connection or track
+    ///   creation. Throws ``LiveKitError`` with type `.invalidState` afterwards,
+    ///   since the underlying peer connection factory is created once per process.
+    static func set(videoEncoderFactory: (any VideoEncoderFactory)?) throws {
+        try RTC.pcFactoryState.mutate {
+            guard !$0.isInitialized else {
+                throw LiveKitError(.invalidState, message: "Cannot set videoEncoderFactory after the peer connection factory has been initialized")
+            }
+            $0.customVideoEncoderFactory = videoEncoderFactory
+        }
+    }
+}
--- a/Sources/LiveKit/Core/LiveKitSDK+VideoEncoderFactory.swift
+++ b/Sources/LiveKit/Core/LiveKitSDK+VideoEncoderFactory.swift
@@ -21,7 +21,9 @@ public extension LiveKitSDK {
     /// replacing the SDK's default VideoToolbox backed encoders.
     ///
     /// The factory is wrapped in WebRTC's simulcast encoder adapter, so each
-    /// simulcast layer is encoded by an encoder created from this factory.
+    /// simulcast layer is encoded by an encoder created from this factory, with
+    /// the SDK's built in encoders as the fallback when an encoder reports
+    /// ``VideoEncoderStatus/fallbackSoftware``.
     ///
     /// Pass `nil` to restore the default factory.
     ///
--- a/Sources/LiveKit/Core/LiveKitSDK+VideoEncoderFactory.swift
+++ b/Sources/LiveKit/Core/LiveKitSDK+VideoEncoderFactory.swift
@@ -33,10 +33,16 @@ public extension LiveKitSDK {
     /// try await room.connect(url: url, token: token)
     /// ```
     ///
-    /// - Warning: Must be called before the first `Room` connection or track
-    ///   creation. Throws ``LiveKitError`` with type `.invalidState` afterwards,
-    ///   since the underlying peer connection factory is created once per process.
+    /// - Warning: This method must be called before any other SDK API is used,
+    ///   e.g. in the `App.init()` or `application(_:didFinishLaunchingWithOptions:)`.
+    ///   Any access to the peer connection factory, such as connecting, creating a
+    ///   track, querying capabilities or setting up E2EE, initializes it once per
+    ///   process, and this method throws ``LiveKitError`` with type `.invalidState`
+    ///   afterwards.
     static func set(videoEncoderFactory: (any VideoEncoderFactory)?) throws {
+        if let videoEncoderFactory, videoEncoderFactory.supportedCodecs.isEmpty {
+            throw LiveKitError(.invalidParameter, message: "videoEncoderFactory must advertise at least one supported codec")
+        }
         try RTC.pcFactoryState.mutate {
             guard !$0.isInitialized else {
                 throw LiveKitError(.invalidState, message: "Cannot set videoEncoderFactory after the peer connection factory has been initialized")
--- a/Sources/LiveKit/Core/LiveKitSDK+VideoEncoderFactory.swift
+++ b/Sources/LiveKit/Core/LiveKitSDK+VideoEncoderFactory.swift
@@ -17,13 +17,18 @@
 import Foundation
 
 public extension LiveKitSDK {
-    /// Sets a custom ``VideoEncoderFactory`` used to create video encoders,
-    /// replacing the SDK's default VideoToolbox backed encoders.
+    /// Sets a custom ``VideoEncoderFactory`` used to create video encoders for
+    /// the codecs it supports, taking over from the SDK's default VideoToolbox
+    /// backed encoders for those codecs.
     ///
     /// The factory is wrapped in WebRTC's simulcast encoder adapter, so each
-    /// simulcast layer is encoded by an encoder created from this factory, with
-    /// the SDK's built in encoders as the fallback when an encoder reports
-    /// ``VideoEncoderStatus/fallbackSoftware``.
+    /// simulcast layer is encoded by an encoder created from this factory. The
+    /// SDK's built in encoders remain the fallback, both for codecs the factory
+    /// declines and when an encoder reports ``VideoEncoderStatus/fallbackSoftware``.
+    ///
+    /// Only H264, H265 and AV1 can be supplied by a custom factory. VP8 and VP9
+    /// require codec specific layer information that the bridge cannot carry
+    /// yet, so advertising them throws ``LiveKitError`` with type `.invalidParameter`.
     ///
     /// Pass `nil` to restore the default factory.
     ///
@@ -40,8 +45,15 @@ public extension LiveKitSDK {
     ///   process, and this method throws ``LiveKitError`` with type `.invalidState`
     ///   afterwards.
     static func set(videoEncoderFactory: (any VideoEncoderFactory)?) throws {
-        if let videoEncoderFactory, videoEncoderFactory.supportedCodecs.isEmpty {
-            throw LiveKitError(.invalidParameter, message: "videoEncoderFactory must advertise at least one supported codec")
+        if let videoEncoderFactory {
+            let codecs = videoEncoderFactory.supportedCodecs
+            guard !codecs.isEmpty else {
+                throw LiveKitError(.invalidParameter, message: "videoEncoderFactory must advertise at least one supported codec")
+            }
+            let unsupported = codecs.map(\.name).filter { !Self.bridgeableCodecNames.contains($0.uppercased()) }
+            guard unsupported.isEmpty else {
+                throw LiveKitError(.invalidParameter, message: "videoEncoderFactory cannot supply encoders for \(unsupported.joined(separator: ", ")), only H264, H265 and AV1 are supported")
+            }
         }
         try RTC.pcFactoryState.mutate {
             guard !$0.isInitialized else {
@@ -50,4 +62,8 @@ public extension LiveKitSDK {
             $0.customVideoEncoderFactory = videoEncoderFactory
         }
     }
+
+    /// Codecs whose RTP packetization needs no codec specific info beyond what
+    /// ``EncodedVideoFrame/CodecSpecificInfo`` can express.
+    private static let bridgeableCodecNames: Set<String> = ["H264", "H265", "AV1"]
 }
--- a/Sources/LiveKit/Core/LiveKitSDK+VideoEncoderFactory.swift
+++ b/Sources/LiveKit/Core/LiveKitSDK+VideoEncoderFactory.swift
@@ -45,8 +45,10 @@ public extension LiveKitSDK {
     ///   process, and this method throws ``LiveKitError`` with type `.invalidState`
     ///   afterwards.
     static func set(videoEncoderFactory: (any VideoEncoderFactory)?) throws {
-        if let videoEncoderFactory {
-            let codecs = videoEncoderFactory.supportedCodecs
+        // Read once and stored alongside the factory, so validation, the advertised
+        // list and the codecs the adapter will accept all come from the same snapshot.
+        let codecs = videoEncoderFactory?.supportedCodecs ?? []
+        if videoEncoderFactory != nil {
             guard !codecs.isEmpty else {
                 throw LiveKitError(.invalidParameter, message: "videoEncoderFactory must advertise at least one supported codec")
             }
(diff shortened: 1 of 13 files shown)


Sources/LiveKit/Types/EncodedVideoFrame.swift
```diff
@@ -0,0 +1,187 @@
...
+        image.contentType = contentType == .screenshare ? .screenshare : .unspecified
+        // Encode timing, NTP time and timing flags are filled in by WebRTC itself
+        // once the encoded image is matched to its source frame by RTP timestamp,
+        // so they are intentionally not part of the public type.
+        // Always set: the native side reads `intValue`, so a nil would land as 0,
+        // while -1 is what the quality scaler treats as unknown.
+        image.qp = NSNumber(value: qp ?? -1)
+
+        let info: CodecSpecificInfo? = switch codecSpecificInfo {
+        case let .some(info): info
```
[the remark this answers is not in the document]
hiroshihorie: Fixed in d6be745b. The adapter now keeps the full VideoCodecInfo and toRTCType always builds the codec specific info for the codec the encoder was created for, taking only the packetization mode from the frame. A mismatched header can no longer reach the packetizer.

Sources/LiveKit/Types/EncodedVideoFrame.swift
```diff
@@ -0,0 +1,187 @@
...
+        case nil:
+            switch codecName.uppercased() {
+            case "H264": .h264(packetizationMode: .nonInterleaved)
+            case "H265": .h265(packetizationMode: .nonInterleaved)
```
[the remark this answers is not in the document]
hiroshihorie: Fixed in d6be745b. The nil default now follows the negotiated packetization-mode parameter, so mode 0 yields single NAL unit packetization. When the parameter is absent we keep non interleaved on purpose: RFC 6184 says mode 0, but in single NAL mode WebRTC drops any frame whose NAL unit exceeds the packet size, so a factory that omits the parameter would silently lose most key frames. The built in encoders make the same choice and receivers depacketize both modes regardless.

Sources/LiveKit/Core/RTC.swift
```diff
@@ -129,17 +129,25 @@ extension RTC {
...
+        var customVideoEncoderFactory: (any VideoEncoderFactory)?
     }
 
     static let pcFactoryState = StateSync(PeerConnectionFactoryState())
 
     // global properties are already lazy
 
+    // Must not be forced from inside a `pcFactoryState` read or mutate block, since
+    // its initializer reads that state and `StateSync` is not reentrant.
     static let encoderFactory: LKRTCVideoEncoderFactory & Sendable = {
```
pblazej, line 148: One gap in the guard here: `set(videoEncoderFactory:)` checks `pcFactoryState.isInitialized`, but that flag is only set inside `peerConnectionFactory`'s initializer — `encoderFactory` is an independent lazy static that can be resolved first, which `CodecTests.swift:41` already does (`RTC.encoderFactory.supportedCodecs()`).

Once that has happened, `set()` passes the guard, stores the factory and returns success with no effect: the app publishes with the built-in VideoToolbox encoders while believing its own encoder is installed. Might be worth tracking whether `encoderFactory` itself has been resolved, since a silent no-op is hard to diagnose from the outside.

Sources/LiveKit/Core/LiveKitSDK+VideoEncoderFactory.swift
```diff
@@ -0,0 +1,69 @@
...
+    ///
+    /// - Warning: This method must be called before any other SDK API is used,
+    ///   e.g. in the `App.init()` or `application(_:didFinishLaunchingWithOptions:)`.
+    ///   Any access to the peer connection factory, such as connecting, creating a
+    ///   track, querying capabilities or setting up E2EE, initializes it once per
+    ///   process, and this method throws ``LiveKitError`` with type `.invalidState`
+    ///   afterwards.
+    static func set(videoEncoderFactory: (any VideoEncoderFactory)?) throws {
+        if let videoEncoderFactory {
+            let codecs = videoEncoderFactory.supportedCodecs
```
pblazej: Since this check is what keeps a misbehaving factory away from the packetizer abort, it might be worth tightening in two places:

1. `supportedCodecs` is a computed property, read here and again in `VideoEncoderFactoryAdapter.init`. A factory returning `[H264]` here and `[H264, VP8]` there would get VP8 into `supportedRTCCodecs` and the SDP without a second look — snapshotting the array once and passing it to the adapter would close that.
2. `RTCVideoEncoderFactorySimulcast` asks the primary factory for whatever was negotiated, so `createEncoder(for:)` can still see VP8 even when it was never advertised. If a factory doesn't filter, `toRTCType(codec:)` falls to `default: nil` → generic info → `rtp_video_header.codec == kVideoCodecGeneric` with a monostate `video_type_header`, while `RtpPacketizer::Create` still uses the negotiated `kVP8` → `std::get<RTPVideoHeaderVP8>(monostate)` → `__throw_bad_variant_access` under `-fno-exceptions`. A name check in `createEncoder` would cover it, since that's the point that sees the negotiated codec.

Sources/LiveKit/Types/EncodedVideoFrame.swift
```diff
@@ -0,0 +1,192 @@
...
+
+    /// Resolution of the encoded frame.
+    public let dimensions: Dimensions
+
+    /// RTP timestamp in the 90kHz clock, which must be copied from the source
+    /// ``VideoFrame/rtpTimestamp`` and not derived from `timeStampNs`.
+    public let rtpTimestamp: UInt32
+
+    /// Capture time in milliseconds.
+    public let captureTimeMs: Int64
```
pblazej: The encode-timing/NTP reasoning in the description seems to hold for a few more fields than it's applied to: `frame_encode_metadata_writer.cc:328` overwrites `content_type_` unconditionally, and `:336-340` overwrite `capture_time_ms_`, `ntp_time_ms_` and `rotation_` whenever the RTP timestamp matches — which it always will, given the API mandates copying `rtpTimestamp`. That leaves `captureTimeMs` as a non-defaulted init parameter every caller has to compute for nothing, and `.screenshare` documented as affecting the receiver when it can't.

A few others in the same category, worth a look now mostly because they're awkward to remove after release:
- `resolutionAlignment` / `applyAlignmentToAllSimulcastLayers` — the bridge always reports 1 (already noted in the description)
- `.h265(packetizationMode:)` — `rtp_format.cc`'s kH265 branch constructs `RtpPacketizerH265(payload, limits)` with no mode argument
- `VideoCodecInfo.scalabilityModes` — without a `queryCodecSupport` override the default reports every non-nil mode as unsupported
- AV1 is bridgeable but can't emit a dependency descriptor: generic info → `kVideoCodecGeneric` → `GenericToGeneric`, `GenericStructure` returns nullopt so `SetVideoStructure` is never called

No strong opinion on how to resolve it — one option would be H264/H265 for v1, with the rest added alongside the fork fixes.

Sources/LiveKit/Protocols/VideoEncoderFactory.swift
```diff
@@ -0,0 +1,43 @@
...
+import Foundation
+
+/// Provides custom video encoders to the SDK, mirroring WebRTC's `VideoEncoderFactory`.
+///
+/// Register a factory via ``LiveKitSDK/set(videoEncoderFactory:)`` before connecting
+/// to a ``Room`` to replace the SDK's default VideoToolbox backed encoders.
+///
+/// ```swift
+/// final class MyEncoderFactory: VideoEncoderFactory {
+///     var supportedCodecs: [VideoCodecInfo] { [VideoCodecInfo(name: "VP8")] }
```
pblazej: Small thing: the example advertises VP8, which is what `set()` rejects, so the `try LiveKitSDK.set(videoEncoderFactory:)` on the last line throws `.invalidParameter`. H264 would make it copy-pasteable.

Sources/LiveKit/Core/VideoEncoderFactoryAdapter.swift
```diff
@@ -0,0 +1,160 @@
...
+    }
+}
+
+/// Bridges a public ``VideoEncoder`` to WebRTC's `RTCVideoEncoder`.
+final class VideoEncoderAdapter: NSObject, LKRTCVideoEncoder, @unchecked Sendable {
+    /// Holds the current WebRTC callback so the closure handed to the encoder can
+    /// stay the same object across `setCallback` calls, and so clearing it cannot
+    /// race with a frame being delivered from an encoder thread.
+    private final class CallbackBox: @unchecked Sendable {
+        private let lock: some Lock = createLock()
```
pblazej: Reading through the box, I think it earns its keep for a reason the docstring doesn't mention: `SimulcastEncoderAdapter::StreamContext`'s move ctor (`simulcast_encoder_adapter.cc:207`) re-registers a new `EncodedImageCallback*` with no intervening nullptr when `stream_contexts_` reallocates, so something has to absorb that. Three ways to shape it:

- **No box, fresh closure per `setCallback`.** Simplest, but it re-enters the app's `setCallback` on that reallocation, and a mutable block reference read from an encoder thread needs real synchronization for the ARC traffic alone.
- **Holding the lock across `invoke`** (current). I don't think this can quite deliver the guarantee in the comment: the block captures the raw `EncodedImageCallback*` by value (`objc_video_encoder_factory.mm:56`), so retaining it doesn't keep the C++ target alive and the lock narrows the window rather than closing it. Meanwhile it puts a non-recursive lock around `OnEncodedImage` → packetize → `FrameEncryptor` → pacer on every frame, and `release()`/`setCallback` wait behind a full send pass.
- **Copy out, invoke outside the lock, via `StateSync`.** My preference — it keeps the stale-callback drop, avoids the reentrancy case entirely, and stays with the existing primitive. `StateSync<State>` has no `Sendable` bound on `State`, so it can hold the block directly:

```swift
private final class CallbackBox: @unchecked Sendable {
    private struct State {
        var callback: RTCVideoEncoderCallback?
        var isAttached = false
    }

    private let state = StateSync(State())

    func set(_ callback: RTCVideoEncoderCallback?) -> Bool {
        state.mutate {
            $0.callback = callback
            guard callback != nil, !$0.isAttached else { return false }
            $0.isAttached = true
            return true
        }
    }

    func clear() { state.mutate { $0.callback = nil; $0.isAttached = false } }

    func invoke(_ image: LKRTCEncodedImage, _ info: any LKRTCCodecSpecificInfo) -> Bool {
        // Copied out so the callback runs outside the lock.
        guard let callback = state.read({ $0.callback }) else { return false }
        return callback(image, info)
    }
}
```

Builds clean on macOS, swiftlint clean, +22/−20 if it's useful.

One small correction to the docstring while you're here: `clear()` resets `isAttached`, so a re-attach after `release()` does hand the encoder a new closure — the stability holds across consecutive non-nil registrations rather than all `setCallback` calls.

Sources/LiveKit/Core/RTC.swift
hiroshihorie: Good catch, thanks! Fixed in 3ab68fcb: encoderFactory now records its own resolution in a separate flag and set() rejects on either. Kept isInitialized as is since the audio guards read it as the peer connection factory existing.

Sources/LiveKit/Core/LiveKitSDK+VideoEncoderFactory.swift
hiroshihorie: Both done in 842479fc. supportedCodecs is read once in set() and stored with the factory, and the adapter declines anything outside that snapshot so the built in encoder takes it.

Sources/LiveKit/Core/VideoEncoderFactoryAdapter.swift
hiroshihorie: Agreed on all points, took your StateSync version as is in 42ef5161, with the docstring rewritten around the StreamContext re-registration and the corrected stability claim. Thanks for the thorough read!

Sources/LiveKit/Protocols/VideoEncoderFactory.swift
hiroshihorie: Fixed in 9c718c2f, the example uses H264 now.

Sources/LiveKit/Types/EncodedVideoFrame.swift
hiroshihorie: Agree with the analysis and with H264/H265 only for v1. Will trim the overwritten fields and the H265 mode parameter in a follow up commit on this PR.

pblazej (commented):
LGTM, I think the last one: https://github.com/livekit/client-sdk-swift/pull/1109#discussion_r3950353826

Sources/LiveKit/Types/EncodedVideoFrame.swift
hiroshihorie: Done in 7775a9ac and follow ups: captureTimeMs, rotation and contentType are gone from EncodedVideoFrame, the CodecSpecificInfo enum is now a single optional H264 packetizationMode, resolutionAlignment / applyAlignmentToAllSimulcastLayers and scalabilityModes are dropped, and AV1 is out of the bridgeable set so set() accepts H264 and H265 only. Also dropped the positive status codes since the simulcast adapter returns early on anything other than OK, and added 12 unit tests around the bridge. Everything removed can come back additively once the fork carries the missing pieces.

Tests/LiveKitCoreTests/VideoEncoderFactoryTests.swift
pblazej, line 1: Nice 👍

pblazej (approved):

sergeyphi:
@hiroshihorie Thanks for turning this around so fast. I ported our software openH264 encoder onto this branch and it fits the new protocols cleanly. Confirmed it's working end-to-end 👍 

Both of the blockers from #1082 are gone:

- `VideoFrame.rtpTimestamp` is exactly what we needed
- `VideoBuffer.toI420()` gives us the planes, and `I420VideoBuffer` exposes everything openH264 wants

A few things I ran into doing the port.

**1. No way to make the custom factory exclusive**

This is the one that matters for us. Our receiver's hardware decoder can't handle VideoToolbox output, so publishing has to go through our encoder or not at all. Right now there are two ways back to VideoToolbox we can't turn off: it stays the simulcast fallback (so its H264 profiles are advertised alongside ours), and `.fallbackSoftware` routes to it.

842479fc goes the other direction (declining codecs we didn't advertise), which is right for the VP8 abort but doesn't cover this.

On our old fork we passed the custom factory as both primary and fallback. With the codec snapshot in place that's nearly a one-liner: `fallback: adapter` behind an opt-in like `set(videoEncoderFactory:, exclusive: true)`. Happy to send a PR if you like the shape.

**2. `supportsNativeHandle` docstring**

It says `false` only affects crop/scale and frames still arrive native. My reading of `VideoStreamEncoder` is that `false` also makes WebRTC convert to I420 upstream, so `toI420()` becomes a no-op. I haven't confirmed at runtime, so I may be misreading, but it decides whether a planar-only encoder should report `false`, so worth pinning down.

**3. `I420VideoBuffer` pointer lifetime**

`dataY/U/V` are raw pointers into a struct the compiler may release after the last property read, so this mostly works and is wrong:

```swift
let i420 = frame.toI420()!
core.encode(y: i420.dataY, u: i420.dataU, v: i420.dataV, ...)   // pointers may already be dead
```

We wrapped it in `withExtendedLifetime`. A scoped `withUnsafePlanes { }` accessor would make the correct version the only one; failing that, a note in the doc comments.

**4. Two heads-ups for anyone consuming this early**

We cherry-picked the PR onto the v2.16.0 tag rather than taking the branch, because two things on `main` broke our app build:

- `Transport` is `@RTC`-isolated now, so a plain `static` helper added there can't be called from the signal-client delegate. Trivial once you know (`nonisolated static`, like the existing munge helpers).
- The nanopb headers (`#include <pb.h>`) resolve under `swift build` but not when Xcode builds the package — `unknown type name 'pb_msgdesc_t'` and the `PB_PROTO_HEADER_VERSION` guard firing. I didn't chase it, so treat as a report, but it may be worth checking `CLiveKitProto` builds from Xcode.

Your commits cherry-pick onto the tag cleanly, so nothing here is a problem with the PR itself.

**5. Small**

A line in the `VideoEncoder` docs on the threading model: what's sequenced on the encoder queue, and that the callback shouldn't be invoked under a lock (42ef5161's reasoning applies to app-side encoders too; we changed ours to match).

Happy to run any further change through the same device setup before it ships!

PR #1113, opened 2026-09-04
Default to a single peer connection

xianshijing-lk:
Ticket 4 of the connect-latency series. Stacked on #1111 — retarget to `main` once #1110 and #1111 merge.

`RoomOptions.singlePeerConnection` has defaulted to `false` in Swift while client-sdk-js (`defaults.ts:45`) and rust-sdks (`room/mod.rs:471`) have both defaulted to `true`. Swift was the only one of the three opening two peer connections by default — two ICE and DTLS handshakes, and no access to any v1-only connect optimization.

## Why this is the change that matters

Benchmarked on LiveKit Cloud staging, `BM-CONN-001` / `BM-CONN-003`, 25 iterations + 5 warmups each:

| Config | wall p50 | wall p90 | D_WS p50 | **D_TRANSPORT p50** |
|---|---|---|---|---|
| `main` + dual PC *(ships today)* | 508 | 553 | 206 | 298 |
| `main` + single PC | 511 | 675 | 212 | 291 |
| #1110+#1111 + dual PC | 514 | 552 | 214 | 298 |
| **#1110+#1111 + single PC** *(this PR's default)* | **482** | 555 | 219 | **242** |

Read `D_TRANSPORT_MS` (join_recv → pc_connected); `D_WS_MS` is pure TLS/WS handshake and is flat across all four, so it is the noise floor.

**~50 ms (−17%) off the transport phase.** That is one round trip, matching the RTT implied by `D_WS_MS` ≈ 206 ms spanning ~3–4 RTTs. The saving is one RTT, so it scales with client-to-region latency — larger on the high-RTT paths that motivated this work.

The row that justifies the stack: **single PC alone buys nothing** (291 vs 298). The gain is entirely offer-with-join from #1111 — and #1111 only runs on the single-PC path, so without this flip it reaches nobody. Neither PR is worth much without the other.

## Compatibility

`/rtc/v1` requires LiveKit Cloud or LiveKit OSS ≥ 1.9.2. Older servers 404 the endpoint, and the existing `serviceNotFound` fallback (`Room+Engine.swift`) re-runs the connect on the legacy dual-PC path — verified by `v1LocalhostFallback`, which passes. The cost on an old server is one failed request, not a failed connect.

Opt out with `RoomOptions(singlePeerConnection: false)`.

## Testing

Against **LiveKit Cloud staging**:

```
PeerConnectionSignalingTests + RoomTests — 15/15 passed, both PC modes
```

Against **local `livekit-server` 1.13.1**: same suites green.

Builds verified on macOS, Mac Catalyst, iOS Simulator. `swiftlint` and `swiftformat --lint` clean. No API signature change — only default values, so `api-check` should be unaffected.

## ⚠️ Known pre-existing flake, worth deciding on before merge

`dataChannel(mode: V1 Single PC)` intermittently times out at `PeerConnectionSignalingTests.swift:281`. Measured rate against staging:

- on this branch: 1 failure in 4 runs
- on the base (#1111), single PC explicitly: **1 failure in 3 runs**

So this PR does not cause it — it is a pre-existing single-PC data-channel race. But it is the reason to think twice: today almost nobody runs single PC, and this change makes it the path every user takes, so a mode-specific flake becomes a mode-specific production issue.

My recommendation is to root-cause that timeout before or alongside merging this, rather than after. Happy to take it as a follow-up ticket — it looks like a data channel open race on the single-PC path, but I have not diagnosed it.


Sources/LiveKit/Types/Options/RoomOptions.swift
```diff
@@ -79,7 +85,7 @@ public final class RoomOptions: NSObject, Sendable, Loggable {
         e2eeOptions = nil
         encryptionOptions = nil
         reportRemoteTrackStatistics = false
-        singlePeerConnection = false
+        singlePeerConnection = true
```
pblazej: This makes `.publisherOnly` the default data path, and `ensurePublisherConnected()` returns at its `guard case .subscriberPrimary` (`Room+Engine.swift:77`) — so `send(dataPacket:)` never reaches the `publisherDataChannel.openCompleter.wait()` gate just below it. `DataChannelPair`'s own docstring gives that await as the reason a never-connected transport "fails in bounded time instead of hanging forever", so a `publishData()` straight after connect parks at the drain head with no timeout. That matches the `dataChannel(V1 Single PC)` timeout in your known-flake section, which I think makes it a blocker rather than a follow-up.

Sources/LiveKit/Types/Options/RoomOptions.swift
```diff
@@ -95,7 +101,7 @@ public final class RoomOptions: NSObject, Sendable, Loggable {
                 e2eeOptions: E2EEOptions? = nil,
                 encryptionOptions: EncryptionOptions? = nil,
                 reportRemoteTrackStatistics: Bool = false,
-                singlePeerConnection: Bool = false)
+                singlePeerConnection: Bool = true)
```
pblazej, line 104: `Tests/LiveKitTestSupport/Room.swift:54` still defaults this to `false`, so after the flip every E2E test outside `PeerConnectionSignalingTests` exercises a non-default configuration. Worth flipping there too, keeping explicit dual-PC cases where dual PC is the subject.

Sources/LiveKit/Types/Options/RoomOptions.swift
```diff
@@ -62,7 +62,13 @@ public final class RoomOptions: NSObject, Sendable, Loggable {
...
     /// Use a single peer connection for both publishing and subscribing.
     ///
-    /// - Note: Requires LiveKit Cloud or LiveKit OSS >= 1.9.2.
+    /// Defaults to `true`, matching client-sdk-js and rust-sdks. One connection means one ICE and
+    /// DTLS handshake instead of two, and it is the only mode that can carry the publisher offer
+    /// in the JOIN request — which removes an offer/answer round trip from the connect path.
+    ///
+    /// - Note: Requires LiveKit Cloud or LiveKit OSS >= 1.9.2. Against an older server the SDK
+    ///   falls back to the legacy dual peer connection path automatically, so setting this has no
+    ///   effect there beyond one extra failed request.
```
pblazej: Against OSS < 1.9.2 it's a little more than one extra failed request: `singlePC` is re-read from `_state.roomOptions` at `Room+Engine.swift:234`, so every `fullConnectSequence` — initial connect and each full-reconnect attempt — builds an `EarlyPublisher` (publisher PC, three data channels, initial offer), fails the upgrade, runs the validate round trip, tears it down, then builds a second PC on the v0 retry. Caching the v0 outcome on `Room` would skip all of that after the first attempt.

pblazej (commented):
Two lines of diff, but this is the change that exposes the rest of the stack, so the notes below are really about what becomes the default rather than about this file. The first one I believe is the `dataChannel(V1 Single PC)` flake you flagged, which would move it from follow-up to blocker.

PR #1114, opened 2026-09-05
Add end-to-end session lifecycle benchmarks

xianshijing-lk:
`BM-CONN` decomposes `connect()` into its internal phases — good for "where does connect time go", not for "how long does a user wait for the thing they asked for". `BM-SESSION` measures the milestones themselves, across two participants in one room.

## Metrics

All wall-clock milliseconds, all from one session per iteration so the sums hold per-run:

| metric | measured from → to |
|---|---|
| `connect_ms` | publisher `room.connect()` call → return |
| `peer_visible_ms` | subscriber `connect()` call → publisher's `participantDidConnect` |
| `pub_to_sub_ms` | `publish(videoTrack:)` call → subscriber's `didSubscribeTrack` |
| `first_frame_ms` | `didSubscribeTrack` → first frame at the subscriber's renderer |
| `publish_to_first_frame_ms` | `publish` call → first frame (end to end) |
| `data_rtt_ms` | publisher sends on a topic → subscriber echoes → publisher receives |
| `disconnect_ms` | publisher `room.disconnect()` call → return |

Two variants, `BM-SESSION-001-DualPC` and `BM-SESSION-002-SinglePC`, mirroring `BM-CONN`.

Frames come from a buffer-backed track fed synthetic pixel buffers rather than a camera, so this runs headless with no device permissions.

## First results (LiveKit Cloud staging, 15 iterations)

| metric | dual PC p50 | single PC p50 | delta |
|---|---|---|---|
| `connect_ms` | 556 | 489 | **−67 (−12%)** |
| `peer_visible_ms` | 210 | 208 | −2 |
| `pub_to_sub_ms` | 222 | **308** | **+86 (+39%)** |
| `first_frame_ms` | 192 | 103 | −89 |
| `publish_to_first_frame_ms` | 412 | 412 | **0** |
| `data_rtt_ms` | 124 | 126 | +2 |
| `disconnect_ms` | 5 | 4 | −1 |

The reason to have this benchmark, in one row: **single PC buys 67 ms on connect and gives back 86 ms on publish→subscribe, netting exactly zero on time-to-first-frame.** The connect-phase benchmark cannot see that, because the cost lands after `connect()` returns.

The `pub_to_sub` regression is the `MediaSectionsRequirement` round trip: in single PC the client must add receive m-sections and renegotiate before a subscription completes, where dual PC has the server offer on a separate transport. That is precisely what preallocating receive sections was meant to remove (#1112, currently blocked server-side).

## Usage

```bash
cd Benchmarks
LK_BENCHMARK=1 LIVEKIT_URL=wss://… LIVEKIT_API_KEY=… LIVEKIT_API_SECRET=… \
  swiftly run +xcode swift package --disable-sandbox benchmark \
  --filter "BM-SESSION-001-DualPC"
```

`--filter` needs the exact full name. Note that switching branches under `Benchmarks/` (a path dependency) can leave a stale `.build` referencing files from the other branch; `rm -rf Benchmarks/.build` clears it.

## Definitions worth confirming

These are modelled on a metric set from another SDK's harness, and three definitions are judgement calls I'd rather align than guess:

- **`peer_visible_ms`** includes the subscriber's own `connect()`. Measuring only the propagation after connect returns is also defensible and would be a much smaller number.
- **`first_frame_ms`** starts at `didSubscribeTrack` and attaches the renderer at that moment, so it includes encoder start and the first keyframe — hence ~100–200 ms rather than single digits. Starting the clock at renderer attach instead would report near-zero.
- **`disconnect_ms`** is just the call duration (~5 ms); it does not wait for server-side teardown to be observable.

Happy to re-cut any of them.

### DIFF
--- a/Benchmarks/LiveKitBenchmark/BenchmarkSupport/SessionObservers.swift
+++ b/Benchmarks/LiveKitBenchmark/BenchmarkSupport/SessionObservers.swift
@@ -1,0 +1,153 @@
+/*
+ * Copyright 2026 LiveKit
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import CoreMedia
+import CoreVideo
+import Foundation
+import LiveKit
+
+/// Monotonic milliseconds, the clock every lifecycle metric is measured on.
+func nowMs() -> Double { ProcessInfo.processInfo.systemUptime * 1000 }
+
+/// A one-shot latch that records *when* something first happened, so a benchmark can await an
+/// event and then subtract timestamps rather than wrapping each step in its own measurement.
+final class EventLatch: @unchecked Sendable {
+    private let _state = StateSync<(fired: Double?, waiters: [CheckedContinuation<Double, Error>])>((nil, []))
+
+    /// Records the first occurrence and releases anyone waiting. Later calls are ignored, so a
+    /// repeated delegate callback cannot move the mark.
+    func fire(at time: Double = nowMs()) {
+        let waiters: [CheckedContinuation<Double, Error>] = _state.mutate { state in
+            guard state.fired == nil else { return [] }
+            state.fired = time
+            defer { state.waiters.removeAll() }
+            return state.waiters
+        }
+        for waiter in waiters {
+            waiter.resume(returning: time)
+        }
+    }
+
+    /// The timestamp of the first occurrence, waiting up to `timeout` for one.
+    func wait(timeout: TimeInterval = 20) async throws -> Double {
+        if let fired = _state.read({ $0.fired }) { return fired }
+
+        return try await withThrowingTaskGroup(of: Double.self) { group in
+            group.addTask {
+                try await withCheckedThrowingContinuation { continuation in
+                    let fired: Double? = self._state.mutate { state in
+                        if let fired = state.fired { return fired }
+                        state.waiters.append(continuation)
+                        return nil
+                    }
+                    if let fired { continuation.resume(returning: fired) }
+                }
+            }
+            group.addTask {
+                try await Task.sleep(nanoseconds: UInt64(timeout * 1_000_000_000))
+                throw BenchmarkTimeout()
+            }
+            defer { group.cancelAll() }
+            return try await group.next()!
+        }
+    }
+
+    func reset() {
+        _state.mutate { $0 = (nil, []) }
+    }
+}
+
+struct BenchmarkTimeout: Error, CustomStringConvertible {
+    var description: String { "Timed out waiting for a lifecycle event" }
+}
+
+/// Marks the moment a remote participant becomes visible to an already-connected peer.
+final class PeerVisibleObserver: NSObject, RoomDelegate, @unchecked Sendable {
+    let visible = EventLatch()
+
+    func room(_: Room, participantDidConnect _: RemoteParticipant) {
+        visible.fire()
+    }
+}
+
+/// Marks the moment a published track is subscribed, and the moment its first frame renders.
+///
+/// Both are latched from the subscriber side so `pub_to_sub` and `first_frame` are read off one
+/// timeline: subscription is a signalling+negotiation event, the first frame is media actually
+/// flowing on the established transport.
+final class SubscriberObserver: NSObject, RoomDelegate, VideoRenderer, @unchecked Sendable {
+    let subscribed = EventLatch()
+    let firstFrame = EventLatch()
+
+    // Not adaptive: the benchmark must not have layers paused out from under it.
+    var isAdaptiveStreamEnabled: Bool { false }
+    var adaptiveStreamSize: CGSize { CGSize(width: 1280, height: 720) }
+
+    func room(_: Room, participant _: RemoteParticipant, didSubscribeTrack publication: RemoteTrackPublication) {
+        subscribed.fire()
+        // Attaching here rather than polling keeps the gap between the two latches to the
+        // renderer hand-off plus real frame arrival.
+        if let track = publication.track as? RemoteVideoTrack {
+            track.add(videoRenderer: self)
+        }
+    }
+
+    nonisolated func render(frame _: VideoFrame) {
+        firstFrame.fire()
+    }
+}
+
+/// Marks the arrival of an echoed data payload, for the data round trip.
+final class DataEchoObserver: NSObject, RoomDelegate, @unchecked Sendable {
+    let received = EventLatch()
+    private let topic: String
+
+    init(topic: String) {
+        self.topic = topic
+        super.init()
+    }
+
+    func room(_: Room, participant _: RemoteParticipant?, didReceiveData _: Data, forTopic topic: String, encryptionType _: EncryptionType) {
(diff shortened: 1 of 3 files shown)


xianshijing-lk:
## Correction and completed matrix

The results in the description were measured on `sxian/CLT-3305` (#1110 + #1111 + the default flip + an unlanded data channel fix), not on this branch. The description does not say so, which makes it misleading — full four-cell matrix below, all against staging, 15 iterations each.

| metric (p50 ms) | main + dual | main + single | stack + dual | stack + single |
|---|---|---|---|---|
| `connect_ms` | 561 | 554 | 556 | **489** |
| `peer_visible_ms` | 209 | 210 | 210 | 208 |
| `pub_to_sub_ms` | 220 | 311 | 222 | 308 |
| `first_frame_ms` | 193 | 102 | 192 | 103 |
| `publish_to_first_frame_ms` | 415 | 413 | 412 | 412 |
| `data_rtt_ms` | 125 | 125 | 124 | 126 |
| `disconnect_ms` | 6 | 4 | 5 | 4 |

Three things this settles that the two-column table did not:

**1. The stack does not touch dual PC.** main+dual vs stack+dual is within noise on all seven metrics (561→556, 209→210, 220→222, 193→192, 415→412, 125→124, 6→5). Every change in the stack is single-PC-gated, and this is a much broader confirmation than the single `D_TRANSPORT` number in #1111.

**2. The connect win is the stack, not single PC.** main+single is 554 — statistically the same as main+dual's 561. Only stack+single reaches 489. Single PC on its own buys nothing on connect; offer-with-join is the entire gain, and it only runs on the single-PC path.

**3. The `pub_to_sub` regression is single PC itself, not the stack.** 220 → 311 on `main` alone, and the stack neither causes nor fixes it. It is the `MediaSectionsRequirement` round trip, which is what preallocating receive sections (#1112) would remove.

Net on `publish_to_first_frame_ms`: **415 / 413 / 412 / 412** — identical across all four. Faster connect, slower subscribe, same time to first frame.

Benchmarks/LiveKitBenchmark/BenchmarkSupport/SessionObservers.swift
```diff
@@ -0,0 +1,153 @@
...
+                    if let fired { continuation.resume(returning: fired) }
+                }
+            }
+            group.addTask {
+                try await Task.sleep(nanoseconds: UInt64(timeout * 1_000_000_000))
+                throw BenchmarkTimeout()
+            }
+            defer { group.cancelAll() }
+            return try await group.next()!
+        }
```
pblazej, line 71: This is the one thing I'd want resolved before this runs on CI: `withCheckedThrowingContinuation` doesn't observe cancellation, so on a missed milestone the sleep child throws but this one stays unresumed in `_state.waiters`, and `withThrowingTaskGroup` can't propagate until every child drains. Instead of reporting which milestone was missed, the run then wedges until the workflow's 60-minute timeout. Would a `withTaskCancellationHandler` work here — on cancel, pull this continuation out of `_state.waiters` under the lock and resume it with `CancellationError()`, outside the lock as `fire()` already does?

Benchmarks/LiveKitBenchmark/SessionLifecycleBenchmark.swift
```diff
@@ -0,0 +1,191 @@
...
+                                                 options: DataPublishOptions(topic: dataTopic))
+    let echoedAt = try await dataEcho.received.wait()
+    benchmark.measurement(mDataRtt, Int(echoedAt - rttStart))
+
+    // disconnect_ms — publisher teardown only; the subscriber is closed outside the measurement.
+    let disconnectStart = nowMs()
+    await publisher.disconnect()
+    benchmark.measurement(mDisconnect, Int(nowMs() - disconnectStart))
+
+    await subscriber.disconnect()
```
pblazej: Small one, but it compounds: these two `disconnect()` calls are only on the happy path, so if any `wait()` or the `publish` throws, both Rooms stay connected (live peer connections, ADM) for the rest of the run. `defer` can't `await`, so it'd need an explicit `do { … } catch { await publisher.disconnect(); await subscriber.disconnect(); throw error }`. Probably worth it at 15 iterations x 2 variants, since one flaky milestone early on would quietly skew every `connect_ms` after it.

Benchmarks/LiveKitBenchmark/BenchmarkSupport/SessionObservers.swift
```diff
@@ -0,0 +1,153 @@
...
+            let pixelBuffer
+        else { return nil }
+
+        CVPixelBufferLockBaseAddress(pixelBuffer, [])
+        defer { CVPixelBufferUnlockBaseAddress(pixelBuffer, []) }
+        // Mid-grey luma and neutral chroma: encodes cheaply and stays a valid frame.
+        for plane in 0 ..< CVPixelBufferGetPlaneCount(pixelBuffer) {
+            guard let base = CVPixelBufferGetBaseAddressOfPlane(pixelBuffer, plane) else { continue }
+            let bytes = CVPixelBufferGetBytesPerRowOfPlane(pixelBuffer, plane) * CVPixelBufferGetHeightOfPlane(pixelBuffer, plane)
+            memset(base, plane == 0 ? 128 : 128, bytes)
```
pblazej: Tiny thing: both branches of the ternary are `128`, so this is effectively `memset(base, 128, bytes)`. The comment above says "mid-grey luma and neutral chroma", which reads like two values were intended and one got lost. If 128 is right for both planes (it is, for `420YpCbCr8BiPlanarFullRange`), collapsing it would save the next reader a double-take.

pblazej (commented):
This fills a real gap: `BM-CONN` can't see the cost that lands after `connect()` returns, and `publish_to_first_frame_ms` being flat at 412 in both modes is the most useful row in the whole connect-latency series. I'd like this to land ahead of #1111/#1113 for that reason — it's the only thing that can say whether the rest of the stack is worth shipping.

Benchmarks/LiveKitBenchmark/BenchmarkSupport/SessionObservers.swift
pblazej: Fixed in f744d7e1. Went with `withTaskCancellationHandler` as suggested, plus keyed waiters so cancellation settles its own continuation — and a `cancelled` state for the race where the handler runs before the body registers, which would otherwise park a continuation nobody holds. Verified by temporarily awaiting a latch nobody fires: the run now reports `failed with Timed out waiting for a lifecycle event` in ~2s instead of hanging.

pblazej:
Pushed `f744d7e1` addressing the three review notes — two of my inline comments seem to have been deleted, so recording the outcome here.

**`EventLatch.wait` hung instead of timing out** (also Devin's first finding). Waiters are now keyed in a `[UInt64: Waiter]` so cancellation can settle its own continuation, and the suspend sits inside `withTaskCancellationHandler`. There is a `cancelled` state for the race Devin's prompt didn't cover: the handler can fire before the operation body registers, and without it the body would park a continuation nobody holds. Resumes stay outside the lock, as `fire()` already did. `reset()` now settles live waiters too rather than dropping them.

**Rooms leaked on the error path** (also Devin's second finding). The measured body is wrapped in `do`/`catch` with an explicit teardown, since `defer` cannot `await`. Per Devin's constraint the error path emits no `disconnect_ms`, so that metric still only reports a successful publisher teardown.

**Dead ternary** — `memset(base, plane == 0 ? 128 : 128, bytes)` collapsed to `memset(base, 128, bytes)`.

Verification: the timeout path was checked by temporarily awaiting a latch nobody fires — the run reports `failed with Timed out waiting for a lifecycle event` in ~2s where it previously hung. Happy path against a local `livekit-server` after the change: all 7 metrics, 15 samples, no failures.

One observation while running it locally: `peer_visible_ms` came out at 2885 ms p50, against the 210 ms you measured on Cloud staging. I didn't chase it and it may just be the local server, but given it's the one definition you flagged as a judgement call it might be worth a look.

pblazej (approved):

PR #1116, opened 2026-09-08
Wait for the data channel before every send

xianshijing-lk:
Two independent defects that presented as one symptom: `PeerConnectionSignalingTests.dataChannel` timing out, which is what blocks the single-PC default flip in #1113.

## 1. Data sends were not gated on the channel being open (`49cc465d`)

`ensurePublisherConnected()` opened with a guard that returned early for any transport mode other than `.subscriberPrimary`:

```swift
// Only needed when subscriber is primary in dual PC mode
guard case .subscriberPrimary = _state.transport else { return }
```

So in single peer connection and publisher-primary modes the data channel open wait never ran. `connect()` returns once the *primary* transport is connected; the channels open afterwards on the SCTP association, so a send issued right after connect races them. Under `.dropOldest` a lossy write parked through that window resolves its waiter even when evicted, so the send reported success for a packet that never shipped.

The guard conflated two questions. Only the on-demand *negotiation* is specific to subscriber-primary; the channel wait applies to every mode. This splits them — the wait runs unconditionally, negotiation stays gated as before.

Not caused by the single-PC work: the guard read `guard _state.isSubscriberPrimary else { return }` before #919, which only reshaped it for the `TransportMode` enum. `.publisherPrimary` dual PC has always had the same hole.

### How rust-sdks handles it, and how this aligns

rust has never had this bug and needed no fix for it. `ensure_publisher_connected(kind)` (`rtc_session.rs:2415`) resolves the channel for the requested kind on every call and polls until it reports open, with **no transport-mode branch anywhere on the path**:

```rust
let required_dc = self.data_channel(SignalTarget::Publisher, kind).unwrap();
if required_dc.state() == DataChannelState::Open { return Ok(()) }
while !self.publisher_pc.is_connected() || required_dc.state() != DataChannelState::Open {
    tokio::time::sleep(Duration::from_millis(50)).await;
}
```

The revision immediately before rust added single-PC support (`8b3f35a7^`) already had this function unconditional and per-kind — adding single PC did not require touching it. rust also keeps the two concerns separate the way this PR now does: negotiation gated on `has_published`, channel wait unconditional.

| | rust-sdks | Swift (this PR) |
|---|---|---|
| gate runs in all transport modes | yes | yes |
| gate is per channel kind | `data_channel(Publisher, kind)` | `openCompleter(for: kind)` |
| kind comes from the packet | `publish_data(kind)` | `send(dataPacket:)` passes `packet.kind` |
| negotiation decoupled from the wait | `has_published` | `.subscriberPrimary` branch only |

Per-kind rather than pair-wide because reliable and lossy are independent SCTP streams — waiting on both would let a lagging reliable channel stall or fail a send the lossy channel was ready to take. This is not academic here: the failing test publishes **lossy**, so a pair-wide gate would have been waiting on the wrong channel.

**client-sdk-js is not a good reference for this.** It gates from the same place (`sendDataPacket` → `ensurePublisherConnected(kind)`, `RTCEngine.ts:1574`) but memoizes a single connection-scoped `publisherConnectionPromise` rather than one per kind, cleared only when the publisher PC goes closed/disconnected/failed. The first kind to call wins for the life of the connection, so a later send of the other kind awaits an already-resolved promise without its own channel ever being checked — the exact case this gate prevents. The docstring records this so the next reader doesn't "align" Swift back to it.

Two deliberate divergences from rust, flagged for review: the non-subscriber-primary branch waits on the channel opening but not additionally on the PC being connected (an open SCTP channel should already imply DTLS and ICE are up), and it doesn't kick negotiation from inside the gate, since the publisher is already negotiated at connect in those modes.

## 2. The test asserted delivery on a best-effort channel (`40d13d7b`)

`dataChannel` published with a default-constructed `DataPublishOptions`, whose `reliable` defaults to `false`, and then asserted the payload arrived. A lossy channel does not retransmit and does not guarantee delivery, so the test asserted a guarantee the transport does not make.

This — not the send race — is what actually made the test time out, and it was never single-PC specific. Dual PC failed too.

## Verification

All against staging Cloud (`1.13.6`).

**Stress, 12 runs of `dataChannel` per configuration:**

| configuration | pass rate |
|---|---|
| this branch, test as written (lossy) | **5 / 12** |
| this branch, test reliable | **12 / 12** |

**The send fix holds independently.** `send(dataPacket:)` logs a diagnostic when it writes to a channel that is not open. Across all 24 stress runs on this branch it fired **0 times**; on plain `main` it fires. So the packet is no longer handed to a closed channel, regardless of the test's reliability setting.

**Full suite:** `PeerConnectionSignalingTests` 9/9 passing, both modes, 52s.

**Diagnosis method**, in case it matters for review: temporary probes on the send path, on `DataChannelPair.handle(received:)` before any decode, and on server-opened channel routing. On every failure the pattern was identical — send issued with the channel open, and the receiving room logging *no inbound data channel message at all*. Loss in transit, not a decode, routing or dispatch problem. The probes are not part of this PR.

Not verified: `swiftlint` is not installed on this machine, so lint is unrun — CI is the first check.

## One thing worth a separate conversation

A ~58% loss rate on a single lossy datagram sent right after connect is far above ordinary network loss. The most likely explanation is that the receiving side's data path is not warm at that instant and lossy has no queue to cover the gap. That is inherent to a best-effort channel and fine for applications that stream lossy data continuously, but it does mean "send one lossy packet immediately after connect and expect it to arrive" is unreliable by construction. Flagging rather than burying it under a now-green test.

### DIFF
--- a/Sources/LiveKit/Core/DataChannelPair.swift
+++ b/Sources/LiveKit/Core/DataChannelPair.swift
@@ -54,6 +54,21 @@ class DataChannelPair: NSObject, @unchecked Sendable, Loggable {
 
     let openCompleter = AsyncCompleter<Void>(label: "Data channel open", defaultTimeout: .defaultPublisherDataChannelOpen)
 
+    /// Per-channel open latches, for callers that will use one channel rather than the pair.
+    ///
+    /// The two are independent SCTP streams and open independently, so gating a lossy send on the
+    /// reliable channel would stall — or fail — a send that its own channel was ready to take.
+    /// Both references gate per kind for exactly this reason: `ensure_publisher_connected(kind)` →
+    /// `data_channel(Publisher, kind)` in rust-sdks, and `ensureDataTransportConnected(kind)` →
+    /// `dataChannelForKind(kind)` in client-sdk-js.
+    private let reliableOpenCompleter = AsyncCompleter<Void>(label: "Reliable data channel open", defaultTimeout: .defaultPublisherDataChannelOpen)
+    private let lossyOpenCompleter = AsyncCompleter<Void>(label: "Lossy data channel open", defaultTimeout: .defaultPublisherDataChannelOpen)
+
+    /// The open latch for the channel a packet of `kind` will be sent on.
+    func openCompleter(for kind: Livekit_DataPacket_Kind) -> AsyncCompleter<Void> {
+        kind == .lossy ? lossyOpenCompleter : reliableOpenCompleter
+    }
+
     /// Whether *both* channels can currently take bytes. Only the open latch and diagnostics use
     /// this; the send path gates per channel.
     var isOpen: Bool { lossy.isOpen && reliable.isOpen }
@@ -131,6 +146,8 @@ class DataChannelPair: NSObject, @unchecked Sendable, Loggable {
     /// Resolves the open latch once both channels are usable. Reached from either drain's state
     /// callback and from a channel swap.
     private func handleStateChange() {
+        if reliable.isOpen { reliableOpenCompleter.resume(returning: ()) }
+        if lossy.isOpen { lossyOpenCompleter.resume(returning: ()) }
         if isOpen {
             openCompleter.resume(returning: ())
         }
@@ -154,6 +171,8 @@ class DataChannelPair: NSObject, @unchecked Sendable, Loggable {
         // Negotiated per session (from the SDP answer); the next session must not inherit it.
         set(maxMessageSize: Self.defaultMaxMessageSize)
         openCompleter.reset(throwing: error)
+        reliableOpenCompleter.reset(throwing: error)
+        lossyOpenCompleter.reset(throwing: error)
     }
 
     // MARK: - Send
--- a/Sources/LiveKit/Core/DataChannelPair.swift
+++ b/Sources/LiveKit/Core/DataChannelPair.swift
@@ -58,9 +58,11 @@ class DataChannelPair: NSObject, @unchecked Sendable, Loggable {
     ///
     /// The two are independent SCTP streams and open independently, so gating a lossy send on the
     /// reliable channel would stall — or fail — a send that its own channel was ready to take.
-    /// Both references gate per kind for exactly this reason: `ensure_publisher_connected(kind)` →
-    /// `data_channel(Publisher, kind)` in rust-sdks, and `ensureDataTransportConnected(kind)` →
-    /// `dataChannelForKind(kind)` in client-sdk-js.
+    /// rust-sdks resolves the channel per kind on every send (`ensure_publisher_connected(kind)` →
+    /// `data_channel(Publisher, kind)`). client-sdk-js resolves per kind one level down
+    /// (`ensureDataTransportConnected(kind)` → `dataChannelForKind(kind)`) but memoizes a single
+    /// connection-scoped promise above it, so in practice only its first send is gated per kind —
+    /// see the note on `Room.ensureDataChannelReady(kind:)`.
     private let reliableOpenCompleter = AsyncCompleter<Void>(label: "Reliable data channel open", defaultTimeout: .defaultPublisherDataChannelOpen)
     private let lossyOpenCompleter = AsyncCompleter<Void>(label: "Lossy data channel open", defaultTimeout: .defaultPublisherDataChannelOpen)
 
@@ -69,8 +71,13 @@ class DataChannelPair: NSObject, @unchecked Sendable, Loggable {
         kind == .lossy ? lossyOpenCompleter : reliableOpenCompleter
     }
 
-    /// Whether *both* channels can currently take bytes. Only the open latch and diagnostics use
-    /// this; the send path gates per channel.
+    /// Whether the channel a packet of `kind` would be written to can currently take bytes.
+    func isOpen(kind: Livekit_DataPacket_Kind) -> Bool {
+        kind == .lossy ? lossy.isOpen : reliable.isOpen
+    }
+
+    /// Whether *both* channels can currently take bytes. Only the pair-wide latch and diagnostics
+    /// use this; the send path gates per channel.
     var isOpen: Bool { lossy.isOpen && reliable.isOpen }
 
     // MARK: - Private
@@ -146,10 +153,18 @@ class DataChannelPair: NSObject, @unchecked Sendable, Loggable {
     /// Resolves the open latch once both channels are usable. Reached from either drain's state
     /// callback and from a channel swap.
     private func handleStateChange() {
-        if reliable.isOpen { reliableOpenCompleter.resume(returning: ()) }
-        if lossy.isOpen { lossyOpenCompleter.resume(returning: ()) }
+        // Re-armed, not just resolved: libwebrtc closes an SCTP channel itself on a
+        // max-message-size violation, and a latch that only ever resolves would let the next send
+        // through instantly to park in the drain — the same silent drop the gate exists to stop.
+        // `rearm()` leaves in-flight waiters waiting rather than failing them, so a channel that
+        // flaps does not fail a send that the reopened channel can take. `DataTracks` re-arms its
+        // own publisher latch the same way.
+        if reliable.isOpen { reliableOpenCompleter.resume(returning: ()) } else { reliableOpenCompleter.rearm() }
+        if lossy.isOpen { lossyOpenCompleter.resume(returning: ()) } else { lossyOpenCompleter.rearm() }
         if isOpen {
             openCompleter.resume(returning: ())
+        } else {
+            openCompleter.rearm()
         }
     }
 
--- a/Sources/LiveKit/Core/DataChannelPair.swift
+++ b/Sources/LiveKit/Core/DataChannelPair.swift
@@ -159,13 +159,22 @@ class DataChannelPair: NSObject, @unchecked Sendable, Loggable {
         // `rearm()` leaves in-flight waiters waiting rather than failing them, so a channel that
         // flaps does not fail a send that the reopened channel can take. `DataTracks` re-arms its
         // own publisher latch the same way.
-        if reliable.isOpen { reliableOpenCompleter.resume(returning: ()) } else { reliableOpenCompleter.rearm() }
-        if lossy.isOpen { lossyOpenCompleter.resume(returning: ()) } else { lossyOpenCompleter.rearm() }
-        if isOpen {
-            openCompleter.resume(returning: ())
-        } else {
-            openCompleter.rearm()
+        //
+        // Serialized on the existing state lock, and each channel's state is read *inside* it.
+        // `onStateChange` fires both from `setChannel` and from WebRTC's delegate thread, so
+        // sampling `isOpen` outside would let a stale invocation resolve a latch that a newer
+        // close had already re-armed — leaving the next send to skip the gate and park in a
+        // closed drain.
+        _state.mutate { _ in
+            sync(reliableOpenCompleter, isOpen: reliable.isOpen)
+            sync(lossyOpenCompleter, isOpen: lossy.isOpen)
+            sync(openCompleter, isOpen: isOpen)
         }
+    }
+
+    /// Points a latch at `isOpen`, sampled by the caller under the state lock.
+    private func sync(_ completer: AsyncCompleter<Void>, isOpen: Bool) {
+        if isOpen { completer.resume(returning: ()) } else { completer.rearm() }
     }
 
     /// Update the negotiated SCTP max-message-size cap on both channels. Called by the room after
--- a/Sources/LiveKit/Core/Room+Engine.swift
+++ b/Sources/LiveKit/Core/Room+Engine.swift
@@ -90,41 +90,62 @@ extension Room {
     ///
     /// - Parameter kind: Gates on this kind's channel alone. The two are independent SCTP streams,
     ///   so waiting on the pair would let a lagging reliable channel fail a lossy send.
-    func ensurePublisherConnected(kind: Livekit_DataPacket_Kind = .reliable) async throws {
-        // On-demand negotiation is only needed when subscriber is primary in dual PC mode; the
-        // other modes negotiate the publisher during connect.
-        guard case .subscriberPrimary = _state.transport else {
-            try await publisherDataChannel.openCompleter(for: kind).wait()
+    func ensureDataChannelReady(kind: Livekit_DataPacket_Kind) async throws {
+        // Concurrently, not in sequence: in subscriber-primary mode the channel opens on the SCTP
+        // association the transport is still bringing up, so these two complete together.
+        async let transportReady: Void = ensurePublisherConnected()
+        async let channelReady: Void = publisherDataChannel.openCompleter(for: kind).wait()
+        _ = try await (transportReady, channelReady)
+    }
+
+    /// Negotiates the publisher peer connection on demand and waits for it to connect.
+    ///
+    /// Transport only — it does not gate on any data channel. Callers that are about to write a
+    /// packet want ``ensureDataChannelReady(kind:)``; callers that only need the peer connection
+    /// (the data-track publish path, which gates on its own channel) want this.
+    ///
+    /// Explicit over the whole enum rather than `guard case .subscriberPrimary`: `_state.transport`
+    /// is `nil` between `cleanUpRTC` and the next JOIN, and a negated pattern-match folds that case
+    /// in with the publisher-primary modes, which need opposite handling.
+    func ensurePublisherConnected() async throws {
+        switch _state.transport {
+        case .subscriberPrimary:
+            // The only mode that defers publisher negotiation until something is published.
+            let publisher = try requirePublisher()
+
+            let connectionState = await publisher.connectionState
+            if connectionState != .connected, connectionState != .connecting {
+                try await publisherShouldNegotiate()
+            }
+
+            try await publisherTransportConnectedCompleter.wait(timeout: _state.connectOptions.publisherTransportConnectTimeout)
+
+        case .publisherOnly, .publisherPrimary:
+            // The publisher is the primary transport, so `connect()` already waited for it.
+            return
+
+        case nil:
+            // No transport to negotiate — between `cleanUpRTC` and the next JOIN. Returning keeps
+            // the reconnect window a no-op, as it was before the channel gate existed.
             return
         }
-
-        let publisher = try requirePublisher()
-
-        let connectionState = await publisher.connectionState
-        if connectionState != .connected, connectionState != .connecting {
-            try await publisherShouldNegotiate()
-        }
-
-        // Single combined gate: wait for both the publisher PC to be ICE-
-        // connected *and* the channel to reach `.open` concurrently.
-        async let transportReady: Void = publisherTransportConnectedCompleter.wait(timeout: _state.connectOptions.publisherTransportConnectTimeout)
-        async let dataChannelReady: Void = publisherDataChannel.openCompleter(for: kind).wait()
-        _ = try await (transportReady, dataChannelReady)
     }
 
     func send(dataPacket packet: consuming Livekit_DataPacket) async throws {
         // The same field `DataChannelPair.send` routes on, so the channel awaited here is the one
         // the packet is handed to.
-        try await ensurePublisherConnected(kind: packet.kind)
+        let kind = packet.kind
+        try await ensureDataChannelReady(kind: kind)
 
         // At this point publisher should be .connected and dc should be .open
         if await !(_state.transport?.publisher.isConnected ?? false) {
             log("publisher is not .connected", .error)
         }
 
-        let dataChannelIsOpen = publisherDataChannel.isOpen
-        if !dataChannelIsOpen {
-            log("publisher data channel is not .open", .error)
+        // Per kind, matching the gate above: the pair-wide `isOpen` would report a lagging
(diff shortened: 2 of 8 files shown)


Sources/LiveKit/Core/Room+Engine.swift
```diff
@@ -71,10 +71,30 @@ extension Room {
...
+    /// than one per kind, so after the first call resolves a send of the other kind proceeds
+    /// without its own channel having been checked — the case this gate exists to prevent.
+    ///
+    /// - Parameter kind: Gates on this kind's channel alone. The two are independent SCTP streams,
+    ///   so waiting on the pair would let a lagging reliable channel fail a lossy send.
+    func ensurePublisherConnected(kind: Livekit_DataPacket_Kind = .reliable) async throws {
+        // On-demand negotiation is only needed when subscriber is primary in dual PC mode; the
+        // other modes negotiate the publisher during connect.
         guard case .subscriberPrimary = _state.transport else {
+            try await publisherDataChannel.openCompleter(for: kind).wait()
```
pblazej, line 97: This wait fires in the mode the test harness actually runs in (`RoomTestingOptions` defaults `canSubscribe: false`, so the server reports `subscriber_primary: false`), and `MockDataChannelPair` never calls `set(reliable:)`/`set(lossy:)`, so the latch never resolves — measured against a local server, `RoomTests.sendDataPacket` goes 1.13 s pass on `main` to 15.17 s fail here, `publishDtmfSendsSipDtmfPacket` the same, and 12 of 14 `RpcClientTests` fail on the 15 s timeout.

Sources/LiveKit/Core/Room+Engine.swift
```diff
@@ -71,10 +71,30 @@ extension Room {
...
+    /// `RTCEngine.ts:1574`) but memoizes one connection-scoped `publisherConnectionPromise` rather
+    /// than one per kind, so after the first call resolves a send of the other kind proceeds
+    /// without its own channel having been checked — the case this gate exists to prevent.
+    ///
+    /// - Parameter kind: Gates on this kind's channel alone. The two are independent SCTP streams,
+    ///   so waiting on the pair would let a lagging reliable channel fail a lossy send.
+    func ensurePublisherConnected(kind: Livekit_DataPacket_Kind = .reliable) async throws {
+        // On-demand negotiation is only needed when subscriber is primary in dual PC mode; the
+        // other modes negotiate the publisher during connect.
         guard case .subscriberPrimary = _state.transport else {
```
pblazej, line 96: `_state.transport` is optional and `cleanUpRTC` sets it to nil, so this pattern-match also fails for nil in subscriber-primary mode — a `publish(data:)` during the full-reconnect window then takes the `else` branch, skips the `publisherShouldNegotiate()` it needs, and waits out the full timeout on a channel that cannot open; an explicit `switch` over the three modes plus nil would keep those cases distinct.

Sources/LiveKit/Core/Room+Engine.swift
```diff
@@ -71,10 +71,30 @@ extension Room {
...
+    /// (`rtc_session.rs:2415`) resolves `data_channel(Publisher, kind)` on every call and polls
+    /// until that channel reports open, with no transport-mode branch anywhere on the path.
+    /// client-sdk-js gates from the same place (`sendDataPacket` → `ensurePublisherConnected(kind)`,
+    /// `RTCEngine.ts:1574`) but memoizes one connection-scoped `publisherConnectionPromise` rather
+    /// than one per kind, so after the first call resolves a send of the other kind proceeds
+    /// without its own channel having been checked — the case this gate exists to prevent.
+    ///
+    /// - Parameter kind: Gates on this kind's channel alone. The two are independent SCTP streams,
+    ///   so waiting on the pair would let a lagging reliable channel fail a lossy send.
+    func ensurePublisherConnected(kind: Livekit_DataPacket_Kind = .reliable) async throws {
```
pblazej, line 93: The `= .reliable` default silently gives the two no-argument callers in `Room+DataTrack.swift` (:113 and :207) a dependency on the reliable channel they never write to — data tracks use `Labels.dataTrack` and already gate on their own `_publisherChannelOpen` — so a slow reliable channel now makes `publish(name:)` throw a timeout naming the data-track channel; requiring the kind at each call site would keep the coupling visible.

Sources/LiveKit/Core/Room+Engine.swift
```diff
@@ -86,16 +106,16 @@ extension Room {
...
-        async let dataChannelReady: Void = publisherDataChannel.openCompleter.wait()
+        async let dataChannelReady: Void = publisherDataChannel.openCompleter(for: kind).wait()
         _ = try await (transportReady, dataChannelReady)
     }
 
     func send(dataPacket packet: consuming Livekit_DataPacket) async throws {
-        try await ensurePublisherConnected()
+        // The same field `DataChannelPair.send` routes on, so the channel awaited here is the one
+        // the packet is handed to.
+        try await ensurePublisherConnected(kind: packet.kind)
```
pblazej, line 118: The diagnostic a few lines below still tests the pair-wide `publisherDataChannel.isOpen` (`lossy && reliable`) while the gate now waits on one kind, so a successful lossy send logs "publisher data channel is not .open" at `.error` while reliable is still opening — which also means the "diagnostic fired 0 times across 24 runs" evidence no longer measures the channel the gate waited on.

Sources/LiveKit/Core/DataChannelPair.swift
```diff
@@ -131,6 +146,8 @@ class DataChannelPair: NSObject, @unchecked Sendable, Loggable {
     /// Resolves the open latch once both channels are usable. Reached from either drain's state
     /// callback and from a channel swap.
     private func handleStateChange() {
+        if reliable.isOpen { reliableOpenCompleter.resume(returning: ()) }
```
pblazej, line 149: These latches only ever resolve and are never re-armed when a channel leaves `.open` — libwebrtc closes the SCTP channel itself on a max-message-size violation, after which the new gate returns instantly and the write parks in the drain with no timeout, which is the same class of silent drop this PR is fixing; `DataTracks.handlePublisherStateChange` already does the right thing with `_publisherChannelOpen.rearm()`.

Sources/LiveKit/Core/DataChannelPair.swift
```diff
@@ -54,6 +54,21 @@ class DataChannelPair: NSObject, @unchecked Sendable, Loggable {
 
     let openCompleter = AsyncCompleter<Void>(label: "Data channel open", defaultTimeout: .defaultPublisherDataChannelOpen)
 
+    /// Per-channel open latches, for callers that will use one channel rather than the pair.
+    ///
+    /// The two are independent SCTP streams and open independently, so gating a lossy send on the
+    /// reliable channel would stall — or fail — a send that its own channel was ready to take.
+    /// Both references gate per kind for exactly this reason: `ensure_publisher_connected(kind)` →
+    /// `data_channel(Publisher, kind)` in rust-sdks, and `ensureDataTransportConnected(kind)` →
```
pblazej, line 62: This says client-sdk-js gates per kind, but the new docstring on `ensurePublisherConnected` says the opposite — that JS memoizes one connection-scoped promise "rather than one per kind" and is "the case this gate exists to prevent" — so one of the two needs correcting before a future reader uses it to justify collapsing the split.

Tests/LiveKitCoreTests/PeerConnectionSignalingTests.swift
```diff
@@ -297,7 +297,7 @@ struct PeerConnectionSignalingTests {
 
             let room2Watcher: RoomWatcher<TestPayload> = room2.createWatcher()
 
-            try await room1.localParticipant.publish(data: jsonData, options: DataPublishOptions(topic: topic))
+            try await room1.localParticipant.publish(data: jsonData, options: DataPublishOptions(topic: topic, reliable: true))
```
pblazej, line 300: This was the only test exercising the lossy gate, so switching it to `reliable: true` leaves `openCompleter(for: .lossy)` with no coverage at all — a unit assertion in `DataChannelPairTests` that the two completers are distinct instances and that `reset(throwing:)` clears both would cover it without depending on best-effort delivery.

Sources/LiveKit/Core/Room+Engine.swift
xianshijing-lk: Good catch. thanks.

I verified each one rather than take the measurement. Fixed them by MockDataChannelPair now reports itself open (pre-resolved latch plus isOpen), which is what it stands in for,  it replaces the whole pair and nothing ever drives handleStateChange.

Back to baseline on staging: sendDataPacket() 1.825 s, publishDtmfSendsSipDtmfPacket() 1.587 s, full RpcClientTests green. 38/38 across PeerConnectionSignalingTests, RpcClientTests, RoomTests, DataChannelPairTests.

This also explains the 30 minute Build & Test timeouts on this PR, which I had wrongly written off as infrastructure.

Sources/LiveKit/Core/Room+Engine.swift
xianshijing-lk: Fixed. now an explicit switch over .subscriberPrimary / .publisherOnly, .publisherPrimary / nil, as you suggested. 

nil returns without waiting, keeping the reconnect window the no-op it was before the gate existed.

Sources/LiveKit/Core/Room+Engine.swift
xianshijing-lk: Fixed. The new code now splits the two concerns rather than just remove the default.

ensurePublisherConnected() is transport only again, so both data track callers get their previous behaviour back,  including the no-op at :205 that the comment there relies on. send(dataPacket:) calls a new ensureDataChannelReady(kind:) with no default argument, which awaits transport and channel concurrently.

Sources/LiveKit/Core/Room+Engine.swift
xianshijing-lk: Fixed  by adding isOpen(kind:), and the diagnostic now names the channel the gate actually waited on.

One narrow disagreement on the second half: the diagnostic fires when either channel is closed, so zero occurrences implies the gated channel was open,  the inference still holds, it was just a weaker instrument than it should have been. The false positive on lossy sends was real and is gone.

Sources/LiveKit/Core/DataChannelPair.swift
xianshijing-lk: Good catch. Now it should get fixed.  handleStateChange now re-arms each latch when its channel is not open, matching DataTracks.handlePublisherStateChange.

Used rearm() rather than reset(throwing:) deliberately: it clears the cached result without failing in flight waiters, so a channel that flaps doesn't fail a send the reopened channel can take.

Sources/LiveKit/Core/DataChannelPair.swift
xianshijing-lk: Both were true at different layers, which is exactly the problem. Reconciled in latest patch: rust gates per kind on every send; JS gates per kind one level down (ensureDataTransportConnected(kind) → dataChannelForKind(kind)) but memoizes a single connection-scoped promise above it, so in practice only its first send is gated per kind. The note now says that in one place and points at the other.

Tests/LiveKitCoreTests/PeerConnectionSignalingTests.swift
xianshijing-lk: Added in latest patchset, as suggested,  openLatchesAreDistinctPerKind (distinct and stable per kind) and resetFailsWaitersOnBothLatches in DataChannelPairTests. Both pass, and neither depends on best effort delivery.

Sources/LiveKit/Core/Room+Engine.swift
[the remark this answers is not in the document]
xianshijing-lk: it should be resolved now.

Sources/LiveKit/Core/Room+Engine.swift
```diff
@@ -71,40 +71,81 @@ extension Room {
...
-        // Single combined gate: wait for both the publisher PC to be ICE-
-        // connected *and* the data channels to reach `.open` concurrently.
-        // Mirrors the prevailing pattern in client-sdk-js / -rust, where a
-        // single poll loop checks both conditions before any send proceeds.
-        async let transportReady: Void = publisherTransportConnectedCompleter.wait(timeout: _state.connectOptions.publisherTransportConnectTimeout)
-        async let dataChannelReady: Void = publisherDataChannel.openCompleter.wait()
-        _ = try await (transportReady, dataChannelReady)
+        case nil:
+            // No transport to negotiate — between `cleanUpRTC` and the next JOIN. Returning keeps
+            // the reconnect window a no-op, as it was before the channel gate existed.
```
pblazej, line 129: I may be misreading this, but I think it's no longer quite a no-op: `ensurePublisherConnected()` does return here immediately, yet `ensureDataChannelReady` still awaits `openCompleter(for: kind)`, which has no cached result after a `reset` — so a `publish(data:)` in the `cleanUpRTC`→JOIN window looks like it would wait the full `.defaultPublisherDataChannelOpen` (15 s) before throwing, where it used to fail right away. Would failing fast on the `nil` case make sense here? Worth noting that ceiling also isn't exposed on `ConnectOptions` the way `publisherTransportConnectTimeout` is, so callers can't shorten it themselves.

Sources/LiveKit/Core/DataChannelPair.swift
```diff
@@ -131,11 +153,30 @@ class DataChannelPair: NSObject, @unchecked Sendable, Loggable {
     /// Resolves the open latch once both channels are usable. Reached from either drain's state
     /// callback and from a channel swap.
     private func handleStateChange() {
-        if isOpen {
-            openCompleter.resume(returning: ())
+        // Re-armed, not just resolved: libwebrtc closes an SCTP channel itself on a
```
pblazej, line 156: Re-arming makes a lot of sense for a channel that flaps. The one case I'm unsure about is the max-message-size close this comment cites — as I understand it that one is terminal, and the channel won't report `.open` again without renegotiation, so later sends on that kind would wait 15 s and throw `.timedOut` rather than failing immediately with `sendData failed`. Would it be worth separating a transient close from a terminal one before re-arming, so the actionable error survives?

Tests/LiveKitCoreTests/PeerConnectionSignalingTests.swift
pblazej, line 300: `PublishDataTests._publishDataTest` makes the same assertion on the same default: 100 concurrent `publish(data:)` with a default-constructed `DataPublishOptions`, then a 15 s wait for every topic, so a single evicted datagram fails it — that is the `Timed out` at `PublishDataTests.swift:27` / `:32` in 4 of the last 5 `Build & Test` runs on #1124 (one leg each, 18.4 s, both the `canSubscribe: true` and `false` variants), and the SFU side of those rooms is indistinguishable from a passing one (both PCs connected, all three channels open ~30 ms after join, no renegotiation). Since part 2 already makes this call for `dataChannel`, would it make sense to switch that test to `reliable: true` here too, so the two don't diverge on what a lossy assertion means?

Sources/LiveKit/Support/Async/AsyncCompleter.swift
```diff
@@ -189,6 +189,15 @@ final class AsyncCompleter<T: Sendable>: @unchecked Sendable, Loggable {
                 }
 
                 _lock.sync {
+                    // Re-checked here, under the same lock that registers the waiter. The read
+                    // above is only a fast path: between it and this block a `resume` can land,
+                    // cache its result and find no waiter to hand it to, stranding this
+                    // continuation until it times out on an already-resolved completer.
+                    if let result = _result {
+                        continuation.resume(with: result)
```
pblazej, line 197: This closes a real gap — a waiter registered just after a `resume` otherwise sits out the full timeout on an already-resolved completer — but it resumes the continuation while still inside `_lock.sync`, which is the shape that wedged `waitUntillAnyActiveTimeout` on CI: `swift_task_cancel` runs `withTaskCancellationHandler`'s `onCancel` under the task's status-record lock, `onCancel` takes `_lock`, and a resume issued under `_lock` needs that same status-record lock (`WaitEntry.timeout → swift_continuation_throwingResume → waitForStatusRecordUnlock` against `swift_task_cancel → onCancel → _lock` in the stall dumps on #1124). `13058e32` on that branch moves every resume outside the lock; when this rebases, would returning the cached result out of the `sync` block and resuming after it work for you? It keeps the fast path without reintroducing the ordering.

pblazej: Done in 28fb52d9 — and this turned out to be load-bearing rather than optional: merging main auto-merged the two shapes with no conflict, leaving `timeout` and `onCancel` outside the lock and the re-check inside it, so the inversion #1124 removed came back silently.

Tests/LiveKitCoreTests/PeerConnectionSignalingTests.swift
pblazej: Done in b3d400a2 — switched `_publishDataTest` to `reliable: true` so the two tests agree on what a lossy assertion means.

Sources/LiveKit/Core/DataChannelPair.swift
pblazej: Withdrawing this one — I prototyped the split and measured it: `wasReset` is false on ordinary reconnect teardown too, so a "terminal" branch fires ~50x per run on healthy reconnects and caches a failure on the latch, which broke `reliableDelivery(dual reconnect)` under CPU contention (5/128 packets). Your `rearm()` is the right call; sorry for the noise.

pblazej:
@xianshijing-lk I'm trying to fix the remaining failures now 💣 

Moved to draft as I revisit the root cause(s)

Sources/LiveKit/Core/DataChannelPair.swift
```diff
@@ -131,11 +153,30 @@ class DataChannelPair: NSObject, @unchecked Sendable, Loggable {
...
+        //
+        // Serialized on the existing state lock, and each channel's state is read *inside* it.
+        // `onStateChange` fires both from `setChannel` and from WebRTC's delegate thread, so
+        // sampling `isOpen` outside would let a stale invocation resolve a latch that a newer
+        // close had already re-armed — leaving the next send to skip the gate and park in a
+        // closed drain.
+        _state.mutate { _ in
+            sync(reliableOpenCompleter, isOpen: reliable.isOpen)
+            sync(lossyOpenCompleter, isOpen: lossy.isOpen)
+            sync(openCompleter, isOpen: isOpen)
```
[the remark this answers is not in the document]
pblazej: Fixed in b62064ee — `reset(throwing:)` now clears the drains and all three completers inside the same `_state.mutate` that `handleStateChange` samples them under, so the two cannot interleave. Holding the lock across those resets is safe here: the drain detaches its delegate before closing, so nothing re-enters `handleStateChange`, and neither the drain nor the completers settle continuations under their own locks.

Sources/LiveKit/Core/Room+Engine.swift
```diff
@@ -71,40 +71,92 @@ extension Room {
...
+    func ensureDataChannelReady(kind: Livekit_DataPacket_Kind) async throws {
+        // Between `cleanUpRTC` and the next JOIN there is no transport to open a channel, so the
+        // latch below has nothing that could resolve it and the wait would only burn its full
+        // timeout before failing. `ensurePublisherConnected` returns immediately in this state, so
+        // this is the only place the window is visible. Racy by nature — the transport can go away
+        // right after the check — but a waiter already registered when it does is failed by
+        // `DataChannelPair.reset(throwing:)`, so the remaining window is one hop wide.
+        guard _state.transport != nil else {
+            throw LiveKitError(.invalidState, message: "Room is not connected")
         }
```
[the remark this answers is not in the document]
pblazej: Good catch, and it also caught a second bug in my own guard. Fixed in 394b18aa by gating on connection state instead of transport presence: `disconnect()` moves to `.disconnecting` before any of the teardown, which closes the whole window you describe. It also drops a regression the transport check would have introduced — a full reconnect clears the transport too, and a send there should park until the rebuilt channel opens rather than fail, so `.reconnecting` deliberately still waits.

Sources/LiveKit/Core/DataChannelPair.swift
```diff
@@ -148,12 +189,26 @@ class DataChannelPair: NSObject, @unchecked Sendable, Loggable {
...
+        // a latch teardown just cleared — leaving the next send to skip the gate and park in a
+        // drain whose `.fail` event is already gone.
+        //
+        // Safe to hold across the drain and completer resets: both settle their continuations
+        // outside their own locks, and nothing taken under `_state` is acquired by a cancellation
+        // handler, so there is no path back into this lock.
+        _state.mutate { state in
+            state.reliableReceivedState.removeAll()
+            lossy.reset(throwing: error)
+            reliable.reset(throwing: error)
```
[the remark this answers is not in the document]
pblazej: Fixed in 975d02d3 — you are right, and this was mine: the delegate setter is a proxied BlockingCall onto the signaling thread, which may already be in `dataChannelDidChangeState` waiting for `_state`. Split it the way you suggest: `detachSendTarget(throwing:)` does the state transition with no WebRTC call and hands the channel back, `release(_:)` detaches and parks it, and the pair takes only the state transitions (both drains plus all three latches, still atomic against `handleStateChange`) under the lock. It matches a job that hung to its 30-minute cap on the previous run.

pblazej:
Closing in favor of https://github.com/livekit/client-sdk-swift/pull/1131

PR #1120, opened 2026-09-11
fix: fail over to other regions when Cloud rejects a connection with 403

xianshijing-lk:
Part of CLT-3325. Companion PRs: [client-sdk-js#2097](https://github.com/livekit/client-sdk-js/pull/2097), [client-sdk-flutter#1200](https://github.com/livekit/client-sdk-flutter/pull/1200).

## Problem

LiveKit Cloud enforces project-level region pinning by returning **403 on the RTC paths** (`/rtc`, `/rtc/validate`) when a project is not allowed in the region the client geo-routed to. `/settings/regions` is deliberately excluded from that gate so the client can discover its allowed regions and connect there — per the server-side comment, *"allow other APIs to pass because clients will permanently give up if they fail."*

Swift blocked that recovery **twice**. `HTTP.requestValidation` maps any 4xx except 404/429 to `.validation`, and `connectWithCloudRegionFailover` short-circuited on it:

```swift
if let liveKitError = error as? LiveKitError, liveKitError.type == .validation {
    // Don't retry other regions for validation errors.
    throw liveKitError
}
```

and even without that, `isRetryableForRegionFailover` admitted only `.network` / `.timedOut`. So a pinned project that geo-routed to a disallowed region gave up before ever fetching the region list.

## Fix

1. Carry the HTTP status on `LiveKitError` as a new optional `statusCode`, set where `HTTP.requestValidation` classifies the response.
2. Admit a 403 in `isRetryableForRegionFailover`.
3. Drop the `.validation` short-circuit in `connectWithCloudRegionFailover`, which is now redundant — `.validation` without a 403 still returns `false` and throws exactly as before.
4. Report region exhaustion as a `nil` result from `RegionManager.resolveBest` rather than a thrown `.regionManager`, so `connectWithCloudRegionFailover` rethrows the connection error it is already holding instead of replacing it. A throw from `resolveBest` still means a genuine settings fetch / HTTP / parse failure and propagates unchanged.
5. Keep failed regions excluded across a settings refresh in `applyFetchedRegions`, matching `updateFromServerReportedRegions`.
6. Cap the failover loop at 10 attempts, so termination does not rest on an invariant about how `remaining` is maintained.

Net behaviour change: **403 → retry other regions**; 401 stays terminal (no other region will accept the same token), as does the 404 that surfaces as `.serviceNotFound` for the v1 → v0 path fallback.

### Why status and not the error message

The server's body for this case is `"project not allowed in this region."`, but that is an unversioned human-readable string, and today it reaches Swift only embedded in `message` as `"HTTP 403: ..."`. Matching either would mean five SDKs carrying identical literals forever, and a server-side copy edit would silently break already-shipped clients — worst of all on iOS, where a released app cannot be hot-patched. Hence the structured `statusCode`.

The cost of not discriminating is bounded: if a 403 really was a permissions failure, every region attempt fails the same way and the original error still surfaces, one region lookup later.

That last part was **not** true as originally written, and review caught it (thanks @pblazej, Devin). Exhaustion threw `.regionManager("No more remaining regions.")` from inside the same `catch`, discarding the 403 — so a permission failure came back less informative than before this PR. Item 4 above is what actually makes the sentence hold — and follows rust-sdks, whose `livekit-signaling` fallback tracks `last_err` across region attempts and never synthesizes an exhaustion error. A settings fetch/parse failure surfaces the connection error too, matching rust's explicit choice not to mask it. Items 5 and 6 address a related pre-existing gap the wider retry set makes easier to hit: `applyFetchedRegions` refilled `remaining` on every refresh, so once per-region failure outlasts `cacheInterval` the loop could never exhaust and `Room.connect` would hang rather than fail.

## API impact

None. `statusCode` is **internal**, and the public `LiveKitError` initializer keeps its exact signature; a separate internal initializer records the status.

I first added a defaulted `statusCode:` parameter to the existing public init, assuming that was source-compatible — it is for callers, but the Swift API digester reports the changed signature as a *removed declaration*, and `Check Public API` failed on it. Fixed in 376d8da. Exposing `statusCode` publicly would be useful but is a separate decision that shouldn't ride along in a bugfix.

## Cross-SDK status

This bug is not universal. Rust and Android already fail over on any non-cancellation error and are unaffected; JS, Flutter and Swift all bail. Agents SDKs route through the Rust core, so **agents are unaffected**.

## Testing

`xcodebuild build -scheme LiveKit -destination 'platform=macOS'` succeeds. New test `regionPinning403IsRetryableButOtherValidationFailuresAreNot` covers 403 / 401 / 404; the existing `regionManagerShouldRetryConnection` is unchanged and still passes. `RegionManagerTests` passes in full (6 tests).

`LiveKitCoreTests` run against a local `livekit-server --dev` leaves 3 failures — `defineAndGetSchema`, `publishWithFrameMetadata`, `concurrentPush` — all data-track schema tests, which `AGENTS.md` documents as requiring `enable_participant_data_blob: true`; the `livekit-server` build available locally rejects that config key. I was not able to baseline these against `main` (the run exceeded my time budget), so I am flagging rather than asserting they are pre-existing — though this change only touches error classification on the region-failover path and has no plausible connection to data-track schemas.

## Open question

How often the RTC 403 actually fires for pinned projects has **not** been confirmed with the Cloud team — geo-routing may normally land clients in-region, making this an edge case (bad GeoDNS, anycast flap, VPN). The opt-in `prepareConnection()` warm-up also does region selection up front and would mask it for apps that call it. The fix is correct either way and works against today's servers, but severity is unconfirmed.

### DIFF
--- a/Sources/LiveKit/Core/Room+Region.swift
+++ b/Sources/LiveKit/Core/Room+Region.swift
@@ -137,11 +137,10 @@ extension Room {
                     throw error
                 }
 
-                if let liveKitError = error as? LiveKitError, liveKitError.type == .validation {
-                    // Don't retry other regions for validation errors.
-                    throw liveKitError
-                }
-
+                // Validation errors are terminal — another region will reject the same token the
+                // same way — except for the 403 Cloud uses to signal project-level region pinning,
+                // which is precisely the case region failover exists to recover from.
+                // `isRetryableForRegionFailover` makes that distinction on the HTTP status.
                 guard error.isRetryableForRegionFailover else {
                     throw error
                 }
--- a/Sources/LiveKit/Errors.swift
+++ b/Sources/LiveKit/Errors.swift
@@ -143,9 +143,9 @@ public class LiveKitError: NSError, @unchecked Sendable, Loggable {
     public let internalError: Error?
 
     /// HTTP status this error was derived from, when it came from an HTTP response.
-    /// Callers should branch on this rather than parsing `message`, which is a
+    /// Internal consumers branch on this rather than parsing `message`, which is a
     /// human-readable string with no stability guarantees.
-    public let statusCode: Int?
+    let statusCode: Int?
 
     @available(*, deprecated, renamed: "internalError")
     public var underlyingError: Error? { internalError }
@@ -154,33 +154,53 @@ public class LiveKitError: NSError, @unchecked Sendable, Loggable {
         [internalError].compactMap(\.self)
     }
 
-    public init(_ type: LiveKitErrorType,
-                message: String? = nil,
-                internalError: Error? = nil,
-                statusCode: Int? = nil)
+    private static func _userInfo(type: LiveKitErrorType,
+                                  message: String?,
+                                  internalError: Error?) -> [String: Any]
     {
-        func _computeDescription() -> String {
-            var suffix = ""
-            if let message {
-                suffix = "(\(message))"
-            } else if let internalError {
-                suffix = "(\(internalError.localizedDescription))"
-            }
-            return String(describing: type) + suffix
+        var suffix = ""
+        if let message {
+            suffix = "(\(message))"
+        } else if let internalError {
+            suffix = "(\(internalError.localizedDescription))"
         }
 
+        var userInfo: [String: Any] = [NSLocalizedDescriptionKey: String(describing: type) + suffix]
+        if let internalError {
+            userInfo[NSUnderlyingErrorKey] = internalError as NSError
+        }
+        return userInfo
+    }
+
+    public init(_ type: LiveKitErrorType,
+                message: String? = nil,
+                internalError: Error? = nil)
+    {
+        self.type = type
+        self.message = message
+        self.internalError = internalError
+        statusCode = nil
+
+        super.init(domain: "io.livekit.swift-sdk",
+                   code: type.rawValue,
+                   userInfo: Self._userInfo(type: type, message: message, internalError: internalError))
+    }
+
+    /// Records the HTTP status alongside the error. Kept separate from the public initializer so
+    /// that one's signature — and the public API surface — stays unchanged.
+    init(_ type: LiveKitErrorType,
+         message: String? = nil,
+         internalError: Error? = nil,
+         statusCode: Int?)
+    {
         self.type = type
         self.message = message
         self.internalError = internalError
         self.statusCode = statusCode
 
-        var userInfo: [String: Any] = [NSLocalizedDescriptionKey: _computeDescription()]
-        if let internalError {
-            userInfo[NSUnderlyingErrorKey] = internalError as NSError
-        }
         super.init(domain: "io.livekit.swift-sdk",
                    code: type.rawValue,
-                   userInfo: userInfo)
+                   userInfo: Self._userInfo(type: type, message: message, internalError: internalError))
     }
 
     @available(*, unavailable)
--- a/Sources/LiveKit/Errors.swift
+++ b/Sources/LiveKit/Errors.swift
@@ -142,6 +142,11 @@ public class LiveKitError: NSError, @unchecked Sendable, Loggable {
     public let message: String?
     public let internalError: Error?
 
+    /// HTTP status this error was derived from, when it came from an HTTP response.
+    /// Callers should branch on this rather than parsing `message`, which is a
+    /// human-readable string with no stability guarantees.
+    public let statusCode: Int?
+
     @available(*, deprecated, renamed: "internalError")
     public var underlyingError: Error? { internalError }
 
@@ -151,7 +156,8 @@ public class LiveKitError: NSError, @unchecked Sendable, Loggable {
 
     public init(_ type: LiveKitErrorType,
                 message: String? = nil,
-                internalError: Error? = nil)
+                internalError: Error? = nil,
+                statusCode: Int? = nil)
     {
         func _computeDescription() -> String {
             var suffix = ""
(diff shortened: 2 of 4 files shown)


Sources/LiveKit/Core/Room+Region.swift
```diff
@@ -137,11 +137,10 @@ extension Room {
...
-                if let liveKitError = error as? LiveKitError, liveKitError.type == .validation {
-                    // Don't retry other regions for validation errors.
-                    throw liveKitError
-                }
-
+                // Validation errors are terminal — another region will reject the same token the
+                // same way — except for the 403 Cloud uses to signal project-level region pinning,
+                // which is precisely the case region failover exists to recover from.
+                // `isRetryableForRegionFailover` makes that distinction on the HTTP status.
                 guard error.isRetryableForRegionFailover else {
```
pblazej, line 144: Letting 403 through here reaches `try await regionManager.resolveBest(token:)` further down, which throws from inside this same `catch` and discards `error` — so once the region list is exhausted the 403 surfaces as `LiveKitError(.regionManager, "No more remaining regions.")` rather than the server's message, contrary to the PR description; suggest capturing `error` and rethrowing it if `resolveBest` fails.

Sources/LiveKit/Core/Room+Region.swift
```diff
@@ -137,11 +137,10 @@ extension Room {
...
                     throw error
                 }
```
pblazej, line 146: Widening what enters this `while true` also widens exposure to an existing gap: `RegionManager.applyFetchedRegions` resets `remaining` to *all* regions (unlike `updateFromServerReportedRegions`, which filters the failed ids), so when per-region failure is slower than the 30s `cacheInterval` the list refills faster than `markFailed` drains it and `Room.connect` — which puts no attempt cap around this call — never returns.

Sources/LiveKit/Errors.swift
```diff
@@ -230,9 +237,24 @@ extension Error {
...
+    ///
+    /// We key on the status rather than the server's message because that message is an unversioned
+    /// human-readable string; matching it would let a copy edit break already-shipped clients. A 401
+    /// stays terminal — no other region will accept a token this one rejected — as does the 404 that
+    /// surfaces as `.serviceNotFound`. If a 403 really was a permissions failure rather than region
+    /// pinning, every region attempt fails the same way and the original error still surfaces.
     var isRetryableForRegionFailover: Bool {
         if let liveKitError = self as? LiveKitError {
+            if liveKitError.statusCode == 403 {
+                return true
```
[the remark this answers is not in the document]
xianshijing-lk: True, we should surface the right errors. 

I fixed it with the latest patchset: resolveBest returns RegionInfo? with nil for exhaustion, so the failover loop rethrows the connection error it's holding.

Also corrected the isRetryableForRegionFailover docstring.

Sources/LiveKit/Core/Room+Region.swift
xianshijing-lk: Right, thanks. The PR description was wrong about this. Captured via a nil return from resolveBest on exhaustion and throw error at the call site, so the server's 403 is what reaches the caller.

The latest patchset went with a nil result rather than catching the throw because .regionManager is also used for fetch/HTTP/parse failures; those still throw and should reach the caller unchanged.

Sources/LiveKit/Core/Room+Region.swift
xianshijing-lk: Good catch. 

applyFetchedRegions now preserves the failed ids the same way updateFromServerReportedRegions does, so a refresh mid-failover can't refill faster than markFailed drains. Added refreshKeepsFailedRegionsExcluded to cover it.

xianshijing-lk: regarding the attempt cap,  I added a 10 attempt cap so termination doesn't depend on how remaining is maintained.

Sources/LiveKit/Core/Room+Region.swift
xianshijing-lk: Note, I made the change to keep the behavior consistent with Rust:
* Exhaustion returns the last connection error, never a synthetic one
* Region-lookup failure surfaces the connection error, logged as a warning
* Cancellation propagates unchanged
* A failed region isn't handed out again within a cycle

Sources/LiveKit/Core/Room+Region.swift
```diff
@@ -117,6 +120,16 @@ extension Room {
...
 
+    /// Bounds ``connectWithCloudRegionFailover``, independently of how the region list is
+    /// maintained.
+    ///
+    /// The list is refreshed from the server *inside* the loop, so termination must not rest on an
+    /// invariant about how `remaining` is kept. rust-sdks gets this structurally — its fallback
+    /// iterates a list fetched once (`for region_url in urls.iter()` in `livekit-signaling`), so
+    /// nothing can refill it mid-loop. Until this loop is reshaped the same way, the cap is what
+    /// guarantees `connect` returns rather than hangs. Comfortably above any real region count.
+    private static let maxRegionFailoverAttempts = 10
```
xianshijing-lk, line 131: @pblazej , what do you think about 10 attempt cap ? or we should do less?

pblazej: 10 is fine — the server already truncates to 6 ([`len(res.Regions) >= 6`](https://github.com/livekit/cloud/blob/73b992606bc18ea1e07ce24c7a171c2a7cb7402f/pkg/config/region_map.go#L316) in `DefaultRegionMap.GetRegionSettings`, a hard constant rather than a per-geography thing), so the cap never fires in practice; I checked four Cloud projects and got exactly 6 each time. I would not go lower: 10 leaves headroom for the prepared-region path (`initialRegion` is resolved before the loop, so it effectively allows 11) and for that 6 being raised later, whereas anything at or below it turns the safety net into a functional limit.

pblazej (approved):
LGTM, test flakiness addressed elsewhere.

PR #1122, opened 2026-09-12
feat: expose scalabilityMode in VideoPublishOptions

davibittencourtome:
## Problem

`Utils.computeVideoEncodings` hardcodes the scalability mode whenever the preferred codec is SVC:

```swift
if let videoCodec, videoCodec.isSVC {
    return [RTC.createRtpEncodingParameters(encoding: encoding, scalabilityMode: isScreenShare ? .L1T3 : .L3T3_KEY)]
}
```

`ScalabilityMode` is public and `RTC.createRtpEncodingParameters` already takes it, but no public option feeds it, so a camera publish is always `L3T3_KEY`. `livekit-client` exposes the same knob as `VideoPublishOptions.scalabilityMode`, so an app can be configured on the web and not on iOS.

This is not theoretical for us. Against livekit-server 1.9.12 we see the SFU turn the dependency descriptor off for tracks published with `L3T3_KEY` and never turn it back on; every subscriber of that publisher then freezes, while the publisher's own preview looks fine and its outbound stats stay healthy. Our web clients work around it by publishing `L1T3` (one spatial layer, temporal scalability kept). The iOS apps could not, and the only remaining lever was falling back to VP8, which costs noticeably more bitrate for the same quality.

## Fix

Add `VideoPublishOptions.scalabilityMode` (defaults to `nil`, which keeps the current behavior exactly) and honor it in the SVC branch:

```swift
let scalabilityMode = publishOptions.scalabilityMode ?? (isScreenShare ? .L1T3 : .L3T3_KEY)
return [RTC.createRtpEncodingParameters(encoding: encoding, scalabilityMode: scalabilityMode)]
```

The option is threaded through the `overrideVideoCodec` path as well, so it still applies when the backup codec or a republish recomputes the encodings — otherwise the mode would silently revert on the very paths where the freeze shows up.

`ScalabilityMode` gains `Sendable`. `VideoPublishOptions` is a `Sendable`-conforming final class, so a stored property of a non-`Sendable` enum does not compile; the enum is a plain `Int`-backed value, so this is not a behavior change.

The property is included in `isEqual` and `hash`, matching the other publish options.

## Testing

New `VideoEncodingsScalabilityModeTests` (swift-testing):

- `svcDefaultsToL3T3KeyForCamera` / `svcDefaultsToL1T3ForScreenShare` — unchanged defaults when the option is `nil`.
- `explicitScalabilityModeWinsForSvcCamera` — `L1T3` reaches the encoding parameters.
- `explicitScalabilityModeSurvivesCodecOverride` — the option still applies on the `overrideVideoCodec` path.
- `scalabilityModeIsIgnoredForNonSvcCodec` — VP8 still produces simulcast layers, none carrying a scalability mode.
- `optionsEqualityIncludesScalabilityMode` — equality and hash account for the new property.

Red/green checked: reverting only the `Utils+VideoEncodings.swift` change makes the two explicit-mode tests fail and leaves the four others passing. Run with `swift test --filter VideoEncodingsScalabilityModeTests` on macOS.

I did not find an existing issue for this; happy to open one if you prefer to track it separately.


Sources/LiveKit/Support/Utils.swift
```diff
@@ -313,9 +313,12 @@ class Utils: Loggable {
 
         if let videoCodec, videoCodec.isSVC {
             // SVC mode
-            log("Using SVC mode")
-            // VP9/AV1 with screen sharing requires single spatial layer
-            return [RTC.createRtpEncodingParameters(encoding: encoding, scalabilityMode: isScreenShare ? .L1T3 : .L3T3_KEY)]
+            // VP9/AV1 with screen sharing requires single spatial layer.
+            // An explicit `publishOptions.scalabilityMode` wins over the defaults (parity with
+            // livekit-client `scalabilityMode`).
+            let scalabilityMode = publishOptions.scalabilityMode ?? (isScreenShare ? .L1T3 : .L3T3_KEY)
```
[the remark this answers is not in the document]
davibittencourtome: Good catch, fixed. Screen share now keeps `.L1T3` unconditionally and the option applies only to camera encodings:

```swift
let scalabilityMode: ScalabilityMode = isScreenShare ? .L1T3 : (publishOptions.scalabilityMode ?? .L3T3_KEY)
```

The property documentation now states that screen share always uses `.L1T3`. Added `explicitScalabilityModeIsIgnoredForScreenShare`, which fails against the previous version of this line.

Sources/LiveKit/Types/Options/VideoPublishOptions.swift
```diff
@@ -49,6 +59,7 @@ public final class VideoPublishOptions: NSObject, TrackPublishOptions, Sendable
                 screenShareSimulcastLayers: [VideoParameters] = [],
                 preferredCodec: VideoCodec? = nil,
                 preferredBackupCodec: VideoCodec? = nil,
+                scalabilityMode: ScalabilityMode? = nil,
```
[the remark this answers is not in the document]
davibittencourtome: Fixed. The previous initializer is preserved as a convenience init that forwards `scalabilityMode: nil`, keeping its original default arguments so existing Swift source is unaffected too.

One difference from the `AudioCaptureOptions` precedent: I left the selector to be derived rather than declaring `@objc(...)` explicitly. With the explicit attribute, `Check Public API` reported an ObjC name change on that constructor (from derived to explicit); deriving it keeps both sides identical. `Check Public API` now passes on iOS and macOS.

Added `legacyInitializerDefaultsScalabilityModeToNil` for the forwarding behavior.

Sources/LiveKit/Types/Options/VideoPublishOptions.swift
```diff
@@ -37,6 +37,17 @@ public final class VideoPublishOptions: NSObject, TrackPublishOptions, Sendable
...
+    /// Scalability mode used when publishing the **camera** with an SVC codec (VP9/AV1).
+    /// `nil` keeps the default, `L3T3_KEY`. Ignored for non-SVC codecs (VP8/H264), which use
+    /// simulcast instead, and ignored for screen share, which always uses `L1T3` because
+    /// WebRTC does not publish SVC screen share with multiple spatial layers.
+    ///
+    /// Use `.L1T3` to publish a single spatial layer with temporal scalability — the same
+    /// configuration `livekit-client` (JS) exposes as `scalabilityMode: 'L1T3'`. Motivation:
+    /// with `L3T3_KEY` some SFU versions turn the dependency descriptor off for the track
+    /// and subscribers freeze; a single spatial layer avoids it.
+    public let scalabilityMode: ScalabilityMode?
```
pblazej, line 49: `ScalabilityMode?` (an Optional of an `@objc` enum) is not Objective-C-representable, so `@objcMembers` silently drops **both** this property and the new designated initializer from the generated header — verified with `-emit-objc-header`: `@interface VideoPublishOptions` has no `scalabilityMode` and exactly one init, the legacy 10-arg one — which also means the convenience init below buys nothing, and modelling this as a non-optional enum with an `.auto` case (the pattern the sibling `degradationPreference` already uses) would fix both at once.

Sources/LiveKit/Types/Options/VideoPublishOptions.swift
```diff
@@ -37,6 +37,17 @@ public final class VideoPublishOptions: NSObject, TrackPublishOptions, Sendable
 
     public let preferredBackupCodec: VideoCodec?
 
+    /// Scalability mode used when publishing the **camera** with an SVC codec (VP9/AV1).
+    /// `nil` keeps the default, `L3T3_KEY`. Ignored for non-SVC codecs (VP8/H264), which use
+    /// simulcast instead, and ignored for screen share, which always uses `L1T3` because
```
pblazej, line 42: This invariant can't hold on the backup-codec path: `LocalParticipant.publish(additionalVideoCodec:)` (LocalParticipant.swift:520) calls `computeVideoEncodings` without `isScreenShare:`, so a screen-share track with a `preferredBackupCodec` gets its backup stream computed as camera — camera encoding, camera simulcast layers and camera presets, not just the scalability mode — unlike line 648, which correctly passes `isScreenShare: track.source == .screenShareVideo`.

Sources/LiveKit/Support/Utils+VideoEncodings.swift
```diff
@@ -31,9 +31,13 @@ extension Utils {
...
         if let videoCodec, videoCodec.isSVC {
             // SVC mode
-            log("Using SVC mode")
-            // VP9/AV1 with screen sharing requires single spatial layer
-            return [RTC.createRtpEncodingParameters(encoding: encoding, scalabilityMode: isScreenShare ? .L1T3 : .L3T3_KEY)]
+            // VP9/AV1 with screen sharing requires single spatial layer, so `.L1T3` stays forced
+            // there — a multi-spatial mode publishes no frames at all. An explicit
+            // `publishOptions.scalabilityMode` only applies to camera encodings (parity with
+            // livekit-client `scalabilityMode`).
+            let scalabilityMode: ScalabilityMode = isScreenShare ? .L1T3 : (publishOptions.scalabilityMode ?? .L3T3_KEY)
```
pblazej, line 38: Since `videoCodec` is `overrideVideoCodec ?? publishOptions.preferredCodec` (line 30), an app that sets `scalabilityMode` but leaves `preferredCodec` at its `nil` default never reaches this branch even when SDP negotiates VP9 — the option is a complete no-op, and the only log here is `.debug` and prints the *resolved* mode rather than the fact that the caller's value was dropped, so a `.warning` when `publishOptions.scalabilityMode` is non-nil and discarded would save someone a long debugging session.

PR #1124, opened 2026-09-14
test: flakiness improvements

pblazej:
Fixes the causes behind the 30-minute `Build & Test` timeouts of Sep 10–13 — a `flushEvents()` barrier race in `DropOldestContinuationTests`, an uncancellable UniFFI read under `next(within:)`, `RemoteParticipantTests` faking inactivity by wiping participant info, and, found via the new stall-dump watchdog, an `AsyncCompleter` deadlock between its timeout and cancellation paths — one commit each, with retries dropped so a flake fails red instead of doubling the run.


hiroshihorie (approved):

pblazej:
Let's revisit above failures after merging #1116 which seems to be related.

PR #1094, opened 2026-08-20
Enable WARP (SPED + SNAP), gated by the server

hiroshihorie:
Enables WARP, negotiated with the SFU. SPED (DTLS-in-STUN) via the WebRTC-IceHandshakeDtls field trial, SNAP (SCTP INIT in SDP) via enableSctpSnap on the default configuration. Nothing engages unless the server enables warp, otherwise plain DTLS/SCTP. Mirrors livekit/rust-sdks#1342.

Blocked on an m150 webrtc-xcframework release including webrtc-sdk/webrtc#283, then both Package manifest pins need bumping.

Note: SPED re-enters the #929 crash surface that #1015 disabled. The m150 teardown path should be validated (or rust field data observed) before this lands. SNAP has no such history.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1128, opened 2026-09-15
Add WARP support (DTLS handshake carried in the ICE exchange)

cloudwebrtc:
`LiveKitSDK.setWARPEnabled(_:)` opts the process into WARP (draft-uberti-tsvwg-warp), shortening connection setup from ~6 round trips to ~2. It turns on libwebrtc's `WebRTC-IceHandshakeDtls` field trial, which piggybacks the DTLS handshake on the ICE STUN binding exchange so DTLS and ICE negotiate in parallel, and enables SCTP SNAP marking on the peer connections created afterwards.

The trial is written to libwebrtc's process-global field trial string rather than passed to the peer connection factory: the factory snapshots trials into its environment when it is built, while the global string is re-read on every lookup, and
`PeerConnection::CanAttemptDtlsStunPiggybacking()` looks this key up as each peer connection is created. So the setting also reaches rooms connected after the factory exists.

`AudioManager.init()` no longer installs the ADM delegate adapter itself; the peer connection factory installs it on the ADM it creates. Doing it from the initializer built the factory as soon as anything read `AudioManager.shared` — `Room.init` does — locking in the settings that are only read at factory creation (`admType`, `bypassVoiceProcessing`) before the caller had a chance to change them.

Offers now log `negotiate with sped (WARP)` when they carry the `goog-sped-v1` ice-option, which is what libwebrtc adds once it will actually piggyback the handshake.


pblazej:
Thanks @cloudwebrtc — WARP negotiates end-to-end from Swift, both halves. Confirmed against `livekit-server` 1.13.7 (`rtc.enable_warp: true`) and a WARP-enabled staging deployment: libwebrtc logs `DTLS-STUN piggybacking complete` / `success=1`, and the SFU echoes `a=sctp-init:` in both its offer and its answer. Without it, it degrades cleanly and the connection is fine.

Two testing notes: a local dev server **does** do WARP with `rtc.enable_warp: true`, so this is testable without staging; and the SFU does log it — `rtc/transport.go "enable warp"`, one per transport at DEBUG.

### Benchmarks

`BM-CONN` dual PC (the path `connect()` waits on), n=100, p50 unless noted, ms:

| server | | wall p50 | wall p90 | transport | ICE+DTLS |
|---|---|---|---|---|---|
| local, `enable_warp: true` | off | 136 | 175 | 121 | 110 |
| | on | **37** | **43** | **22** | **9** |
| staging, WARP enabled | off | 531 | 596 | 305 | 295 |
| | on | **458** | **504** | **223** | **210** |
| prod, WARP not enabled | off | 300 | — | 164 | 154 |
| | on | 303 | — | 164 | 153 |

Single PC on staging: 463 → 415 wall, 212 → 155 transport; flat on loopback, where offer-with-join already removed that phase.

So ~29% off ICE+DTLS on a real deployment — about one round trip. `ws_open` is unchanged, so the end-to-end number is diluted by it; the loopback figures are inflated by pion's DTLS retransmit timing; against prod it is a clean no-op. p99 omitted — single outlier connects dominate it in both directions.

### One fix before merge

`enableSctpSnap` is set from the process-global `isWARPEnabled` on every `makeRTCConfiguration`, but libwebrtc treats it as immutable per peer connection (`ApplyConfiguration` copies only a whitelist, `operator==` still compares it → `INVALID_MODIFICATION`). Reproduced: `setConfiguration` unchanged succeeds, flipped returns `false`. So any `setWARPEnabled` after a room has connected breaks every later quick reconnect permanently, and fails `connect()` outright on the early-publisher path. #1094 hardcodes `enableSctpSnap = true` in `liveKitDefault()` instead.

Which is your own question from the thread: do we want a runtime toggle at all? Forced on — or set once before the factory exists — the immutability problem disappears, the field-trial string is written once instead of swapped under libwebrtc's readers, and `configureFieldTrials` replaces the deprecated `RTCInitFieldTrialDictionary`. That closes all three open Devin comments at once.

PR #1132, opened 2026-09-18
Switch IPCChannel to raw unix socket transport

creativecbr:
Replaces the Network.framework IPC transport in IPCChannel with a raw UNIX domain stream socket implementation (socket/bind/listen/accept/connect) while preserving LiveKit’s existing IPC framing and public behavior.

On some iOS/iPadOS devices, the current NWListener/NWConnection path for UNIX-socket IPC fails during listener setup with NECP path evaluation errors (observed as SO_NECP_LISTENUUID failed [22: Invalid argument]). In that failure mode, the broadcast extension and host app do not complete IPC handshake, and ReplayKit publishing can come up without frame flow.

What changes in this PR:
- Switches IPCChannel transport from Network.framework to raw POSIX UNIX sockets on iOS.
- Keeps message semantics unchanged: framed messages with header + optional payload, same plist header encoding, same async incoming message shape.
- Preserves compatibility for existing broadcast components (BroadcastReceiver, BroadcastUploader, codecs, SocketPath) with no public API changes.
- Keeps cancellation/error behavior aligned with prior expectations (retry on transient connect errors, close on terminal I/O errors).
- Adds concise [IPC] logs at key lifecycle points (listen/connect/accept/read-loop-end/close), consistent with project logging style.

Scope and non-goals:
- iOS-only IPC transport internals; no changes to cross-platform API surface.
- No protocol/framing redesign; wire format remains equivalent from caller perspective.
- No functional changes outside IPCChannel.

Regression/testing status:

- Manual regression verification completed on:
    - iOS 26.5, iPhone 15
    - iPadOS 17.7, iPad 6th gen

- In both environments, IPC handshake and screen-share flow are stable, and the issue is no longer reproducible.
- Compile/build sanity also passes for the updated path.


creativecbr:
@pblazej @hiroshihorie @xianshijing-lk 

I made the necessary improvements to the first version and refactored it slightly, while trying to keep the changes as atomic as possible. This is a very important change for us, as some devices were not able to start screen sharing at all.

Regression tests were performed on physical devices:
- iPhone 15 (iOS 26.2.1)
- iPad 6th generation (iPadOS 17.7.11)

pblazej:
Hey @creativecbr thank you for a detailed report and PR, I think we need to be a little conservative about essentially reverting to #100 - by understanding the root cause first.

I added more specific logs in this branch, it would be great to check them before the next pass: `blaze/ipc-logs`.

